PR #117, opened 2013-07-18
Fetches all rows by default on query, like libpq.

### LINKED ISSUE #81: Can't run two queries in a transaction?
I have some data I've loaded into a table from a CSV file that I want to massage into a master table.  When I try to do this in a transaction, pq just hangs on the second query I perform for no reason I can ascertain.

Here's my code.  I chopped out everything after the second query; that query checks to see if the data is already migrated, and if not, runs all the create code and such.  If you comment out the "select \* from community_gnis" line, the loop runs to completion, printing every city in the table.  If I leave the query in, it just hangs after the first city name.

```
if db, err := sql.Open("postgres", "user=postgres dbname=geo sslmode=disable"); err == nil {
    defer db.Close()
    if tx, err := db.Begin(); err == nil {
        if cities, err := tx.Query("select name, state, id, lon, lat from noaa.cities where id is not null"); err == nil {

            var name, state, gnis string
            var longitude, latitude float32

            for cities.Next() {

                if err = cities.Scan(&name, &state, &gnis, &longitude, &latitude); err == nil {

                    fmt.Println("Got city", name, "(", gnis, ")")

                    tx.Query("select * from community_gnis where gnis = $1", gnis)
                }
            }
        }
    }
}
```

I'm new to Go, so maybe I'm doing things incorrectly here.  However, if I move the "select \* from community_gnis" up above the "select name, state..." statement, the community_gnis statement runs fine and the "select name, state..." query will then be the one that hangs. 

If I remove the transaction code and run everything against "db", e.g. "db.Query" instead of "tx.Query", etc., everything runs completely fine (though it doesn't run in a transaction, which is what I want).

I've tried various combinations, using prepared statements, adding Close calls everywhere, etc.  Nothing works.  I'm 99% sure it's not the database, because I'm migrating this code from a Node JS app to test whether we should use Go or Node for our project, and the Node version runs completely in a transaction.

RangelReale:
This allows multiple queries to be active simultaneously
Provides a singlerowmode=true parameter, like libpq (to keep working like today)

fixes #81

### DIFF
--- a/buf.go
+++ b/buf.go
@@ -48,6 +48,12 @@ func (b *readBuf) byte() byte {
 
 type writeBuf []byte
 
+func newWriteBuf(c byte) *writeBuf {
+	b := make(writeBuf, 5)
+	b[0] = c
+	return &b
+}
+
 func (b *writeBuf) int32(n int) {
 	x := make([]byte, 4)
 	binary.BigEndian.PutUint32(x, uint32(n))
--- a/conn.go
+++ b/conn.go
@@ -35,17 +35,10 @@ func init() {
 }
 
 type conn struct {
-	c             net.Conn
-	buf           *bufio.Reader
-	namei         int
-	scratch       [512]byte
+	c     net.Conn
+	buf   *bufio.Reader
+	namei int
 	singlerowmode bool
-}
-
-func (c *conn) writeBuf(b byte) *writeBuf {
-	c.scratch[0] = b
-	w := writeBuf(c.scratch[:5])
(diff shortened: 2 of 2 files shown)


timbunce:
@RangelReale any chance you could update this PR?

The current code changes the behavior to fetch-all from the current implicit 'single-row mode'.  If you also inverted the option so the current behaviour was maintained by default I think you'd address the concerns raised in #81 and stand a better chance of getting this merged.

PR #130, opened 2013-08-19
Create a parser for Arrays, and wire up decode to use it for 1d arrays.

jagregory:
Hello. I've started implementing array support, but it's not complete yet. Consider this pull request to be the start of a conversation, if you're interested.

This PR includes a parser, which is heavily inspired by the Go JSON parser, with various tweaks for the PG array syntax (mainly around unquoted strings). It can handle `string`, `int`, `float`, and `bool` arrays.

The following now works:

    var strings []string
    var ints []int
    db.QueryRow("select {\"a\", \"b\", \"c\"}::varchar[], {1,2,3}::int[]").Scan(&arr)

    // strings = ["a", "b", "c"]
    // ints = [1, 2, 3]

The parser has support for multidimensional arrays, but this feature hasn't been wired up to the decoder yet. To do this may require some further changes to the pq decoder or providing a `PgArray` type for usage similar to `NullString`.

Currently only supports unmarshaling of data. Marshaling/using arrays as input parameters can be worked around by using string formatting.

My questions to you are:

  * Is it desirable to have Array parsing (and I imagine eventually HStore) in PQ core, or provide some kind of extension mechanism to allow 3rd parties to create support?
  * Is it important to have multi-dimensions supported before this feature can be considered complete, or is single dimension support good enough on it's own to start with?
  * How extensive should type support be before considering this for merging?
  * Does it make sense to have reading without writing support?

### DIFF
--- a/arrays/decode.go
+++ b/arrays/decode.go
@@ -266,7 +266,9 @@ func (d *decodeState) array(v reflect.Value) {
 		if op == scanEndArray {
 			break
 		}
-		if op != scanArrayValue {
+		if op == scanError {
+			d.error(d.scan.err)
+		} else if op != scanArrayValue {
 			d.error(errPhase)
 		}
 	}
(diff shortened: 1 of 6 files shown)


flimzy:
What's the current status? Is there work to be done I can help with?

freeformz:
I think this can be closed since #466 was merged. Please re-open this issue if I am wrong.

PR #175, opened 2013-10-20
Add sslmode=prefer option

uhoh-itsmaciek:
Unfortunately, we have no existing tests for sslmode and without
something like https://github.com/kr/pgtest there's no easy way
to sensibly test this. For now, verified locally.


uhoh-itsmaciek:
Closing this since I won't retrofit it to be more like libpq's `prefer`, it's questionable to put it in without parity in semantics with `libpq`, and I don't think we should add another pq-specific `PGSSLMODE` option.

PR #219, opened 2014-01-01
Use communicationError for network errors

johto:
This seems cleaner than trying to maintain the list of errors which might have originated from operations on the network socket / TLS layer.

Any thoughts?

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -69,6 +69,12 @@ func (s transactionStatus) String() string {
 	panic("not reached")
 }
 
+// Any failure in communication should be raised as a communicationError to
+// allow errRecover to know what happened.
+type communicationError struct {
+	error
+}
+
 type conn struct {
 	c         net.Conn
 	buf       *bufio.Reader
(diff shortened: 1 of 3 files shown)


tamird:
This seems wrong based on discussion in #422. This change will cause all network errors to bubble up as `driver.ErrBadConn`, despite the fact that some work may have been done on the server.

johto:
> This seems wrong based on discussion in #422. This change will cause all network errors to bubble up as driver.ErrBadConn, despite the fact that some work may have been done on the server.

That's already what happens.  This PR tried to keep the same semantics but without the ugly string comparisons (which, I'm guessing, could fail at any point).

johto:
I wrote:
> I wish there was a way to say "this connection is bad", but also specify an error to be returned to the user if database/sql is going to give up on retrying. The way things currently work inside database/sql is pretty ugly.

Now we actually have the necessary pieces to do this.

PR #246, opened 2014-03-15
Bring johto/notifyutils/NotifySemaphore into core

johto:
We've been running this code in production for a while now and it's been working flawlessly.  I feel like this makes the most common use case of LISTEN / NOTIFY very simple and convenient.

Any thoughts?

### DIFF
--- a/notifysemaphore/notifysemaphore.go
+++ b/notifysemaphore/notifysemaphore.go
@@ -232,7 +232,7 @@ func (s *NotifySemaphore) Close() error {
 
 	// wait for all channels to be closed
 	s.closeWaitGroup.Wait()
-	return nil
+	return s.listener.Close()
 }
 
 // Broadcast a nil *Notification to all listeners.  Caller must be holding
--- a/notifysemaphore/notifysemaphore.go
+++ b/notifysemaphore/notifysemaphore.go
@@ -1,0 +1,297 @@
+/*
+
+NotifySemaphore is a utility type for consumers using LISTEN / NOTIFY to avoid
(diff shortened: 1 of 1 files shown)


dlsniper:
Hi, can I kindly bump this for a merge? Thank you!

johto:
I think "NotifySemaphore" is a bad name, since this doesn't implement the full semaphore spec.  Anyone got any better ideas?

johto:
I guess this is quite close to a sync.Cond, so maybe NotifyCond could work.  I'm still happy to hear other opinions or better ideas, though.

johto:
Also, we've been using [NotifyDispatcher](http://godoc.org/github.com/johto/notifyutils/notifydispatcher) in production (within [allas](https://github.com/johto/allas)]) for more than a month now, and it has also worked flawlessly.  What do we think about merging both of them into `pq`?

uhoh-itsmaciek:
For what it's worth, we've stopped using LISTEN / NOTIFY in our app, so I can't add more field reports. It's not that we hit problems in the implementation; I think ultimately LISTEN / NOTIFY was just not a good fit for what we are doing (we're now polling records with an indexed `updated_at` field that changed recently).

johto:
Yeah, if you don't mind the latency, polling can work just as well if not better.

But I take it you don't object to bringing these two in?  This was initially discussed while adding the Listener code but I don't recall us ever deciding whether to provide these in core or not.

uhoh-itsmaciek:
Yeah, I'm all for it, I just can't personally vouch for it anymore.

PR #300, opened 2014-09-25
Support extra decode types

chris-baynes:
This adds support for decoding `varchar`, `text` -> **string** and `numeric` -> **float**.

The use case is scanning a row with an `interface{}` type when the postgres column types are unknown.This will currently decode columns of text and numeric types as `[]uint8`.

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -75,6 +75,8 @@ func decode(parameterStatus *parameterStatus, s []byte, typ oid.Oid) interface{}
 			errorf("%s", err)
 		}
 		return f
+	case oid.T_varchar, oid.T_text:
+		return string(s)
 	}
 
 	return s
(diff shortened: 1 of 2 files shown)


cbandy:
The docs say not to return `string` from `Scan`, but there is an [open issue](https://github.com/golang/go/issues/6497) about it.

There is also an [older, rejected issue](https://github.com/golang/go/issues/5066).

tamird:
Upstream isn't going to change the rules https://go-review.googlesource.com/#/c/19439/

johto:
Yeah, I believe we can forget this idea.

tamird:
Update: this has new life - upstream just merged my change https://github.com/golang/go/commit/7162c4d05c1fed9db517c92b9a5232243ec453a2

@johto can you reopen this? @chris-baynes are you up for rebasing this after @johto reopens?

chris-baynes:
Yes, will push the rebase once this is reopened - congrats on getting that change pushed through!

tamird:
Thanks! It was a struggle.

I also reopened #436 which is ready to merge now, though this PR should go in as well to do the varchar bit.

PR #317, opened 2015-01-04
Optionally enable infinity timestamp(tz)

truongsinh:
If EnableInfinityTs is not called, "-infinity" and "infinity" will yield error when decoding.

If EnableInfinityTs is called once, it enables this driver to decode Postgres' "-infinity" and "infinity" to predefined boundary minimum and maximum time. When encoding, any time that equals or exceeds predefined minimum or maximum time will be encoded to "-infinity" and "infinity" respectively.

If EnableInfinityTs is called more than once, it will panic.

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -207,11 +207,68 @@ func (c *locationCache) getLocation(offset int) *time.Location {
 	return location
 }
 
+var infinityTsEnabled = false
+var infinityTsNegative time.Time
+var infinityTsPositive time.Time
+
+const (
+	negativeInfinityString          = "-infinity"
+	positiveInfinityString          = "infinity"
+	infinityTsEnabledAlready        = "pq: infinity timestamp enabled already"
+	infinityTsNegativeMustBeSmaller = "pq: infinity timestamp: negative value must be smaller (before) than positive"
+)
+
+/**
+ * If EnableInfinityTs is not called, "-infinity" and "infinity" will return
+ * []byte("-infinity") and []byte("infinity") respectively, and potentially
+ * cause error "sql: Scan error on column index 0: unsupported driver -> Scan pair: []uint8 -> *time.Time",
+ * unless the input param is `*interface{}` instead of `*time.Time`
+ *
+ * If EnableInfinityTs is called with negative >= positive, it will panic
+ *
+ * If EnableInfinityTs is called once correctly, it enables this driver to decode
+ * Postgres' "-infinity" and "infinity" to predefined boundary minimum and maximum time.
+ * When encoding, any time that equals or exceeds predefined minimum or maximum time
+ * will be encoded to "-infinity" and "infinity" respectively.
+ *
+ * If EnableInfinityTs is called more than once, it will panic.
+ */
+func EnableInfinityTs(negative time.Time, positive time.Time) {
+	if infinityTsEnabled {
+		panic(infinityTsEnabledAlready)
+	}
+	if !negative.Before(positive) {
+		panic(infinityTsNegativeMustBeSmaller)
+	}
+	infinityTsEnabled = true
+	infinityTsNegative = negative
+	infinityTsPositive = positive
+}
+
+/**
+ * Testing might want to toggle infinityTsEnabled
+ */
+func disableInfinityTs() {
+	infinityTsEnabled = false
+}
+
 // This is a time function specific to the Postgres default DateStyle
 // setting ("ISO, MDY"), the only one we currently support. This
 // accounts for the discrepancies between the parsing available with
 // time.Parse and the Postgres date formatting quirks.
-func parseTs(currentLocation *time.Location, str string) (result time.Time) {
+func parseTs(currentLocation *time.Location, str string) interface{} {
+	switch str {
+	case negativeInfinityString:
+		if infinityTsEnabled {
+			return infinityTsNegative
+		}
+		return []byte(negativeInfinityString)
+	case positiveInfinityString:
+		if infinityTsEnabled {
+			return infinityTsPositive
+		}
+		return []byte(positiveInfinityString)
+	}
 	monSep := strings.IndexRune(str, '-')
 	// this is Gregorian year, not ISO Year
 	// In Gregorian system, the year 1 BC is followed by AD 1
@@ -307,6 +364,16 @@ func parseTs(currentLocation *time.Location, str string) (result time.Time) {
 
 // formatTs formats t into a format postgres understands.
 func formatTs(t time.Time) (b []byte) {
+	if infinityTsEnabled {
+		// t <= -infinity : ! (t > -infinity)
+		if !t.After(infinityTsNegative) {
+			return []byte(negativeInfinityString)
+		}
+		// t >= infinity : ! (!t < infinity)
+		if !t.Before(infinityTsPositive) {
+			return []byte(positiveInfinityString)
+		}
+	}
 	// Need to send dates before 0001 A.D. with " BC" suffix, instead of the
 	// minus sign preferred by Go.
 	// Beware, "0000" in ISO is "1 BC", "-0001" is "2 BC" and so on
--- a/encode_test.go
+++ b/encode_test.go
@@ -78,7 +78,11 @@ func tryParse(str string) (t time.Time, err error) {
 			return
 		}
 	}()
-	t = parseTs(nil, str)
+	i := parseTs(nil, str)
+	t, ok := i.(time.Time)
+	if !ok {
+		err = fmt.Errorf("Not a time.Time type, got %#v", i)
+	}
 	return
 }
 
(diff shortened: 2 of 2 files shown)


truongsinh:
Latest commit failed on Go 1.1 and 1.2 with "Scanning infinity, expected error, "pq: infinity timestamp(tz) is not enabled, got -infinity", got "sql: no rows in result set"
I have no idea why Go 1.1 and 1.2 yield no rows!

johto:
QueryRow was [buggy](https://github.com/golang/go/commit/1f20ab1116ab6cb0b77e22ffba3de9919e9def50) pre-1.3.  We have [some](1e2ccf84caca692327146254602c580cbfd0a690) [workarounds](f8ffc32df8b9c5fd7d5ca1ac8345d75e82234edd) in place, but those only work for ErrorResponses from the server, not actual errors from Next().

I guess you could implement your own queryRow helper and use that..

truongsinh:
Should I just ignore new test case for pre-1.3?

cbandy:
Should this setting be per `DB`?

johto:
> Should I just ignore new test case for pre-1.3?

No.  Either implement a non-buggy queryRow() or use something like [this](https://github.com/lib/pq/blob/3e3efe51a0c4b23e817075438703cb15bd58c1a8/conn_test.go#L1122).

johto:
> Should this setting be per DB?

Honestly, I really don't like the interface right now.  I can see why the author chose to go with it, but having package-level variables affect all connections like that feels a bit icky.

It's not clear that this would be much better, but one idea would be to just use values which are out of range for PG's timestamps but not Go's time.Time, and provide exported constants for comparisons.  For example '-10000-01-01' and '1000000-01-01' for -infinity and infinity, respectively.  These would "just work" for comparisons in the Go app, just like the infinities do on the PG side of things.

truongsinh:
@cbandy I would love that approach too, but found no way at the moment.
@johto I was thinking of timestamp outside PG's range too, but, for example, my project heavily utilize JSON, and Go's bulitin JSON limit year when Marshaling JSON [0000,9999].

truongsinh:
I spent time and found that the workaround you referenced fixed Go's bug only for error received from PG; any error raised by [decode function](https://github.com/lib/pq/blob/master/encode.go#L48) will yield `"sql: no rows in result set"`. In fact, https://github.com/lib/pq/blob/master/encode.go#L65 and https://github.com/lib/pq/blob/master/encode.go#L75 will, so I still have no idea how to proceed.

johto:
> I spent time and found that the workaround you referenced fixed Go's bug only for error received from PG; any error raised by decode function will yield "sql: no rows in result set".

Yes, that's what I said.

> In fact, https://github.com/lib/pq/blob/master/encode.go#L65 and https://github.com/lib/pq/blob/master/encode.go#L75 will, so I still have no idea how to proceed.

Really?  Because I've already suggested something like [this](https://gist.github.com/johto/c896b3ae2a03976ce622) twice, and one other alternative at least once.  Was there something wrong with either suggestion?

truongsinh:
I applied @johto 's gist. Thank you, I thought you were talking to *another* queryRow helper, which I just wanted to copy-and-paste.

truongsinh:
I really like the idea of postponing error until the latest moment possible, assuming it works. Let me try and make some tests for this approach to confirm.

truongsinh:
I reverted the workaround, as `decode` no longer yields error, but returns []byte in such cases:
```
If EnableInfinityTs is not called, "-infinity" and "infinity" will return []byte("-infinity") and []byte("infinity") respectively, and potentially cause error "sql: Scan error on column index 0: unsupported driver -> Scan pair: []uint8 -> *time.Time", unless the input param is `*interface{}` instead of `*time.Time`
```

encode.go
```diff
@@ -212,11 +212,68 @@ func (c *locationCache) getLocation(offset int) *time.Location {
 	return location
 }
 
+var infinityTsEnabled = false
+var infinityTsNegative time.Time
+var infinityTsPositive time.Time
+
+const (
+	_infinityString                 = "-infinity"
```
johto: I think this should be called something else.  negativeInfinityString, for example.

encode.go
```diff
@@ -212,11 +212,68 @@ func (c *locationCache) getLocation(offset int) *time.Location {
...
+	infinityTsNegativeMustBeSmaller = "pg: infinity timestamp: negative value must be smaller (before) than positive"
+)
+
+/**
+ * If EnableInfinityTs is not called, "-infinity" and "infinity" will return
+ * []byte("-infinity") and []byte("infinity") respectively, and potentially
+ * cause error "sql: Scan error on column index 0: unsupported driver -> Scan pair: []uint8 -> *time.Time",
+ * unless the input param is `*interface{}` instead of `*time.Time`
+ *
+ * If EnableInfinityTs is called with negative >= positive, it will panic
```
johto: I think we should allow the case where negative = positive.  This check feels somewhat arbitrary as-is, but I guess it's OK if we allow them to be equal.

encode.go
```diff
@@ -212,11 +212,68 @@ func (c *locationCache) getLocation(offset int) *time.Location {
...
 }
 
+var infinityTsEnabled = false
+var infinityTsNegative time.Time
+var infinityTsPositive time.Time
+
+const (
+	_infinityString                 = "-infinity"
+	infinityString                  = "infinity"
+	infinityTsEnabledAlready        = "pg: infinity timestamp enalbed already"
```
johto: Typo: enalbed

encode.go
```diff
@@ -212,11 +212,68 @@ func (c *locationCache) getLocation(offset int) *time.Location {
...
+	infinityTsNegativeMustBeSmaller = "pg: infinity timestamp: negative value must be smaller (before) than positive"
```
johto: These both error message should have the prefix `pq` (pee cue), not `pg`.

encode_test.go
```diff
@@ -243,6 +247,131 @@ func TestTimestampWithOutTimezone(t *testing.T) {
...
+		err = db.QueryRow(q.Query, q.Param).Scan(&resultT)
+		if err.Error() != q.ExpectedErr.Error() {
+			t.Errorf("Scanning -/+infinity, expected error, \"%s\", got \"%s\"", q.ExpectedErr, err)
+		}
+	}
+	// yield []byte
+	for _, q := range tc {
+		err = db.QueryRow(q.Query, q.Param).Scan(&resultI)
+		if err != nil {
+			t.Errorf("Scanning -/+infinity, expected no error, got \"%s\"", err)
```
johto: Better to just t.Error(err) here.  The test output will show which line the error came from, so it's obvious that an error was not expected.

Ditto for all similar cases below.

encode_test.go
```diff
@@ -243,6 +247,131 @@ func TestTimestampWithOutTimezone(t *testing.T) {
...
+	// try to assert []byte to time.Time
+	for _, q := range tc {
+		err = db.QueryRow(q.Query, q.Param).Scan(&resultT)
+		if err.Error() != q.ExpectedErr.Error() {
+			t.Errorf("Scanning -/+infinity, expected error, \"%s\", got \"%s\"", q.ExpectedErr, err)
+		}
+	}
+	// yield []byte
+	for _, q := range tc {
+		err = db.QueryRow(q.Query, q.Param).Scan(&resultI)
```
johto: resultI is only used in this loop; can we move its definition here?  It's clearer that way.

encode_test.go
```diff
@@ -243,6 +247,131 @@ func TestTimestampWithOutTimezone(t *testing.T) {
...
+		{"SELECT $1::timestamp", "-infinity", expectedError, _infinityString},
+		{"SELECT $1::timestamptz", "-infinity", expectedError, _infinityString},
+		{"SELECT $1::timestamp", "infinity", expectedError, infinityString},
+		{"SELECT $1::timestamptz", "infinity", expectedError, infinityString},
+	}
+	// try to assert []byte to time.Time
+	for _, q := range tc {
+		err = db.QueryRow(q.Query, q.Param).Scan(&resultT)
+		if err.Error() != q.ExpectedErr.Error() {
+			t.Errorf("Scanning -/+infinity, expected error, \"%s\", got \"%s\"", q.ExpectedErr, err)
```
johto: It's better to use %q than \"%s\"; less typing and automatically escaped correctly.

Ditto for all similar uses below.

johto:
I've left some comments, but I think this is a reasonable workaround for now.  Ideally, we'd deprecate this interface in the future in favour of something better, but I think it's unfair to just sit idle waiting for it to come along.

truongsinh:
Sorry, family emergency! I saw that you also merged with change my another PR. Will come back to this soon!

johto:
> Sorry, family emergency! I saw that you also merged with change my another PR. Will come back to this soon!

Oh, take all the time you need.  The comment about "sitting idle" was not directed at you.  What I meant with that was that we shouldn't reject this PR, because it solves the problem now, and we can easily deprecate this interface *if* something better comes along later.

encode_test.go
truongsinh: How about current test cases? I think they all have verbose description.

encode.go
truongsinh: done

encode.go
truongsinh: done

encode.go
truongsinh: done

encode_test.go
truongsinh: done

encode_test.go
truongsinh: done

encode.go
truongsinh: i'm not sure it's a good idea to allow negative infinity = positive infinity, because at the equal point, we don't know whether to put negative or positive into PG

johto: Ah.  That's a good point.

I was worried about someone who can't process infinities at all and just wants to display a nice error if it gets those back from the database, but I guess in such a situation you can already do that with interface{} and dealing with the time.Time/[]byte difference.

truongsinh:
Travis CI:
```
--- FAIL: TestDecodeBool (20.00s)
	bench_test.go:352: read tcp 127.0.0.1:5432: i/o timeout
```
not my fault, is it?

uhoh-itsmaciek:
Don't think so... I restarted the Travis build in case it was a transient failure.

johto:
Hmm.  What up with the unrelated changes?  Can you rebase this on top of master?

truongsinh:
I squashed all the commits on top of master.

truongsinh:
CI failed due to 1 "FAIL: TestSSLClientCertificates" and 2 timeouts

johto:
I've pushed this in b34cf4c2f62cb77dd41e169c33c9856e98469c94 with some small changes to the comments.  I also got rid of the `positiveInfinityString` and `negativeInfinityString` constant, since I found them way less readable than just "infinity" and "-infinity".

Thanks for your work here!

Issue #320, opened 2015-01-04
incorrect encoding of B.C. years


truongsinh:
Commit b7af72a1fc1649a8871af32e23229fb89ae6cdc7 wrongly assume that year 0 in Go is year 0 in PG. Here's the copy from that commit discussion:

Beware, ["year 0"](http://en.wikipedia.org/wiki/0_(year)) in Go is actually "0001 BC" in PG. Consequently, `time.Date(-10000, time.February, 3, 4, 5, 6, 0, time.FixedZone("", 0))` is "0011-02-03 04:05:06 BC" (**not "0010-02-03 04:05:06 BC"**)

You can try this is PG:
```sql
 SELECT '0001-02-16 20:38:40'::timestamp; -- ok
 SELECT '0001-02-16 20:38:40 BC'::timestamp; -- ok
 SELECT '0000-02-16 20:38:40'::timestamp; -- parse error
 SELECT '0000-02-16 20:38:40 BC'::timestamp; -- parse error
```

johto:
[Yeah.](http://play.golang.org/p/xRi3oicSyL).  That sucks.

PR #321, opened 2015-01-05
fix ISO year vs Gregorian year.

truongsinh:
In this Gregorian system, the year 1 BC is followed by AD 1. In ISO system, year 1 is followed by year 0, and then -1.

Ref b7af72a and #320

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -218,6 +218,8 @@ func (c *locationCache) getLocation(offset int) *time.Location {
 // time.Parse and the Postgres date formatting quirks.
 func parseTs(currentLocation *time.Location, str string) (result time.Time) {
 	monSep := strings.IndexRune(str, '-')
+	// this is Gregorian year, not ISO Year
+	// In Gregorian system, the year 1 BC is followed by AD 1
 	year := mustAtoi(str[:monSep])
 	daySep := monSep + 3
 	month := mustAtoi(str[monSep+1 : daySep])
@@ -245,7 +247,6 @@ func parseTs(currentLocation *time.Location, str string) (result time.Time) {
 
 	nanoSec := 0
 	tzOff := 0
-	bcSign := 1
 
 	if remainderIdx < len(str) && str[remainderIdx:remainderIdx+1] == "." {
 		fracStart := remainderIdx + 1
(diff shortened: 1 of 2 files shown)


cbandy:
Are there some links on this topic, calendar of Go and/or Postgres?

encode.go
```diff
@@ -306,15 +310,26 @@ func parseTs(currentLocation *time.Location, str string) (result time.Time) {
...
-	if b[0] == '-' {
-		b = append(b[1:], ' ', 'B', 'C')
+	// Beware, "0000" in ISO is "1 BC", "-0001" is "2 BC" and so on
+	yyyy := t.Year()
+	if yyyy > 0 {
+		yearString := fmt.Sprintf("%04d", yyyy)
+		timeString = yearString + timeString
+	} else {
+		yearString := fmt.Sprintf("%04d", 1-yyyy)
+		timeString = yearString + timeString + " BC"
```
cbandy: Can this concat be done directly in `Sprintf`?

johto: Can't this simply be expressed as:

```
if t.Year() <= 0 {
    t = t.AddDate(-1, 0, 0)
}
```

truongsinh:
@cbandy you can try this in PG:
```sql
select to_timestamp('0001','IYYY'), to_timestamp('0001','IYYY') - '1 year'::interval, to_timestamp('0000','IYYY'), to_timestamp('-0001','IYYY')
--- "0001", "0001 BC", "0001 BC", "0002 BC"
```

Go example is already [demonstrated](http://play.golang.org/p/xRi3oicSyL) by @johto

encode.go
truongsinh: So, in general, reuse `time.RFC3339Nano` by:
```go
bc := false
if t.Year() <= 0 {
    // flip year sign, and add 1, e.g: "-10" will be "11"
    t = t.AddDate((-t.Year)*2+1, 0, 0)
    bc = true
}
b = []byte(t.Format(time.RFC3339Nano))
if bc {
		b = append(b, '  BC'...)
}
```

johto: Ah.  Nice.  I think that's better.

johto:
Thanks!  Pushed in 4b119a25550a9897e86a04e6baf2ef25d82d5dc4.

PR #322, opened 2015-01-06
fix formatTs

truongsinh:
My previous PR didn't add any test cases to formatTs. This PR adds them, and discovers some bugs, thus it fixes format TS, as there are more than one `return` statements in `formatTs` function, but "BC" should only be appended to the very end of time string.

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -324,9 +324,12 @@ func formatTs(t time.Time) (b []byte) {
 		bc = true
 	}
 	b = []byte(t.Format(time.RFC3339Nano))
-	if bc {
-		b = append(b, "  BC"...)
-	}
+	// there are more than one `return` statements in this function
+	defer func (){
+		if bc {
+			b = append(b, " BC"...)
+		}
+	}()
 
 	_, offset := t.Zone()
 	offset = offset % 60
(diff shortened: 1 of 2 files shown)


johto:
Ugh.  That's a horrifying use of defer.

In this case it's trivial to get rid of the multiple returns.  Just do:

```
…
offset = offset % 60
if offset != 0 {
    // relevant code here
}
if bc {
    b = append(b, " BC"…)
}
return b
```

truongsinh:
Can you please explain more why you would like to avoid `defer` as much as possible? As I understand, you want to move all `// relevant code here` to inside an if block. IMO, it creates one more indent level, but is still okay anyway.

johto:
> Can you please explain more why you would like to avoid defer as much as possible?

I generally don't try and avoid defer as much as possible.  It's not free (it does an allocation, which can be costly, but often that won't even show up), but it can often make code cleaner and easier to read.

In this instance, though, changing the return value after the function logic has finished?  I don't like that, not one bit.  It's not the fact that it's a defer, it's the fact that it's a defer with non-obvious side effects.

> As I understand, you want to move all // relevant code here to inside an if block. IMO, it creates one more indent level, but is still okay anyway.

Yes.  That's probably why the code used two returns originally.  But I feel like changing it to use only a single return is fine in this case, because 1) it's still only two levels of indentation, and 2) it's way better than using defer to mutate the return value.

johto:
Thanks, merged with changes in f709b902ce82147d238cd7c548d6857908e7c6be.

PR #323, opened 2015-01-08
Get current OS user from USER variable first, fallback to user.Current

sosedoff:
When cross-compiling projects that use `pq`, call to `user.Current` could result in error, due to cgo usage:

```
user: Current not implemented on linux/amd64
```

Change in this PR will try to detect current user from `$USER` env variable, then fallback to standard method.
Let me know if you have any interest in merging it in.


johto:
Why not os.Getenv()?

sosedoff:
@johto actually yes, i think i got that part from some experiments. Let me update and squash.

johto:
Also, I think we should only look at $USER if user.Current() failed.  I don't see the point of changing the behaviour for the (probably vast majority) people who are not cross-compiling.

sosedoff:
@johto Ok.

sosedoff:
thanks!

PR #325, opened 2015-01-21
remove double call to t.Format() in formatTs

hydrogen18:
From what I can tell there is no reason to call t.Format() twice. The first result is always discarded no matter what. The tests passed before and after.

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -323,9 +323,12 @@ func formatTs(t time.Time) (b []byte) {
 		bc = true
 	}
 	b = []byte(t.Format(time.RFC3339Nano))
-	if bc {
-		b = append(b, "  BC"...)
-	}
+	// there are more than one `return` statements in this function
+	defer func (){
+		if bc {
+			b = append(b, " BC"...)
+		}
+	}()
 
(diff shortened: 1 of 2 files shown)


encode.go
```diff
@@ -324,25 +323,26 @@ func formatTs(t time.Time) (b []byte) {
...
-		offset = -offset
+		if offset < 0 {
+			offset = -offset
+		}
+
+		b = append(b, ':')
+		if offset < 10 {
+			b = append(b, '0')
+		}
+		b = strconv.AppendInt(b, int64(offset), 10)
```
cbandy: I see that you're moving existing code, but I'm curious why we convert offset to int64 here?

johto: AppendInt takes an int64.  The code doesn't compile without the conversion.

cbandy: Ah yes. I should have looked.

johto:
Thank you, merged in e9da0370c75812f614163c0f93851c2902a820b8 and f709b902ce82147d238cd7c548d6857908e7c6be.

PR #328, opened 2015-01-22
Add test for timezone bug

e-dard:
This PR adds a test case to highlight a bug in how `lib/pq` is parsing date/time types that don't have timezone information.


e-dard:
This is for #329

PR #330, opened 2015-01-29
Add support for postgresql JSON type

supherman:


json/json.go
```diff
@@ -0,0 +1,44 @@
...
+	"database/sql/driver"
+	"encoding/json"
+	"errors"
+)
+
+type JSON struct {
+	Raw json.RawMessage
+}
+
+func (jv *JSON) Scan(value interface{}) error {
```
echlebek: How about io.Reader here?

json/json_test.go
```diff
@@ -0,0 +1,182 @@
...
+	jsonValue := JSON{}
+
+	// Test null values
+	err = db.QueryRow("SELECT null::json").Scan(&jsonValue)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if jsonValue.Raw != nil {
+		t.Fatalf("expected empty value")
```
cbandy: `Fatal ` here.

cbandy:
I'm not sure I understand what new feature this enables. I am able to scan into a `json.RawMessage` already (without this code.) The struct seems like unnecessary indirection.

I haven't personally tried using `json.RawMessage` as a Value, but I don't see anything here that... does anything?

supherman:
@cbandy you know what, you're right, I misunderstood how the scan method works, I can do the same thing just passing a []byte value and it's going to work, the good part is that lib/pq supports postgresql's json type out of the box, anyway I will close this, thanks for the quick feedback, fortunately this just took me 30 minutes.

cbandy:
I personally think it may still be worthwhile to have some tests that ensure/show `json.RawMessage` is a viable Value.

It seems like "the" type to represent... arbitrary JSON.

PR #331, opened 2015-01-30
Remove workaround for golang/go#3487

cbandy:
Fixed in Go 1.1


johto:
Nice catch.  Thank you.

PR #337, opened 2015-02-08
Speed up bytea encoding

cbandy:
before:
```
BenchmarkEncodeByteaHex  1000000              1606 ns/op             313 B/op          8 allocs/op
```

after:
```
BenchmarkEncodeByteaHex  5000000               386 ns/op              96 B/op          2 allocs/op
```

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -391,7 +391,10 @@ func parseBytea(s []byte) (result []byte) {
 func encodeBytea(serverVersion int, v []byte) (result []byte) {
 	if serverVersion >= 90000 {
 		// Use the hex format if we know that the server supports it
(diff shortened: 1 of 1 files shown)


cbandy:
Failure is tip + 9.4 ssl. Shall we make tip "okay to fail" in travis?

johto:
> Failure is tip + 9.4 ssl. Shall we make tip "okay to fail" in travis?

I'll have another look at the problem in a VM later today.  If I still can't reproduce, we can push such a change.

johto:
Big difference with a really small change.  Thanks, pushed in d6717c737b67b442e4d922235ade199af3a36002!

Issue #334, opened 2015-02-05
Possible SSL issue with tip and 9.4?


johto:
[This test](https://travis-ci.org/lib/pq/jobs/49568736) has now failed on two consecutive runs.  There have been recent changes to TLS in Go in tip, and IIRC there were some changes in 9.4 as well, so this might be legitimate.

johto:
Interestingly, I can't reproduce on OS X.

uhoh-itsmaciek:
Interesting. I restarted the Travis build, and it failed again the same way. Do you know what changes happened in 9.4?

johto:
It appears to be caused by ssl_prefer_server_ciphers being 'on'.  I'm not sure why it didn't break on OS X, though.  Will keep investigating..

johto:
OK, looks like a Go bug to me.  Opened https://github.com/golang/go/issues/9808.

johto:
Fixed via [ebe3d693d472f69cb68d945466eac0f770abf29b](https://github.com/golang/go/commit/ebe3d693d472f69cb68d945466eac0f770abf29b).

Issue #329, opened 2015-01-22
PG types without timezone being parsed into time.Times with.


e-dard:
Hi,

I briefly discussed this with @johto in `#postgresql`.

Currently, when a time-based PG type that doesn't have timezone information is parsed into a Go type, it's parsed into a `time.Time` that does contain timezone information. Specifically, the resulting `time.Time` contains a single `zone` in the `Location`, while the zero-value for a `time.Location` would contain an empty slice of `zones`, i.e., the absence of any timezones.

Given the `PG` value `'2015-01-13'::date` it should be equivalent to the following `time.Time`:

```go
time.Time{
        sec: 63556704000, 
        nsec: 0, 
        // zero value for a *time.Location
        loc: &time.Location{
                name: "", 
                zone: []time.zone(nil), 
                tx: []time.zoneTrans(nil), 
                cacheStart: 0, 
                cacheEnd: 0, 
                cacheZone: (*time.zone)(nil),
         },
}
```

Currently, `lib/pq` would return the following `time.Time`:

```go
time.Time{
    sec: 63556704000, 
    nsec: 0, 
    loc: &time.Location{
        name: "", 
        zone: []time.zone{
            time.zone{
                name: "", 
                offset: 0, 
                isDST: false,
            },
        }, 
        tx: []time.zoneTrans{
            time.zoneTrans{
                when: -9223372036854775808, 
                index: 0x0, 
                isstd: false, 
                isutc: false},
            },
        },
        cacheStart: -9223372036854775808, 
        cacheEnd: 9223372036854775807, 
        cacheZone:(*time.zone)(0xc20801eaa0),
    },
}
```

I'm don't think this is correct because the `time.Location` produced by `lib/pq` contains zone information, even if the zone is itself the zero-value.

I would be happy to put a fix in for this, but wanted to see if there was support for that idea or not?  I've added a PR with a test case highlighting the issue.

johto:
I think what we'd need to consider here are:

  1. How many applications would changing this realistically break
  2. What are the practical gains (meaning "outside of test cases") of changing this?  Are there, for example, any calculations which give the wrong (or at least a dubious) answer with the time.Time value returned now?

I can't claim to know the time package well enough to answer the second question without any research.

johto:
I've given this some more thought, and if this doesn't break any calculations, I think we should fix this.  It might break someone's test cases, but I think doing the correct think going forward is more important than that.

e-dard:
[...]

PR #338, opened 2015-02-10
Speed up timestamptz decoding

cbandy:
Before:
```
BenchmarkDecodeTimestamptz       1000000              1196 ns/op              64 B/op          2 allocs/op
BenchmarkDecodeTimestamptzMultiThread    1000000              1223 ns/op              64 B/op          2 allocs/op
```

After:
```
BenchmarkDecodeTimestamptz       3000000               473 ns/op              32 B/op          1 allocs/op
BenchmarkDecodeTimestamptzMultiThread    3000000               481 ns/op              32 B/op          1 allocs/op
```

This expects the input to be correctly formatted from the backend and asserts nothing. I'm not sure how I feel about that... Thoughts?

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -51,7 +51,7 @@ func decode(parameterStatus *parameterStatus, s []byte, typ oid.Oid) interface{}
 	case oid.T_bytea:
 		return parseBytea(s)
 	case oid.T_timestamptz:
-		return parseTs(parameterStatus.currentLocation, string(s))
+		return decodeTimestamptzISO(s, parameterStatus.currentLocation)
 	case oid.T_timestamp, oid.T_date:
 		return parseTs(nil, string(s))
 	case oid.T_time:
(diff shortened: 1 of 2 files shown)


cbandy:
I've included assertions, as they are certainly necessary. This implementation is twice as fast with half the allocations:

Before this PR:
```
BenchmarkDecodeTimestamptz       1000000              1202 ns/op              64 B/op          2 allocs/op
BenchmarkDecodeTimestamptzMultiThread    1000000              1191 ns/op              64 B/op          2 allocs/op
```

Without assertions:
```
BenchmarkDecodeTimestamptz       3000000               466 ns/op              32 B/op          1 allocs/op
BenchmarkDecodeTimestamptzMultiThread    3000000               484 ns/op              32 B/op          1 allocs/op
```

After this PR:
```
BenchmarkDecodeTimestamptz       3000000               519 ns/op              32 B/op          1 allocs/op
BenchmarkDecodeTimestamptzMultiThread    3000000               527 ns/op              32 B/op          1 allocs/op
```

uhoh-itsmaciek:
This looks good. It's unfortunate that this would lead to two very similar code paths for parsing time, but maybe that's worth it. My original code certainly wasn't optimized for performance. I'll merge this if there are no objections.

(oh, and I checked Travis and it's the same tip / 9.4 issue as described in #334)

johto:
How much can we speed up parseTs() if we avoid the second allocation?  Where is it coming from in the first place?  It's not obvious from just looking at the code.

I'm a bit worried about having this much code to maintain just for parsing timestamps.

cbandy:
> How much can we speed up parseTs()

Since parseTs() is used only for `timestamp` and `date`, code there around offset/location can probably be jettisoned.

> the second allocation? Where is it coming from in the first place?

Perhaps the conversion to string? This version operates on the byte slice. We also called out to `strconv.Atoi`, which uses the generic `ParseUint`.

> having this much code to maintain just for parsing timestamps

In general, or because I added a function instead of touching parseTs? I tried using [time.Parse more heavily](https://gist.github.com/cbandy/77c104ff37d9c8a1ea7a), but it was slower than parseTs.

> two very similar code paths for parsing time

Parsing other time-ish types is being debated in #329. I focused on `timestamptz` since it seemed to be immune to the debate.

cbandy:
I've added one more commit that I think makes this function easier to read.

```
HEAD is now at 08ad791... Assert expected bytes as they are processed
BenchmarkDecodeTimestamptz       3000000               514 ns/op              32 B/op          1 allocs/op
BenchmarkDecodeTimestamptzMultiThread    3000000               528 ns/op              32 B/op          1 allocs/op

HEAD is now at 013da70... Reduce some duplication when decoding timestamptz
BenchmarkDecodeTimestamptz       3000000               492 ns/op              32 B/op          1 allocs/op
BenchmarkDecodeTimestamptzMultiThread    3000000               521 ns/op              32 B/op          1 allocs/op
```

johto:
>> having this much code to maintain just for parsing timestamps

> In general, or because I added a function instead of touching parseTs? I tried using time.Parse more heavily, but it was slower than parseTs.

I don't mind not using time.Parse, but I wouldn't want two distinct hand-written routines for parsing time-related data.  Naively thinking, we'd have one function which:

  1. Parses a date in ISO format.  If the data type is "date", the end of string must be nearby as well.
  2. Parses a time.  If the data type is "without time zone", the end of string must be there.
  3. Parses the UTC offset information.  This must be the end of string.

If we can make that function fast while supporting all three type, things ought to be pretty good.

cbandy:
> I wouldn't want two distinct hand-written routines for parsing time-related data.

I agree. This PR should probably wait for a resolution to #329.

PR #340, opened 2015-02-19
Reset deadline on TCP connection only if a one had been set

minusnine:
Prior to this change, the connection logic would unconditionally reset
the deadline on the TCP connection, even if a connection timeout hadn't
been specified. This prevented a TCP connection tunnelled through an SSH
session, as those connections do not support SetDeadline.

Now, only call SetDeadline if a connection timeout had been set.


johto:
Looks good, pushed.  Thanks!

Issue #196, opened 2013-12-03
Float precision loss


MattNewberry:
When inserting a number with higher precision than 6 decimal places (for example, longitude of `35.03554004971999`) the precision past the 6th place is lost. The field type is `float64` and has a data type in Postgres as `numeric`.

Example:

```
package main

import (
        "database/sql"
        "fmt"
        _ "github.com/lib/pq"
)

func main() {
        db, _ := sql.Open("postgres", "user=gorm dbname=gorm sslmode=disable")
        sql := `INSERT INTO users (latitude) VALUES ($1) RETURNING "id"`
        res, err := db.Exec(sql, 35.03554004971999)
        fmt.Println(res, err)
}
```

PR #351, opened 2015-03-15
Test that Exec and Query are only called with certain types

cbandy:
Because `driver.Value` is implemented as an `interface{}`, its value can actually be any type. It is the responsibility of the `sql` and `driver` packages to emit [only a few types](http://golang.org/pkg/database/sql/driver/#Value) for these values.

This test ensures those packages are fulfilling their contract in cases we care about.


johto:
What was the conclusion with the float32 handling in encode()?  Is it dead code?

cbandy:
Yes, it's dead.

cbandy:
@johto would you like me to include a commit that removes the dead code?

johto:
@cbandy: Please do.

cbandy:
I've added a commit that removes code for encoding `float32` values. The tests for #196, #197, and #198 remain and pass.

johto:
> I've added a commit that removes code for encoding float32 values. The tests for #196, #197, and #198 remain and pass.

Thanks!  I'll try and have a look at this later today.

cbandy:
@johto I think this (or a test like this) is still valuable to include in our suite. What would you like me to do with this PR?

johto:
> @johto I think this (or a test like this) is still valuable to include in our suite. What would you like me to do with this PR?

I agree, but I haven't had a chance to review it yet.  Sorry about that.

PR #355, opened 2015-03-20
remove unreachable code (return after panic)

dmitris:
flagged by go tool vet:
```
$ go tool vet .
notify_test.go:47: unreachable code
```


johto:
Thanks, merged.

PR #358, opened 2015-04-01
Disable SSL when connecting over a UNIX domain socket

johto:
Any objections to applying the following?  It's silly to require the user to set sslmode=disable for UNIX domain connections, and it's not compatible with libpq's behavior.


erykwalder:
:+1: I have to set that quite frequently

johto:
Never again.  a33d6053e025943d5dc89dfa1f35fe5500618df7 is here.

PR #362, opened 2015-04-11
Allow using CopyIn in table associated with triggers that use raise stmt

ympons:
Hello,
  Thanks a lot for this package. We're using CopyIn in [Urban4m](http://www.urban4m.com) and it's really fast. When we use this feature in tables associated with triggers that use raise statement to report messages, the package throws the error "unknown response during CopyIn: N". To avoid it, we can add this case as valid response into resploop function.

### DIFF
--- a/copy_test.go
+++ b/copy_test.go
@@ -115,9 +115,8 @@ func TestCopyInRaiseStmtTrigger(t *testing.T) {
 			$BODY$ begin
 				raise notice 'Hello world';
 				return new;
-			end; $BODY$ 
-			LANGUAGE plpgsql
-		`)
+			end $BODY$
+			LANGUAGE plpgsql`)
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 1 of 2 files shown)


johto:
This does seem like an omission, but it should be a separate case in the switch.  The comment "complete" doesn't apply at all.

copy_test.go
```diff
@@ -94,6 +94,85 @@ func TestCopyInMultipleValues(t *testing.T) {
...
+	defer txn.Rollback()
+
+	_, err = txn.Exec("CREATE TEMP TABLE temp (a int, b varchar)")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	//Check if test function exists 
+	var num int
+	err = txn.QueryRow(`SELECT COUNT(*) FROM pg_proc where proname = 'temptest'`).Scan(&num)
```
johto: You should instead create the function here.  You can create functions in the pg_temp schema, and they will act like temporary tables.

ympons:
@johto I agree. Thanks!

copy_test.go
ympons: @johto I agree. Thanks!

johto:
I've pushed this in 35aad2043dc9040a6d1f1f6ad264fb4377b1e551 after fixing the tests for 8.4.  Thanks!

Issue #327, opened 2015-01-21
Array Support


erykwalder:
I've been browsing the issues and PRs to see what the status of array support is. Along with a bunch of others, support is something I'd like to see.

I'm curious as to what the consensus is as of now. Would it be nice for support to be baked into this package, or should it come from another package? Should it affect the encode/decode process, or simply be a set of scanners?

To speak more generally, what should be supported, what cases should be handled, and what should be avoided?

This is something I'm willing to work on; I just want to know what will be mergable.

johto:
> I'm curious as to what the consensus is as of now. Would it be nice for support to be baked into this package, or should it come from another package? Should it affect the encode/decode process, or simply be a set of scanners?
>
> To speak more generally, what should be supported, what cases should be handled, and what should be avoided?

I've made some of my thoughts public in #302.  I'd very much like to have array support built into the package so that users don't have to keep reinventing the wheel.

But the most important thing would be to first come out with a detailed proposal on how the interface would look like.  Once the interface has been agreed on, it shouldn't be more than a small matter of programming.

tarikjn:
Hey @johto, what is the current state of that interface? What is the best proposal so far and what is missing? Also, how about composite/row types?

johto:
@tarikjn: I have not seen an interface addressing my comments in #302.  Perhaps at this point I should just write an interface proposal I like, and then have everyone else shoot it down.  If we ever reach a point where nobody is too unhappy with it, someone can write the code.  (That I won't be doing any time soon).

Re: composite/row types, I don't remember them having been mentioned before at any point.  I haven't spent any time thinking about what their interface should look like, either.  If you have time to spend on this, feel free to try and come up with something.  I'll be happy to review an interface proposal.

cbandy:
@johto I've read #302, but I wasn't able to follow the arguments (possibly because the PR diff has gone away.) The `Valuer` [you demonstrated](https://gist.github.com/johto/2914cd7309dedc5fb0d8) seems like the right approach to me:

```go
type IntArray []int
func (IntArray) Value() (driver.Value, error) {}
```

`Valuer`s and `Scanner`s seem like the only appropriate way to do this, iiuc. There are only a few Go types we're allowed to return from the driver...

erykwalder:
I'm thinking there should just be one array type that wraps an interface, and can Scan/Value into that interface, using something like the Unmarshaller in #302.

Having separate wrappers per slice type seems less than ideal.

[...]

PR #364, opened 2015-04-22
Array valuer with delimiter interface

cbandy:
Another solution for #327. Similar to #353, but includes an interface for the delimiter character.

### DIFF
--- a/array.go
+++ b/array.go
@@ -12,26 +12,6 @@ import (
 
 var typeByteSlice = reflect.TypeOf([]byte{})
 
-// Array returns the optimal driver.Valuer for an array or slice of any
-// dimension.
-//
-// For example:
-//  db.Query(`SELECT * FROM t WHERE id = ANY($1)`, pq.Array([]int{235, 401}))
-func Array(a interface{}) driver.Valuer {
-	switch a := a.(type) {
-	case []bool:
-		return BoolArray(a)
-	case []float64:
-		return Float64Array(a)
-	case []int64:
-		return Int64Array(a)
-	case []string:
-		return StringArray(a)
-	}
-
-	return GenericArray{a}
-}
-
 // ArrayValuer may be optionally implemented by driver.Valuer to override the
 // array delimiter used by GenericArray.
 type ArrayValuer interface {
--- a/array.go
+++ b/array.go
@@ -1,0 +1,124 @@
+package pq
+
+import (
+	"bytes"
+	"database/sql/driver"
+	"fmt"
+	"reflect"
+	"strings"
+)
+
+// Array implements the driver.Valuer interface for an array or slice.
+//
+// For example:
+//  db.Query("SELECT * FROM t WHERE id = ANY($1)", pq.Array{[]int{235, 401}})
+type Array struct{ A interface{} }
+
+// Value implements the driver.Valuer interface.
+func (a Array) Value() (driver.Value, error) {
+	if a.A == nil {
+		return nil, nil
+	}
+
+	rv := reflect.ValueOf(a.A)
+
+	if k := rv.Kind(); k != reflect.Array && k != reflect.Slice {
+		return nil, fmt.Errorf("pq: Unable to convert %T to array", a.A)
+	}
+
+	b, _, err := appendArray(nil, rv)
+	return b, err
+}
+
+// appendArray appends rv to the buffer, returning the extended buffer and
+// the delimiter used between elements.
+//
+// It panics if rv's Kind is not reflect.Array nor reflect.Slice.
+func appendArray(b []byte, rv reflect.Value) ([]byte, string, error) {
+	n := rv.Len()
+	if n == 0 {
+		return append(b, "{}"...), "", nil
+	}
+
+	// TODO if b == nil { b = make([]byte, 0, /* best guess */) }
+
+	b = append(b, '{')
+
+	var del string
+	var err error
+	if b, del, err = appendArrayElement(b, rv.Index(0)); err != nil {
+		return b, del, err
+	}
+
+	for i := 1; i < n; i++ {
+		b = append(b, del...)
+		if b, del, err = appendArrayElement(b, rv.Index(i)); err != nil {
+			return b, del, err
+		}
+	}
+
+	return append(b, '}'), del, nil
+}
+
+// appendArrayElement appends rv to the buffer, returning the extended buffer
+// and the delimiter to use before the next element.
+//
+// When rv's Kind is neither reflect.Array nor reflect.Slice, it is converted
+// using driver.DefaultParameterConverter and the resulting []byte or string
+// is double-quoted.
+//
+// See http://www.postgresql.org/docs/current/static/arrays.html#ARRAYS-IO
+func appendArrayElement(b []byte, rv reflect.Value) ([]byte, string, error) {
+	if k := rv.Kind(); k == reflect.Array ||
+		(k == reflect.Slice && rv.Type() != reflect.TypeOf([]byte{})) {
+		return appendArray(b, rv)
+	}
+
+	var del string = ","
+	var err error
+	var iv interface{} = rv.Interface()
+
+	if iv, err = driver.DefaultParameterConverter.ConvertValue(iv); err != nil {
+		return b, del, err
+	}
+
+	switch v := iv.(type) {
+	case nil:
+		return append(b, "NULL"...), del, nil
+
+	case []byte:
+		b = append(b, '"')
+		for {
+			i := bytes.IndexAny(v, `"\`)
+			if i < 0 {
+				b = append(b, v...)
+				break
+			}
+			b = append(b, v[:i]...)
+			b = append(b, '\\', v[i])
+			v = v[i+1:]
+		}
+		return append(b, '"'), del, nil
+
+	case string:
+		b = append(b, '"')
+		for {
+			i := strings.IndexAny(v, `"\`)
+			if i < 0 {
+				b = append(b, v...)
+				break
+			}
+			b = append(b, v[:i]...)
+			b = append(b, '\\', v[i])
+			v = v[i+1:]
+		}
+		return append(b, '"'), del, nil
+	}
+
+	b, err = appendValue(b, iv)
+	return b, del, err
+}
+
+func appendValue(b []byte, v driver.Value) ([]byte, error) {
+	return append(b, encode(nil, v, 0)...), nil
+}
--- a/array.go
+++ b/array.go
@@ -46,10 +46,10 @@ func (a BoolArray) Value() (driver.Value, error) {
 		b[0] = '{'
 		b[2*n] = '}'
 
-		return b, nil
+		return string(b), nil
 	}
 
-	return []byte{'{', '}'}, nil
+	return "{}", nil
 }
 
 // ByteaArray represents a one-dimensional array of the PostgreSQL bytea type.
@@ -82,10 +82,10 @@ func (a ByteaArray) Value() (driver.Value, error) {
 		b[0] = '{'
 		b[size-1] = '}'
 
-		return b, nil
+		return string(b), nil
 	}
 
-	return []byte{'{', '}'}, nil
+	return "{}", nil
 }
 
 // Float64Array represents a one-dimensional array of the PostgreSQL double
@@ -110,10 +110,10 @@ func (a Float64Array) Value() (driver.Value, error) {
 			b = strconv.AppendFloat(b, a[i], 'f', -1, 64)
 		}
 
-		return append(b, '}'), nil
+		return string(append(b, '}')), nil
 	}
 
-	return []byte{'{', '}'}, nil
+	return "{}", nil
 }
 
 // GenericArray implements the driver.Valuer interface for an array or slice
@@ -138,10 +138,10 @@ func (a GenericArray) Value() (driver.Value, error) {
 		b := make([]byte, 0, 1+2*n)
 
 		b, _, err := appendArray(b, rv, n)
-		return b, err
+		return string(b), err
 	}
 
-	return []byte{'{', '}'}, nil
+	return "{}", nil
 }
 
 // Int64Array represents a one-dimensional array of the PostgreSQL integer types.
@@ -165,10 +165,10 @@ func (a Int64Array) Value() (driver.Value, error) {
 			b = strconv.AppendInt(b, a[i], 10)
 		}
 
-		return append(b, '}'), nil
+		return string(append(b, '}')), nil
 	}
 
-	return []byte{'{', '}'}, nil
+	return "{}", nil
 }
 
 // StringArray represents a one-dimensional array of the PostgreSQL character types.
@@ -192,10 +192,10 @@ func (a StringArray) Value() (driver.Value, error) {
 			b = appendArrayQuotedString(b, a[i])
 		}
 
-		return append(b, '}'), nil
+		return string(append(b, '}')), nil
 	}
 
-	return []byte{'{', '}'}, nil
+	return "{}", nil
 }
 
 // appendArray appends rv to the buffer, returning the extended buffer and
--- a/array.go
+++ b/array.go
@@ -21,10 +21,6 @@ type ArrayValuer interface {
 
 	// ArrayDelimiter returns the delimiter character(s) for this element's type.
 	ArrayDelimiter() string
-
-	// ArrayValue returns the value of this array element. []byte and string
-	// values are double-quoted by Array.
-	ArrayValue() (interface{}, error)
 }
 
 // Value implements the driver.Valuer interface.
@@ -98,9 +94,6 @@ func appendArrayElement(b []byte, rv reflect.Value) ([]byte, string, error) {
 
 	if av, ok := iv.(ArrayValuer); ok {
 		del = av.ArrayDelimiter()
-		if iv, err = av.ArrayValue(); err != nil {
-			return b, del, err
-		}
 	}
 
 	if iv, err = driver.DefaultParameterConverter.ConvertValue(iv); err != nil {
--- a/array.go
+++ b/array.go
@@ -3,6 +3,7 @@ package pq
 import (
 	"bytes"
 	"database/sql/driver"
+	"encoding/hex"
 	"fmt"
 	"reflect"
 	"strconv"
@@ -62,6 +63,41 @@ func (a BoolArray) Value() (driver.Value, error) {
 
 		b[0] = '{'
 		b[2*n] = '}'
+
+		return b, nil
+	}
+
+	return []byte{'{', '}'}, nil
+}
+
+type ByteaArray [][]byte
+
+// Value implements the driver.Valuer interface. It uses the "hex" format which
+// is only supported on PostgreSQL 9.0 or newer.
+func (a ByteaArray) Value() (driver.Value, error) {
+	if a == nil {
+		return nil, nil
+	}
+
+	if n := len(a); n > 0 {
+		// There will be at least two curly brackets, 2*N bytes of quotes,
+		// 3*N bytes of hex formatting, and N-1 bytes of delimiters.
+		size := 1 + 6*n
+		for _, x := range a {
+			size += hex.EncodedLen(len(x))
+		}
+
+		b := make([]byte, size)
+
+		for i, s := 0, b; i < n; i++ {
+			o := copy(s, `,"\\x`)
+			o += hex.Encode(s[o:], a[i])
+			s[o] = '"'
+			s = s[o+1:]
+		}
+
+		b[0] = '{'
+		b[size-1] = '}'
 
 		return b, nil
 	}
@@ -158,7 +194,7 @@ func (a StringArray) Value() (driver.Value, error) {
 	}
 
 	if n := len(a); n > 0 {
-		// There will be at least two curly brackets, 2*N bytes of values,
+		// There will be at least two curly brackets, 2*N bytes of quotes,
 		// and N-1 bytes of delimiters.
 		b := make([]byte, 1, 1+3*n)
 		b[0] = '{'
--- a/array.go
+++ b/array.go
@@ -37,22 +37,21 @@ func (a Array) Value() (driver.Value, error) {
 		return nil, fmt.Errorf("pq: Unable to convert %T to array", a.A)
 	}
 
-	b, _, err := appendArray(nil, rv)
-	return b, err
+	if n := rv.Len(); n > 0 {
+		// TODO b = make([]byte, 0, /* best guess */)
+
+		b, _, err := appendArray(nil, rv, n)
+		return b, err
+	}
+
+	return []byte{'{', '}'}, nil
 }
 
 // appendArray appends rv to the buffer, returning the extended buffer and
 // the delimiter used between elements.
 //
-// It panics if rv's Kind is not reflect.Array nor reflect.Slice.
-func appendArray(b []byte, rv reflect.Value) ([]byte, string, error) {
-	n := rv.Len()
-	if n == 0 {
-		return append(b, "{}"...), "", nil
-	}
-
-	// TODO if b == nil { b = make([]byte, 0, /* best guess */) }
-
+// It panics when n <= 0 or rv's Kind is not reflect.Array nor reflect.Slice.
+func appendArray(b []byte, rv reflect.Value, n int) ([]byte, string, error) {
 	b = append(b, '{')
 
 	var del string
@@ -82,7 +81,11 @@ func appendArray(b []byte, rv reflect.Value) ([]byte, string, error) {
 func appendArrayElement(b []byte, rv reflect.Value) ([]byte, string, error) {
 	if k := rv.Kind(); k == reflect.Array ||
 		(k == reflect.Slice && rv.Type() != reflect.TypeOf([]byte{})) {
-		return appendArray(b, rv)
+		if n := rv.Len(); n > 0 {
+			return appendArray(b, rv, n)
+		}
+
+		return b, "", nil
 	}
 
 	var del string = ","
--- a/array.go
+++ b/array.go
@@ -14,6 +14,17 @@ import (
 //  db.Query("SELECT * FROM t WHERE id = ANY($1)", pq.Array{[]int{235, 401}})
 type Array struct{ A interface{} }
 
+// ArrayValuer is an interface used by Array.
+type ArrayValuer interface {
+
+	// ArrayDelimiter returns the delimiter character(s) for this element's type.
+	ArrayDelimiter() string
+
+	// ArrayValue returns the value of this array element. []byte and string
+	// values are double-quoted by Array.
+	ArrayValue() (interface{}, error)
+}
+
 // Value implements the driver.Valuer interface.
 func (a Array) Value() (driver.Value, error) {
 	if a.A == nil {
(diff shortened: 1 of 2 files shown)


johto:
Pardon me if this is a stupid question, but why do we need to support arbitrary delimiters?

cbandy:
> why do we need to support arbitrary delimiters?

The array delimiter is an [attribute of the type](http://www.postgresql.org/docs/current/static/arrays.html#ARRAYS-INPUT). While most types have a comma delimiter, some do not (e.g. `box` has a semicolon delimiter.)

I thought this would be a convenient way for a `driver.Valuer` to also leverage `pq.Array`. For example,

```go
type Box struct { ... }
func (b Box) Value() (driver.Value, error) { ... }
func (b Box) ArrayDelimiter() string { return ";" }
func (b Box) ArrayValue() (driver.Value, error) { return b.Value() }

db.Query(`SELECT $1::box[]`, pq.Array{[]Box{ ... }})

```

(Looking at it now, `ArrayValue()` seems redundant. I'll find out if that is technically true.)

cbandy:
`ArrayValue` was redundant.

cbandy:
As @erykwalder demonstrated, the cost of reflection is significant:

```
$ go test -bench 'ArrayValue' -benchmem
PASS
BenchmarkBoolArrayValue          5000000               278 ns/op              64 B/op          2 allocs/op
BenchmarkFloat64ArrayValue        300000              4906 ns/op             784 B/op          6 allocs/op
BenchmarkInt64ArrayValue          500000              2446 ns/op             784 B/op          6 allocs/op
BenchmarkStringArrayValue         100000             16939 ns/op            2048 B/op          7 allocs/op
BenchmarkGenericArrayValueBools           500000              3429 ns/op             368 B/op         24 allocs/op
BenchmarkGenericArrayValueFloat64s        200000             10172 ns/op            1392 B/op         45 allocs/op
BenchmarkGenericArrayValueInt64s          300000              5894 ns/op            1184 B/op         26 allocs/op
BenchmarkGenericArrayValueByteSlices      100000             15652 ns/op            1872 B/op         17 allocs/op
BenchmarkGenericArrayValueStrings         100000             18311 ns/op            1712 B/op         17 allocs/op
```

Also, like @erykwalder, I have come up against choosing escape or hex format to (or to not) support arrays of `bytea`. My current inclination is to encode to the hex format only, ignoring support for 8.4 or earlier.

johto:
> As @erykwalder demonstrated, the cost of reflection is significant:

That was quite expected.

> Also, like @erykwalder, I have come up against choosing escape or hex format to (or to not) support arrays of bytea. My current inclination is to encode to the hex format only, ignoring support for 8.4 or earlier.

That sounds reasonable to me.  8.4 has been unsupported for a while, anyway.

johto:
> The array delimiter is an attribute of the type. While most types have a comma delimiter, some do not (e.g. box has a semicolon delimiter.)

Oh, right.  That exception.

OK.  I'll give this a review over the weekend.

cbandy:
I'll try to add `ByteaArray` tonight.

pkieltyka:
hey guys, thanks for the work on this - what about the array scanners tho?

cbandy:
@pkieltyka, it's probably not too difficult to do this for the `[]bool`, `[][]byte`, `[]float64`, `[]int64`, and `[]string` types here. I may explore that in another PR.

However, I'm not sure what it would take to support multidimensional arrays or arrays/slices of `Scanner`s. I'll think about it.

johto:
@cbandy: Do you think this is in a committable state right now?  It's missing documentation, but perhaps we should commit this interface as an "experimental" feature to try and get some more feedback on it.

cbandy:
It needs an integration test--something that asserts the backend agrees. I'd also like the documentation to be adequate before merge. Can you please suggest some ways to improve it?

I can address both in the coming days.

johto:
> It needs an integration test--something that asserts the backend agrees.

All right.

> I'd also like the documentation to be adequate before merge. Can you please suggest some ways to improve it?

Uh.  This is going to sound stupid, but I completely forgot that Go projects don't put documentation in a dedicated set of files (which was what I was looking for).  That said, I'll be happy to do a pass over them once I have an environment where I can see what the godoc looks like.  (Is there a way to make godoc do that for github PRs?  That could be awesome.)

piotrkowalczuk:
Since main discusion is right here i decided to move my comment in here.
In my opinion providing types like `sql.NullInt64` is most idiomatic and cover 90% of cases. For example:

**type.go**
```go
package postgres

const (
	// ArraySeparator ...
	ArraySeparator = ", "
	// ArrayBeginningChar ...
	ArrayBeginningChar = "{"
	// ArrayEndChar ...
	ArrayEndChar = "}"
)
```
**array_int64.go**
```go
package postgres

import (
	"bytes"
	"database/sql/driver"
	"errors"
	"strconv"
	"strings"
)

var (
	//ErrArrayInt64UnsupportedSourceType  ...
	ErrArrayInt64UnsupportedSourceType = errors.New("pq: array int64 unsupported source type")
	// ErrArrayInt64UnsupportedSourceFormat ...
	ErrArrayInt64UnsupportedSourceFormat = errors.New("pq: array int64 unsupported source format")
	// ErrArrayInt64UnsupportedValueType ...
	ErrArrayInt64UnsupportedValueType = errors.New("pq: array int64 unsupported value type")
)

// ArrayInt64 is a slice of int64s that implements necessary interfaces.
type ArrayInt64 []int64

// Scan satisfy sql.Scanner interface.
func (a *ArrayInt64) Scan(src interface{}) error {
	var tmp []string
	var srcs string

	switch t := src.(type) {
	case []byte:
		srcs = string(t)
	case string:
		srcs = t
	default:
		return ErrArrayInt64UnsupportedSourceType
	}

	l := len(srcs)

	if l < 2 {
		return ErrArrayInt64UnsupportedSourceFormat
	}

	if string(srcs[0]) != ArrayBeginningChar || string(srcs[l-1]) != ArrayEndChar {
		return ErrArrayInt64UnsupportedSourceFormat
	}

	tmp = strings.Split(string(srcs[1:l-1]), ArraySeparator)

	for _, v := range tmp {
		i, err := strconv.ParseInt(v, 10, 64)
		if err != nil {
			return ErrArrayInt64UnsupportedValueType
		}

		*a = append(*a, i)
	}

	return nil
}

// Value satisfy driver.Valuer interface.
func (a ArrayInt64) Value() (driver.Value, error) {
	var buffer bytes.Buffer

	buffer.WriteString(ArrayBeginningChar)

	for i, v := range a {
		if i > 0 {
			_, err := buffer.WriteString(ArraySeparator)
			if err != nil {
				return nil, err
			}
		}
		_, err := buffer.WriteString(strconv.FormatInt(v, 10))
		if err != nil {
			return nil, err
		}
	}

	buffer.WriteString(ArrayEndChar)

	return buffer.Bytes(), nil
}
```
**array_int64_test.go**
```go
package postgres

import (
	"testing"

	"github.com/stretchr/testify/assert"
)

func TestArrayInt64Value(t *testing.T) {
	success := map[string]ArrayInt64{
		"{1, 2, 3, 4}": ArrayInt64{1, 2, 3, 4},
		"{}":           ArrayInt64{},
	}

	for expected, array := range success {
		got, err := array.Value()
		if assert.NoError(t, err) {
			assert.Equal(t, []byte(expected), got)
		}
	}
}

func TestArrayInt64Scan(t *testing.T) {
	success := map[string]ArrayInt64{
		"{1, 2, 3, 4}": ArrayInt64{1, 2, 3, 4},
	}

	for src, expected := range success {
		var got ArrayInt64
		err := got.Scan(src)
		if assert.NoError(t, err) {
			assert.Equal(t, expected, got)
		}
	}

	fail := map[string]error{
		"{string1, string2}": ErrArrayInt64UnsupportedValueType,
		"}":                  ErrArrayInt64UnsupportedSourceFormat,
		"{":                  ErrArrayInt64UnsupportedSourceFormat,
		"{12412s}":           ErrArrayInt64UnsupportedValueType,
	}

	for src, expected := range fail {
		var got ArrayInt64
		err := got.Scan(src)
		assert.EqualError(t, expected, err.Error())
	}
}
```

Ofc i dont see anything wrong to provide proxy like functions. Overall API in my opinion could looks like:
```go
type ArrayInt64 []int64
type ArrayFloat64 []float64
type ArrayString []string
type ArrayBool []bool
type ArrayBytes [][]byte

func FromArray(interface{}) driver.Valuer {}
func ToArray(interface{}) io.Scanner {}
```

Idea of  `GenericArray` is a little bit odd. I cant imagin situation where it can be usefull. If I understand correctly only reason why its there is because `default` case is not an option. 
Instead of that i can suggest smth like that:

```go
type errorValuer error

func (ev errorValuer) Value() (driver.Value, error) {
        return nil, ev
}

func Array(a interface{}) driver.Valuer {
	switch a := a.(type) {
	case []bool:
		return BoolArray(a)
	case []float64:
		return Float64Array(a)
	case []int64:
		return Int64Array(a)
	case []string:
		return StringArray(a)
        default:
                return errorValuer(errors.New("pq: unsupported data type")) // message could be better
	}
}
```

Example usage for more complex situation could be:

```go
type Custom pq.ArrayInt64

// or 

type MoreCustom []int64

func (mc MoreCustom) Value() (driver.Value, error) {
    return pq.ArrayInt64(mc).Value()
}
```

johto:
@piotrkowalczuk wrote:
> Ofc i dont see anything wrong to provide proxy like functions. Overall API in my opinion could looks like:
> 
> type ArrayInt64 []int64
> [ .. etc .. ]
> 
> func FromArray(interface{}) driver.Valuer {}

I don't have any strong objections to providing the FromArray "auto-choose" wrapper here.  Not sure which version I'd use myself.

> Idea of GenericArray is a little bit odd. I cant imagin situation where it can be usefull. If I understand correctly only reason why its there is because default case is not an option. 
> Instead of that i can suggest smth like that:
> return errorValuer(errors.New("pq: unsupported data type")) // message could be better

It's useful in the case where you have an array where the element is *any* type the driver does not know about.  For example, assume you have an array of UUIDs and you want to fetch those from the database.  You'd do something like:

```go
// pseudo-code, since I don't have a compiler anywhere near me
uuidArray := … // whatever
rows, err := db.Query(`SELECT * FROM $1::box[]`, pq.Array{uuidArray})
// ...
```

I don't see any case here where limiting what the user is allowed to do would be beneficial for us, or the user.  Just sending over an array of the string representation of the elements would Just Work, without the user having to mess around with custom types and what not.

cbandy:
>  (Is there a way to make godoc do that for github PRs? That could be awesome.)

Not that I can see. It seems to find the code from the import path the same way as `go get`.

cbandy:
> Idea of GenericArray is a little bit odd. I cant imagin situation where it can be usefull. If I understand correctly only reason why its there is because default case is not an option. 
Instead of that i can suggest smth like that:

> return errorValuer(errors.New("pq: unsupported data type")) // message could be better

> > It's useful in the case where you have an array where the element is any type the driver does not know about.

This includes something as simple as `[]int`. When one calls `db.Query()` with an `int`, `uint`, `byte`, or `rune`, the `sql` package converts it to `int64` before passing it to the `pq` package. The `pq` package doesn't have code to handle `int` arguments, but the user expects them to work. I think users would expect slices of `int`, `uint`, etc to behave in a similar way.

`GenericArray` also supports nested or multidimensional arrays (see SQL:2003.) For example, a matrix of integers could be held in `[][]int` in Go, passed to `sql`/`pq` via `GenericArray`, and stored in `int[][]` in Postgres. I think our "array" feature must include this.

johto:
@cbandy wrote:
> This includes something as simple as []int. When one calls db.Query() with an int, uint, byte, or rune, the sql package converts it to int64 before passing it to the pq package. The pq package doesn't have code to handle int arguments, but the user expects them to work. I think users would expect slices of int, uint, etc to behave in a similar way.

Good point.

johto:
I'm going to look into committing this as an "experimental" feature next.  I think it'd be nice to have both Values and Scanners committed while still reserving the "right" to make changes to the interface, in case we find something that doesn't quite work as initially committed.

cbandy:
8.4 and 9.0 failed in cases that include slash. I suspect standard_conforming_strings. :suspect:

erykwalder:
@cbandy Is there anything left to get this merged?

pkieltyka:
Where is the line drawn on minimum server version for pq?

uhoh-itsmaciek:
@pkieltyka we probably want to support all the releases [supported by the server upstream](http://www.postgresql.org/support/versioning/). That means we could drop 8.4 from `.travis.yml`, but 9.0 is supported until September.

However, `standard_conforming_strings` is meaningful beyond those versions, so ideally any parameter encoding would take that into account insofar as it affects this...

uhoh-itsmaciek:
(and I think a2e94b1d77e1194a8ecb7294592b221179386ad8 is a perfectly reasonable way of doing that)

pkieltyka:
I completely agree, therefore we shouldnt care about 8.4. As well, IMO, 9.0 is as good as here.. aka, by the time this PR is merged, it will be September.

johto:
> @pkieltyka we probably want to support all the releases supported by the server upstream. That means we could drop 8.4 from .travis.yml, but 9.0 is supported until September

I don't think we need to drop it until there's a really good reason to.  I'm fine with not supporting 8.4 for all of the new functionality if it's difficult or impossible, but if we stop testing in Travis things will probably break for no real reason.

pkieltyka:
Yea I agree, for new functionality there's no need to worry about < 9.0 ... But I guess if Travis complains, then it'll feel like something is broken that doesn't need to be supported... You'd have to skip certain test cases for older dbs by checking db version while tests are running

johto:
> But I guess if Travis complains, then it'll feel like something is broken that doesn't need to be supported... You'd have to skip certain test cases for older dbs by checking db version while tests are running

We already do that.

cbandy:
> @cbandy Is there anything left to get this merged?

The documentation is too sparse in my opinion. Each of the array types have only a single sentence, `Value implements the driver.Valuer interface.`, and the `Array` function seems vague as well. I'm open to suggestions here.

@johto mentioned that "it'd be nice to have both Valuers and Scanners committed," and I think I agree. It feels odd to not support both the sending and receiving of a type. I think we should ditch the `Array` function (and its Valuer return type) until we have implemented Scanners as well.

I'm okay with merging these Valuers once these two things are addressed.

cbandy:
> I'm okay with merging these Valuers once these two things are addressed.

Added two commits.

array.go
```diff
@@ -0,0 +1,302 @@
...
+		return nil, nil
+	}
+
+	if n := len(a); n > 0 {
+		// There will be at least two curly brackets, 2*N bytes of quotes,
+		// and N-1 bytes of delimiters.
+		b := make([]byte, 1, 1+3*n)
+		b[0] = '{'
+
+		b = appendArrayQuotedString(b, a[0])
```
johto: Is there a reason we're not doing `appendArrayQuotedBytes(b, []byte(a[0]))` here that I'm not seeing?  We're screwed anyway if the input is not UTF-8.

johto:
Hrmph.  The generic array valuer seems to be having a bit of a problem with types where the underlying type is a slice.  For example:

```go
package main

import (
    "github.com/lib/pq"
    "database/sql"
    "fmt"
    "database/sql/driver"
)

type uuid []byte
// driver.Valuer
func (u uuid) Value() (driver.Value, error) {
    return "618705ec-3dec-4f39-8a36-5b1dd69ab7db", nil
}

func main() {
    db, err := sql.Open("postgres", "host=/tmp")
    if err != nil {
        panic(err)
    }

    a := []uuid{uuid{},uuid{}}
    fmt.Printf("len: %d\nvalue: %+#v\n", len(a), a)
    var l int
    err = db.QueryRow("SELECT count(*) FROM unnest($1::uuid[])", pq.Array(a)).Scan(&l)
    if err != nil {
        panic(err)
    }
    fmt.Printf("in-db len: %d\n", l)
}
```

prints:

```
len: 2
value: []main.uuid{main.uuid{}, main.uuid{}}
in-db len: 0
```

because appendArrayElement considers its kind to be slice.  I'm kind of tempted to solve this by doing the following instead:

```go
    _, implementsValuer := rv.Interface().(driver.Valuer)                                                 
    if k := rv.Kind(); !implementsValuer && (k == reflect.Array || (k == reflect.Slice && rv.Type() != typeByteSlice)) {
```

Which allows the type implementation to bypass the multidimensional array handling code by implementing driver.Valuer.  Or does anyone have any better ideas?  I don't see any straightforward way of distinguishing the two using the reflect package, but I might as well be missing something.

johto:
> @johto mentioned that "it'd be nice to have both Valuers and Scanners committed," and I think I agree. It feels odd to not support both the sending and receiving of a type. I think we should ditch the Array function (and its Valuer return type) until we have implemented Scanners as well.

I don't think ditching it is necessary.  What I've done in my local branch yesterday is to change pq.Array() to return an interface{}, which currently only implements driver.Valuer, but would in the future implement sql.Scanner.

johto:
> What I've done in my local branch yesterday is to change pq.Array() to return an interface{}, which currently only implements driver.Valuer, but would in the future implement sql.Scanner.

Though there is a problem with that: a Scanner will require a pointer to a slice, whereas a Valuer can work with just a slice "value".  One option is to divide the structs into Valuers and Scanners; e.g. Int64Valuer, Int64Scanner.  Another option would be to do something like this:

```go
type Int64Array struct { A *[]int64 }

 // Value implements the driver.Valuer interface.
func (arr Int64Array) Value() (driver.Value, error) {
    a := *arr.A
    if a == nil {
        return nil, nil
```

I've already gone ahead and done the work for option two.  Does anybody want to object to this way of writing these?  Or anyone have any better ideas?

johto:
Also, all of these methods are currently returning byte slices rather than strings, which quite badly interferes with #357.  There appears to be a small performance penalty (~1010 ns/op vs. ~1070ns/op on my computer) for adding a conversion back to string to the `return` statements of the Valuers, but I don't see that as an issue.  Does anyone have a reason for not doing that?

array.go
cbandy: It causes more allocs but also seems faster...

`appendArrayQuotedString(v)`:
```
BenchmarkStringArrayValue         100000             16721 ns/op            2048 B/op          7 allocs/op
BenchmarkGenericArrayValueStrings         100000             17584 ns/op            1712 B/op         17 allocs/op
```

`appendArrayQuotedBytes([]byte(v))`:
```
BenchmarkStringArrayValue         100000             16130 ns/op            2624 B/op         16 allocs/op
BenchmarkGenericArrayValueStrings         100000             16993 ns/op            2352 B/op         27 allocs/op
```

Up to you!

cbandy:
> The generic array valuer seems to be having a bit of a problem with types where the underlying type is a slice.

Addressed in 166afcf3a69c68651ecff1f5b24271bee6b2fc1d.

cbandy:
> Also, all of these methods are currently returning byte slices rather than strings, which quite badly interferes with #357. There appears to be a small performance penalty (~1010 ns/op vs. ~1070ns/op on my computer) for adding a conversion back to string to the `return` statements of the Valuers, but I don't see that as an issue. Does anyone have a reason for not doing that?

It incurs another alloc and copy of the contents. I'm not opposed, however. Addressed in 2810e56c71b3175fd495ea338313d0935c1b0498.

cbandy:
>> What I've done in my local branch yesterday is to change pq.Array() to return an interface{}, which currently only implements driver.Valuer, but would in the future implement sql.Scanner.
> Though there is a problem with that: a Scanner will require a pointer to a slice, whereas a Valuer can work with just a slice "value".

Which is why I think we should wait to add any function(s) that try to lump these together. :wink:

> Does anybody want to object to this way of writing these? Or anyone have any better ideas?

I think one should be scanning into the our types directly, e.g.

```go
var a Int64Array
err := db.QueryRow(`...`).Scan(&a)
```

The implementation looks normal:

```go
func (a *Int64Array) Scan(src interface{}) error {
  for ... src ... {
    *a = append(*a, intValue)
  }
}
```

erykwalder:
I think ideally we have an interface that encompasses both Scanner/Valuer and return that from pq.Array.

As far as what's passed in, would we be able to do something like for Valuing:
```go
rows, err := db.Query(q, pq.Array(someArr))
```

and something like this for scanning:
```go
err := rows.Scan(pq.Array(&someArr))
```

But maybe I'm just not thinking about this right. I know it works pretty simply for the generic, not sure about the others.

johto:
> As far as what's passed in, would we be able to do something like for Valuing:
> 
> rows, err := db.Query(q, pq.Array(someArr))
> and something like this for scanning:
> 
> err := rows.Scan(pq.Array(&someArr))

Yeah, this is what I have in my local branch right now.  In the first case we store a pointer to someArr inside a the chosen array implementation and dereference it when Value()ing.

cbandy:
If you all are determined to prescribe the interface for Scanners, then let's just implement Scanners now...

Some questions I've encountered while thinking about Scanners:

1. Should a user be able to `var a pq.Int64Array; a.Scan("{1,2,3}")`?
2. What should happen when a user tries to scan dimensions that don't match, e.g. a PostgreSQL `int[]` into a Go `[][]int`?
3. What should we do when the user explicitly sets the lower bound, e.g. `'[5:6]={2,3}'::int[]`?

cbandy:
If you all are determined to prescribe the interface for Scanners, then let's just implement Scanners now...

Some questions I've encountered while thinking about Scanners:

1. Should a user be able to `var a pq.Int64Array; a.Scan("{1,2,3}")`?
2. What should happen when a user tries to scan dimensions that don't match, e.g. a PostgreSQL `int[]` into a Go `[][]int`?
3. What should we do when the user explicitly sets the lower bound, e.g. `'[5:6]={2,3}'::int[]`?
4. What destination types should we support? The Valuer implementation delegated this responsibility to `driver.DefaultParameterConverter`, but I don't see any exported access to the Scan equivalent, [`convertAssign`](https://github.com/golang/go/blob/release-branch.go1.4/src/database/sql/convert.go#L85).

johto:
> Some questions I've encountered while thinking about Scanners:
> 
> Should a user be able to var a pq.Int64Array; a.Scan("{1,2,3}")?

I think that could work.  Given:

```go
type Int64Array struct { A *[]int64 }
```

Scanning into that directly would realize the A pointer is nil and would create a new slice.  If you went through the pq.Array() interface and gave a pointer to your own slice, that would be first cleared and then appended to instead.  AFAICT that would work nicely for all three use cases (two different types of Scanning, and one Valuer).

> What should happen when a user tries to scan dimensions that don't match, e.g. a PostgreSQL int[] into a Go [][]int?

An error.

> What should we do when the user explicitly sets the lower bound, e.g. '[5:6]={2,3}'::int[]?

I don't really care.  I see exactly zero use for arrays like this.  Perhaps just an error from the depths of hell would work.

> What destination types should we support? The Valuer implementation delegated this responsibility to driver.DefaultParameterConverter, but I don't see any exported access to the Scan equivalent, convertAssign.

Hmm.  Perhaps we could just support the "native" ones, and then the "generic" implementation would require Scanner to be implemented for the types of the elements?

erykwalder:
There's also the possibility of something like json.Unmarshal, which just tries to figure out scanning for generic types, if possible, returning an error otherwise.

cbandy:
> ```go
type Int64Array struct { A *[]int64 }
```
> Or anyone have any better ideas?

The following appears to work:

```go
func (a *BoolArray) Scan(src interface{}) error { ... }

func TestBoolArrayScan(t *testing.T) {
        var b []bool
        var c = (*BoolArray)(&b)

        err = c.Scan([]byte(`{t,f,t}`))
        if err != nil {
                t.Fatalf("Expected no error, got %v", err)
        }
        if !reflect.DeepEqual(b, []bool{true, false, true}) {
                t.Errorf("got %q", a)
        }
}
```

I think this means we can implement the Scanner interface on `*{Bool,Bytea,Float64,Int64,String}Array` normally and have `Array()` perform a conversion.

I would very much like to keep the types/definitions the same as they are in this PR currently.

erykwalder:
Would it be helpful for me to spike out a generic scanner?

cbandy:
> Would it be helpful for me to spike out a generic scanner?

It would definitely help me see how it can work.

cbandy:
I've written [some tests for the basic scanners](https://github.com/cbandy/go-pq/compare/array-valuer...array-scanner), but I haven't come upon implementations that I like.

johto:
> I've written some tests for the basic scanners, but I haven't come upon implementations that I like.

The work here seems to have been sitting idle for a while.  I don't see the problem with doing this the way I suggested:

`type Int64Array struct { A *[]int64 }`

AFAICT this will provide a nicer interface for both scanners and valuers.

cbandy:
> > I've written some tests for the basic scanners, but I haven't come upon implementations that I like.

> The work here seems to have been sitting idle for a while.

@johto I've finished implementing Scanners on the basic/one-dimensional arrays in another branch. I was going to post here once I finished the generic implementation, but I've stalled out re-implementing `convertAssign`.

> > What destination types should we support? The Valuer implementation delegated this responsibility to `driver.DefaultParameterConverter`, but I don't see any exported access to the Scan equivalent, [`convertAssign`](https://github.com/golang/go/blob/release-branch.go1.4/src/database/sql/convert.go#L85).

> Hmm. Perhaps we could just support the "native" ones, and then the "generic" implementation would require Scanner to be implemented for the types of the elements?

I will post patches for this approach ASAP. Shall I go ahead and add Scanners to this PR?

johto:
> Shall I go ahead and add Scanners to this PR?

Sounds good.

> I will post patches for this approach ASAP.

.. though don't hurry too much.  I'm going try to and spend the new couple of weeks on Postgres stuff, which means I'll likely not have too much time to spend on reviewing this patch.

xentek:
Thank you for all of the hard work on implementing this type in a way that has a nice interface and conforms to go idioms and interfaces. Any chance this will see the light of day soon?

piotrkowalczuk:
For anybody that don't want to wait:
https://godoc.org/github.com/piotrkowalczuk/pqt#ArrayInt64
https://godoc.org/github.com/piotrkowalczuk/pqt#ArrayString

ibash:
Is this still being worked on?

cbandy:
> Is this still being worked on?

Things got busy before and after a new baby. As I recall, the pared-down goal of a slice of `Scanners` still had a problem because the method is defined on a pointer:

```go
var x Thing
rows.Scan(&x) // success
rows.Scan(x) // failure
```

I think a user would expect to do the following:

```go
var y []Thing
rows.Scan(&y)
```

However, `Thing` doesn't implement `Scanner` (`*Thing` does.) This isn't a slice of `Scanner`. Creating and checking the pointer of an element at each level started to feel like magic. Maybe it wasn't so bad...

At this point, I'd like to get _something_ published. I think the quickest way forward is to support sending and receiving only one-dimensional arrays, and in the generic case, only Valuer and Scanner types.

Are these two limitations acceptable to others?

---

- The "native" types are done.
- The code to send multidimensional arrays is done.
- The code to parse the multidimensional array format is done.
- ToDo: Scan the parsed elements into a slice of (almost) Scanner.

ibash:
@cbandy congrats!
I have no qualms with publishing with the two limitations mentioned.

douglarek:
Is this still being worked on?

cbandy:
I rebased this work, included scanners, and opened a fresh PR. Please take a look. I think I addressed all the comments here, but if not, I'd like to continue the discussion there.

freeformz:
I think this can be closed since #466 was merged. Please re-open if I am wrong.

PR #366, opened 2015-05-12
Prepared Statement Caching

bfrog:
Currently Go's database/sql package does not allow you to keep prepared queries around per connection, only per transaction. This fixes that by keeping around unclosed prepared queries by naming them using a hash of their query string and holding on to a map[string]*stmt per connection.

This greatly speeds up our application which uses many complex queries containing 20+ joins. The plan was taking nearly 200ms, while the query itself only ~20ms. Our application performance was greatly improved by this.

I'm PRing this in hope it might be useful to others and that we might get some feedback.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -239,6 +239,7 @@ func DialOpen(d Dialer, name string) (_ driver.Conn, err error) {
 	cn.ssl(o)
 	cn.buf = bufio.NewReader(cn.c)
 	cn.startup(o)
+	cn.prepared = make(map[string]*stmt)
 	// reset the deadline, in case one was set (see dial)
 	if timeout := o.Get("connect_timeout"); timeout != "" && timeout != "0" {
 		err = cn.c.SetDeadline(time.Time{})
@@ -717,7 +718,7 @@ func (cn *conn) Prepare(q string) (_ driver.Stmt, err error) {
 		return prepared0, nil
 	}
 	prepared, err := cn.prepareTo(q, name)
-	if err != nil {
+	if err == nil {
 		cn.prepared[name] = prepared
 	}
 	return prepared, err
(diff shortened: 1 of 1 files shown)


johto:
I'm too tired to think about this right now, so I'm just going to do a quick brain dump here:

I still think https://go-review.googlesource.com/#/c/2035/ is the better solution.

That said, this is a bit different, since the caching happens regardless of how the prepared statement is created.  Unsure whether that's a desirable feature or not.  If it is, it's not completely clear whether this should happen in the driver or (when the CL above is finally merged) in a driver-agnostic layer on top of database/sql.

conn.go
```diff
@@ -23,6 +24,7 @@ import (
 	"unicode"
 
 	"github.com/lib/pq/oid"
+	"github.com/spaolacci/murmur3"
```
uhoh-itsmaciek: I don't think we want to introduce an external dependency here, since we have none right now. Is there a way to avoid that? But johto's comments are probably more germane.

pkieltyka: agreed, please no more external dependencies

pkieltyka: what is another way to generate a fast hash of the query with stdlib?

johto: Why do we need to hash at all?  Can't we just put the query string into the map and let its implementation take care of it?

bfrog: Thats a good point, when I did this I was thinking I'd need to pull the name out of pg_prepared_statements on the server at some point, that clearly doesn't need to happen. Let me remove that.

johto:
Hmm.  As written, I don't think this would work for the *Tx.Stmt() case in Go 1.4 or later, since after commit 5f739d9dcd01730e10d829968755c7b1a7b5f2b7 prepared statements created in a transaction are always cleaned up by database/sql.  I think if this was to be merged, I'd rather have the entire functionality be optional, and no statements would ever be cleaned up on the server side.  That way both *Tx.Prepare() and *Tx.Stmt() would benefit.

But I'm not sure this is worth it, to be honest, especially given how many options we've added lately.  If someone really has an issue like this, they can probably roll a patched version of pq until the shortcoming in database/sql is fixed, and if someone wants a generic "reuse any statements based on the query text" functionality, they should probably write that on *top* of database/sql, not way below it in the driver layer.

PR #370, opened 2015-06-01
avoid index out of range panic if string is shorter than we expect it

jnewmano:


johto:
Can we also add a test case for this?

jnewmano:
Ultimately errorf(...) is going to panic anyways, so this PR doesn't really matter. The panic escaping out of pq is causing issues. Is there a reason pq should panic in this case instead of returning an error back to the code that called it? Perhaps there should be a panic recover in rows.Next(), instead of rewriting to pass the error all the way back up.

 https://github.com/lib/pq/blob/master/conn.go#L1406

I can submit another PR with the recover in it, if you agree.

johto:
Does [this line](https://github.com/lib/pq/blob/a8d8d01c4f91602f876bf5aa210274e8203a6b45/conn.go#L1414) not cover it?

jnewmano:
It still passes the panic on runtime errors, which in our case was hit by an index out of bounds caused by expect.

https://github.com/lib/pq/blob/master/error.go#L482

PR #371, opened 2015-06-01
handle postgres date type when dates are BC

jnewmano:
Handle the case when a postgres column of date type has date values that are BC.  For example, 0011-02-03 BC.

### DIFF
--- a/encode_test.go
+++ b/encode_test.go
@@ -71,6 +71,24 @@ var timeTests = []struct {
 	{"123456-02-03 04:05:06.1", time.Date(123456, time.February, 3, 4, 5, 6, 100000000, time.FixedZone("", 0))},
 }
 
+var dateTests = []struct {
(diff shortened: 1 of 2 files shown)


encode_test.go
```diff
@@ -100,10 +118,23 @@ func TestParseTs(t *testing.T) {
...
+				i, tt.str, tt.timeval, val)
+		}
+	}
+}
+
 // Now test that sending the value into the database and parsing it back
 // returns the same time.Time value.
 func TestEncodeAndParseTs(t *testing.T) {
-	db, err := openTestConnConninfo("timezone='Etc/UTC'")
+	db, err := openTestConnConninfo("")
```
johto: Why this change?

jnewmano:
I have updated the pull request to get rid of the unnecessary code change that you pointed out.

PR #373, opened 2015-06-18
Receive UUID in binary format when possible

cbandy:
I wasn't able to write a test that demonstrated a difference in performance for either UUID format. @johto perhaps you have a way to ascertain the impact?


tamird:
This LGTM. Any reason not to get it in?

mjibson:
This needs a rebase.

cbandy:
Rebased and added an assertion on the expected input length.

PR #383, opened 2015-08-09
Small documentation spelling mistake

poopoothegorilla:


johto:
`iff` is a somewhat common shortened version of `if and only if`, so it's intentional in this context.

PR #388, opened 2015-09-03
Fix COPY FROM statements when copying all columns

tegioz:


freeformz:
My only though on this is that it would actually be good to test the generated statements against an actual table to ensure that they work across all versions of psql.

Otherwise +1

Issue #121, opened 2013-08-02
inet, cidr, macaddr support


georgyo:
Just wondering if other postgres types are planned to be supported? Specifically inet, cidr and macaddr.

Similar to time.Time, if stored in the database that way, its also most likely better to have a better in code representation of it.

Something along the lines of: https://gist.github.com/georgyo/6143350

The biggest problem I see is that these types currently get caught by the return s statement, meaning that this has the possibility to break existing code... Still curious because of time.Time.

gucki:
:+1:

petergoldstein:
@johto Is this an issue that is of interest to the library maintainers?  I'm asking because it's clearly a very old issue (2+ years), and doesn't seem to have received any attention during that time. 

I have a need for this functionality, and would be open to scheduling some time to contribute to the project.  But before I devote the time, I'd like to know if there's a realistic chance to get it merged in.

On a related note, is the backwards compatibility concern raised by @georgyo in his initial comment a real concern for you?  As is it seems like the pq library just doesn't handle queries against those columns properly at all, so any code that relied on existing behavior would be problematic at best.

@johto If you indicate support for a possible PR (subject to review of course) I'm happy to contribute one with necessary tests, etc.

johto:
> @johto Is this an issue that is of interest to the library maintainers? I'm asking because it's clearly a very old issue (2+ years), and doesn't seem to have received any attention during that time.

I've not needed to interpret inet or cidr types from any Go application, so I've not spent any time on this myself.  But the fact that it hasn't received any attention from me doesn't really matter; I've spent a lot of time merging in features I've never used and probably never will.  The fact that this issue has been open for two years suggests that most people don't care about this feature, or at least not enough to contribute it.

> On a related note, is the backwards compatibility concern raised by @georgyo in his initial comment a real concern for you? As is it seems like the pq library just doesn't handle queries against those columns properly at all, so any code that relied on existing behavior would be problematic at best.

There's nothing problematic with the current behavior.  You can scan any "custom" type into a string and then post-process to do whatever you need with it.  Breaking that is not only out of the question, but in violation of the database/sql/driver interface:

```
// Next is called to populate the next row of data into
// the provided slice. The provided slice will be the same
// size as the Columns() are wide.
//
// The dest slice may be populated only with
// a driver Value type, but excluding string.
// All string values must be converted to []byte.
```

> @johto If you indicate support for a possible PR (subject to review of course) I'm happy to contribute one with necessary tests, etc.

[...]

PR #390, opened 2015-09-08
Add Scanner/Value implementations for Postgres network address types.

petergoldstein:
An implementation of Scanner / Value types for Postgres network address types, intended to make it easier to work with these types.  Includes the three current network address types - cidr, inet, macaddr.  Addresses issue #121 

Some quick comments:
1. The package name 'netaddr' was chosen to conform with Go recommended best practices.  Happy to change this package name if there's a strong preference on the maintainer's part for another package name.
2. I decided to give all of the types a consistent behavior with regards to NULL handling.  Like the native value types (Bool, Int64, etc.) the corresponding structs have a 'Valid' member.  When false, this indicates that the corresponding database value is NULL.  Technically the native types for inet and macaddr support a nil value, and so for these cases a simpler struct could have been used.  But I opted for a more consistent interface.
3. It was unclear to me from the existing Hstore implementation whether an error in a Scan should result in a panic or the explict return of an error.  I chose to explicitly return an error rather than panic if Scan received a value that cannot be coerced to a byte array.

Happy to discuss or revisit any of the above.

### DIFF
--- a/netaddr/inet.go
+++ b/netaddr/inet.go
@@ -1,0 +1,44 @@
+package netaddr
+
+import (
+	"database/sql/driver"
+	"errors"
+	"net"
+)
+
+// A wrapper for transferring Inet values back and forth easily.
+type Inet struct {
+	Inet  net.IP
+	Valid bool
+}
+
+// Scan implements the Scanner interface.
+func (i *Inet) Scan(value interface{}) error {
+	i.Inet = nil
+	i.Valid = false
+	if value == nil {
+		i.Valid = false
+		return nil
+	}
+	ipAsBytes, ok := value.([]byte)
+	if !ok {
+		return errors.New("Could not convert scanned value to bytes")
+	}
+	parsedIP := net.ParseIP(string(ipAsBytes))
+	if parsedIP == nil {
+		i.Valid = false
+		return nil
+	}
+	i.Valid = true
+	i.Inet = parsedIP
+	return nil
+}
+
+// Value implements the driver Valuer interface. Note if i.Valid is false
+// or i.IP is nil the database column value will be set to NULL.
+func (i Inet) Value() (driver.Value, error) {
+	if i.Valid == false || i.Inet == nil {
+		return nil, nil
+	}
+	return []byte(i.Inet.String()), nil
+}
--- a/netaddr/inet.go
+++ b/netaddr/inet.go
@@ -8,28 +8,20 @@ import (
 
 // A wrapper for transferring Inet values back and forth easily.
 type Inet struct {
-	Inet  net.IP
-	Valid bool
+	Inet net.IP
 }
 
 // Scan implements the Scanner interface.
 func (i *Inet) Scan(value interface{}) error {
 	i.Inet = nil
-	i.Valid = false
-	if value == nil {
-		i.Valid = false
-		return nil
-	}
 	ipAsBytes, ok := value.([]byte)
 	if !ok {
 		return errors.New("Could not convert scanned value to bytes")
 	}
 	parsedIP := net.ParseIP(string(ipAsBytes))
 	if parsedIP == nil {
-		i.Valid = false
 		return nil
 	}
-	i.Valid = true
 	i.Inet = parsedIP
 	return nil
 }
@@ -37,7 +29,7 @@ func (i *Inet) Scan(value interface{}) error {
 // Value implements the driver Valuer interface. Note if i.Valid is false
 // or i.IP is nil the database column value will be set to NULL.
 func (i Inet) Value() (driver.Value, error) {
-	if i.Valid == false || i.Inet == nil {
+	if i.Inet == nil {
 		return nil, nil
 	}
 	return []byte(i.Inet.String()), nil
--- a/netaddr/inet.go
+++ b/netaddr/inet.go
@@ -14,6 +14,9 @@ type Inet struct {
 // Scan implements the Scanner interface.
 func (i *Inet) Scan(value interface{}) error {
 	i.Inet = nil
+	if value == nil {
+		return nil
+	}
 	ipAsBytes, ok := value.([]byte)
 	if !ok {
 		return errors.New("Could not convert scanned value to bytes")
@@ -26,8 +29,8 @@ func (i *Inet) Scan(value interface{}) error {
 	return nil
 }
 
-// Value implements the driver Valuer interface. Note if i.Valid is false
-// or i.IP is nil the database column value will be set to NULL.
+// Value implements the driver Valuer interface. Note if
+// i.IP is nil the database column value will be set to NULL.
 func (i Inet) Value() (driver.Value, error) {
 	if i.Inet == nil {
 		return nil, nil
--- a/netaddr/inet.go
+++ b/netaddr/inet.go
@@ -6,7 +6,7 @@ import (
 	"net"
 )
 
-// A wrapper for transferring Inet values back and forth easily.
+// Inet is a wrapper for transferring Inet values back and forth easily.
 type Inet struct {
 	Inet net.IP
 }
--- a/netaddr/inet_test.go
+++ b/netaddr/inet_test.go
@@ -1,0 +1,68 @@
+package netaddr
+
+import (
+	"bytes"
+	"net"
+	"testing"
+
+	_ "github.com/lib/pq"
+)
+
+func TestInet(t *testing.T) {
+	db := openTestConn(t)
+	defer db.Close()
+
+	inet := Inet{}
+
+	// Test scanning NULL values
+	err := db.QueryRow("SELECT NULL::inet").Scan(&inet)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if inet.Valid {
+		t.Fatalf("expected null result")
+	}
+
+	// Test setting NULL values
+	err = db.QueryRow("SELECT $1::inet", inet).Scan(&inet)
+	if err != nil {
+		t.Fatalf("re-query null value failed: %s", err.Error())
+	}
+	if inet.Valid {
+		t.Fatalf("expected null result")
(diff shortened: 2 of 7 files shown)


xentek:
:+1:

initzero:
Agree with @jvehent; this would be a great additional to the library, for my use case.

bakape:
Any reason for this not being merged? Currently forced to store IPs as TEXT.

johto:
> Any reason for this not being merged? Currently forced to store IPs as TEXT.

Who's forcing you?  This PR is about convenience, not about adding functionality that couldn't be achieved otherwise.

bakape:
"Force" is probably the wrong word here. Rather it's a feature of Postgres that is missing from the driver. You can either not use the feature, implement it yourself or fork this repository, merge this PR and maintain a separate version for your own purposes - not a good long term solution. 

If there are any particular reasons why this PR is not being merged, perhaps I can help resolve them?

dimitertodorov:
I think this should be merged. The inet and cidr types are part of core postgres. 
I would be happy to address any concerns with the PR as well.

makslevental:
+1 for merging this in

netaddr/inet_test.go
```diff
@@ -0,0 +1,68 @@
...
+	testBidirectional := func(i Inet, label string) {
+		err = db.QueryRow("SELECT $1::inet", i).Scan(&inet)
+		if err != nil {
+			t.Fatalf("re-query %s inet failed: %s", label, err.Error())
+		}
+		if !inet.Valid {
+			t.Fatalf("expected non-null value, got null for %s", label)
+		}
+		if bytes.Compare(i.Inet, inet.Inet) != 0 {
+			t.Fatalf("expected IP addresses to match, but did not for %s - %s %s", label, inet.Inet.String(), inet.Inet.String())
```
makslevental:
```go
t.Fatalf("expected IP addresses to match, but did not for %s - %s %s", label, inet.Inet.String(), inet.Inet.String())
``` 
should be

```go
t.Fatalf("expected IP addresses to match, but did not for %s - %s %s", label, i.Inet.String(), inet.Inet.String())
```

Lekensteyn:
Any reason to keep the `Valid` field? `net.IP` happens to be a `[]byte`, so it can be nulled.

Removing it seems to work, see this test:
```
scanned: ::1
valued: [58 58 49]
scanned: 127.0.0.1
valued: [49 50 55 46 48 46 48 46 49]
scanned: ::
valued: [58 58]
scanned: <nil>
valued: <nil>
```

```go
package main

import (
	"database/sql"
	"database/sql/driver"
	"errors"
	"fmt"
	"net"

	_ "github.com/lib/pq"
)

func main() {
	db, err := sql.Open("postgres", "sslmode=disable")
	if err != nil {
		panic(err)
	}
	if err = db.Ping(); err != nil {
		panic(err)
	}
	queries := []string{
		`SELECT '::1'::inet`,
		`SELECT '127.0.0.1'::inet`,
		`SELECT '00::'::inet`,
		`SELECT NULL::inet`,
	}
	for _, q := range queries {
		var ip Inet
		err = db.QueryRow(q).Scan(&ip)
		if err != nil {
			panic(err)
		}
		fmt.Printf("scanned: %v\n", ip)

		var val interface{}
		err = db.QueryRow(`SELECT $1::inet`, ip).Scan(&val)
		if err != nil {
			panic(err)
		}
		fmt.Printf("valued: %v\n", val)
	}
}

// A wrapper for transferring Inet values back and forth easily.
type Inet struct {
	net.IP
}

// Scan implements the Scanner interface.
func (i *Inet) Scan(value interface{}) error {
	i.IP = nil
	if value == nil {
		return nil
	}
	ipAsBytes, ok := value.([]byte)
	if !ok {
		return errors.New("Could not convert scanned value to bytes")
	}
	parsedIP := net.ParseIP(string(ipAsBytes))
	if parsedIP == nil {
		return nil
	}
	i.IP = parsedIP
	return nil
}

// Value implements the driver Valuer interface.
func (i Inet) Value() (driver.Value, error) {
	if i.IP == nil {
		return nil, nil
	}
	return []byte(i.IP.String()), nil
}
```

Matt-Texier:
Hi Guys,

I would like to know where we are with this Pull Request ... It would be really important for a project I am working on at the time being to support it.

Could you please update the status of this PR ?

Thanks Matthieu.

uhoh-itsmaciek:
I think https://github.com/lib/pq/pull/390/files#r109256763 needs to be addressed and @Lekensteyn's comment above seems valid, but otherwise this can probably be merged? Support for built-in Postgres data types (at least where there is a reasonable Go data type to map to) certainly seems within the purview of the driver.

kr:
Removing `Valid` for `Inet` seems fine to me, and then it would admit being defined as `type Inet net.IP`, which might be nice.

For anyone else who'd like to use this, note that you can always copy or reimplement these types in your own code — it's just a convenience to have it bundled in lib/pq.

petergoldstein:
I rebased against master, as this PR was substantially out of date.
I also addressed the typo in the fatal message pointed out by @makslevental  .  Thanks!

I haven't looked at this code in almost 3 years, so I don't have strong opinions about the `Valid` on `Inet`.  I'd be happy to update based on consensus.

If a committer wants to shepherd this in, I'm happy to make minor changes to get that done in the near term.

uhoh-itsmaciek:
Let me rephrase that: is there any advantage to keeping `Valid`? In general I think less is better...

kr:
> is there any advantage to keeping `Valid`?

None that I can see. (I mean, there's a small advantage in that it's more prominent than the distinction between nil and non-nil-zero-length. But I think the number of situations where that distinction is relevant isn't enough to optimize for that usage.)

uhoh-itsmaciek:
Empty strings are not valid for any of these types in Postgres though, right? If you can update to avoid `Valid` @petergoldstein and there are no objections, I can merge. I've restarted Travis and tests pass (seemed like a Travis issue last time).

netaddr/inet.go
```diff
@@ -0,0 +1,36 @@
...
+	parsedIP := net.ParseIP(string(ipAsBytes))
+	if parsedIP == nil {
+		return nil
+	}
+	i.Inet = parsedIP
+	return nil
+}
+
+// Value implements the driver Valuer interface. Note if i.Valid is false
+// or i.IP is nil the database column value will be set to NULL.
```
uhoh-itsmaciek: This comment is stale, no?

uhoh-itsmaciek:
@tsuki can you comment here re: https://github.com/lib/pq/pull/780#issuecomment-416841101 and provide more detail as to why that approach is better? That seems to be what @kr was suggesting in https://github.com/lib/pq/pull/390#issuecomment-412249089 too.

bakape:
A falsification of meritocracy.

uhoh-itsmaciek:
So reviewing this again before merging, I was struck by the inconsistency across the types introduced:

```
type Cidr struct {
  Cidr net.IPNet
  Valid bool
}

type Inet struct {
  Inet net.IP
}

type Macaddr struct {
  Macaddr net.HardwareAddr
  Valid bool
}
```

Two types have a valid flag and one checks the nil-ness of the embedded type. I also thought again about #780--the approach there is to just define a type as `type X net.T` directly, which seems simpler. I think the only reason to really have a wrapper struct with a `Valid` flag is in case the zero value of the type is actually meaningful on its own in representing the underlying SQL type (e.g., since the empty string is distinct from a null). Can we change these all to

```
type Cidr net.IPNet
type Inet net.IP
type Macaddr net.HardwareAddr
```

and use nil checks instead of `Valid` checks? Thoughts?

Qhesz:
While this PR is unlikely to ever get merged (it's about 5 years old)... I'll leave this here for anyone who might use this as a base for their own Scan/Value implementations:

Instead of 
`type Inet net.IP`
It should be 
`type Inet net.IPNet`

The Postgres inet type can be a HostAddress/Prefix pair, while the Golang net.IP type only contains a HostAddress. The Golang net.IPNet type is correct because it can hold any HostAddress/Prefix pair, including /32 to represent a single address.

The inet type also omits the /Prefix part on output for /32's for ipv4 (and /128's for ipv6), so you need to:
```
ip := net.ParseIP()
if ip != nil {
  // Output had no prefix, manually add the /32 for ipv4s, /128 for ipv6
  ...
} else {
  // Parse it as 'host/prefix'
  ... := net.ParseCIDR(...)
}
```

kinghrothgar:
> So reviewing this again before merging, I was struck by the inconsistency across the types introduced:
> 
> ```
> type Cidr struct {
>   Cidr net.IPNet
>   Valid bool
> }
> 
> type Inet struct {
>   Inet net.IP
> }
> 
> type Macaddr struct {
>   Macaddr net.HardwareAddr
>   Valid bool
> }
> ```
> Two types have a valid flag and one checks the nil-ness of the embedded type. I also thought again about #780--the approach there is to just define a type as `type X net.T` directly, which seems simpler. I think the only reason to really have a wrapper struct with a `Valid` flag is in case the zero value of the type is actually meaningful on its own in representing the underlying SQL type (e.g., since the empty string is distinct from a null). Can we change these all to
> 
> ```
> type Cidr net.IPNet
> type Inet net.IP
> type Macaddr net.HardwareAddr
> ```
> and use nil checks instead of `Valid` checks? Thoughts?

Is this modification the only thing keeping this from being merged? I can do this if that will finally get this in.

lanrat:
Any updates on this issue? I too would like to get this merged.

bakape:
The update seems to be "use pgx". Having more stable maintenance-mode
libraries around is not a bad thing, mind you.

cmorent:
Any hope this PR get merged someday?

bakape:
None, it seems. Just use pgx, if you want more features.

PR #396, opened 2015-10-04
cache SSL client certs

kr:
benchmark                                       old ns/op     new ns/op     delta
    BenchmarkSetupSSLClientCertificatesParse-8      379742        77.8          -99.98%
    BenchmarkSetupSSLClientCertificatesNoFile-8     2592          59.7          -97.70%
    
    benchmark                                       old allocs     new allocs     delta
    BenchmarkSetupSSLClientCertificatesParse-8      2702           0              -100.00%
    BenchmarkSetupSSLClientCertificatesNoFile-8     15             0              -100.00%
    
    benchmark                                       old bytes     new bytes     delta
    BenchmarkSetupSSLClientCertificatesParse-8      206032        0             -100.00%
    BenchmarkSetupSSLClientCertificatesNoFile-8     680           0             -100.00%

### DIFF
--- a/bench_test.go
+++ b/bench_test.go
@@ -5,11 +5,14 @@ package pq
 import (
 	"bufio"
 	"bytes"
+	"crypto/tls"
 	"database/sql"
 	"database/sql/driver"
 	"io"
+	"io/ioutil"
 	"math/rand"
 	"net"
+	"os"
 	"runtime"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


kr:
Not sure what's going on with the builds from Go tip. This patch doesn't touch anything related to the tests that are failing.

kr:
The travis failures appear to be unrelated to this change. See #398.

johto:
Gah.  I'd already forgotten that there's no way to associate anything with an *sql.DB instance with the way database/sql has been designed.

I'm really not sure this is a good idea, at least as currently written.  Any application which previously allowed you to reload certificates by letting go of its *sql.DB handle and creating a new one will suddenly not be able to do that at all.  That seems like a big issue.

johto:
> Gah. I'd already forgotten that there's no way to associate anything with an *sql.DB instance with the way database/sql has been designed.

I wonder if a better way forward would be to suggest a fix for this problem first.  The drivers could implement an optional interface which would be called on sql.Open(), and instead of storing the shared "driver" object, DB.driver would be set to a new object returned by this new suggested method.  A driver factory of a kind, if you will.  The interface of Open already supports that since you can it can return an error.

kr:
The cert and key are global data. I don't see what they have to do with a *sql.DB. The same cert can and should be used for more than one *sql.DB instance.

> reload certificates by letting go of its *sql.DB handle

This seems like a peculiar way to load new global certificate data. This package certainly never worked that way before. As a user, I'd be surprised to see such behavior. Actually, I'd probably never need to do it in the first place, but if I did, I'd expect to call a function like pq.ReloadCerts, or, preferably, to have it match the current behavior: to load the current cert data for each new connection.

For my purposes, I don't really need this patch any more, so I'll just abandon it here. But anyone else feel free to pick it up and take it across the line.

PR #397, opened 2015-10-05
get home directory from the environment

### LINKED ISSUE #395: home directory lookup can fail
Method setupSSLClientCertificates uses uses getpwuid_r (via user.Current) to find the user's home directory.

```go
        // Automatically load certificates from ~/.postgresql.
        user, err := user.Current()
        if err != nil {
            // user.Current() might fail when cross-compiling.  We have to
            // ignore the error and continue without client certificates, since
            // we wouldn't know where to load them from.
            return
        }
```

It would be better to use the environment variable `HOME` (or, on windows, a combination of `HOMEDRIVE` and `HOMEPATH`). That way the user can control where it looks.

Avoiding os/user would also prevent having to call into cgo, which would make cross compiling a bit easier and behave more consistently. For example, the comment above about how the lookup "might fail" would no longer be necessary.

If this sounds ok, I'll send a patch.

kr:
Fixes #395.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -14,7 +14,6 @@ import (
 	"io/ioutil"
 	"net"
 	"os"
-	"os/user"
 	"path"
 	"path/filepath"
 	"strconv"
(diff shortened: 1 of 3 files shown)


kr:
The travis failures appear to be unrelated to this change. See #398.

johto:
At least in my environment this might actually break with e.g. sudo which worked just fine before:

```
% sudo -u qwr sh -c 'echo "$HOME"'
/Users/marko
% sudo -u qwr -H sh -c 'echo "$HOME"'
/Users/qwr
```

And if user.Current() actually panics instead of returning an error in your environment, we can't fall back to `$HOME` either.  (We perhaps could for cross-compiling but that might be surprising and doesn't solve the original problem.)

I really don't know what to do here.  Any other committers want to express their thoughts on the matter?

freeformz:
@johto it could work differently than currently at times (since it depends on the environment), but this change could be more understandable to users. 

@kr can you reproduce the panic that causes this? It would be interesting to look at the value of `C.sysconf(C._SC_GETPW_R_SIZE_MAX)` if so.

freeformz:
I also took a look at the C libpq and it's [effectively doing](https://github.com/postgres/postgres/blob/master/src/port/thread.c#L95) what user.CurrentUser() is doing and we may not want to vary too far from that here.

tamird:
@kr master is now fixed, could you rebase this?

Also, this approach seems overly heavy-handed. If `user.Current()` succeeds, it should be used - the environment should be consulted only as a fallback.

kr:
While I disagree with what the C libpq does, it's behavior is not likely to change. It's probably best for us to match whatever it does, so I'll just close this.

PR #398, opened 2015-10-05
test the travis build

kr:
NOT FOR SUBMISSION

This is an empty commit just to see if the previous commit still passes on Go tip in Travis. See #396 and #397.


kr:
Ok, so commit b269bd035a727d6c1081f76e7a239a1b00674c40, the latest on master, also currently fails on Travis, as can be seen in this PR.

I'd suggest we consider removing the "tip" version from the travis tests. (And add Go 1.5 while we're at it.)

cbandy:
We could allow failures for tip, which will not show red here if/when it fails.

http://docs.travis-ci.com/user/customizing-the-build/#Rows-that-are-Allowed-to-Fail

kr:
> We could allow failures for tip

Nice, that would allow failing tip results to still be available, without blocking otherwise-successful pull requests.

+1 from me.

kr:
I just opened #399 to suggest the "allowed failures" fix. Closing this one.

PR #399, opened 2015-10-05
allow Go tip to fail

kr:
Go 1.5 was released in August of 2015.

The "tip" version of Go isn't guaranteed to be backwards
compatible or even stable, so those builds might fail
for reasons unrelated to any lib/pq pull request. It's
useful to see the results, but they shouldn't block a
pull request from being landed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,10 @@ go:
   - 1.5
   - tip
 
+matrix:
+  allow_failures:
+    - go: tip
+
 before_install:
   - psql --version
   - sudo /etc/init.d/postgresql stop


kr:
PTAL

uhoh-itsmaciek:
This is probably wise for PR management sanity, but it does have the unfortunate side effect of sweeping tip failures under the rug. This is both bad for us (if Go changes in ways that are `database/sql` spec-compatible but violate some assumptions we are making) and bad for Go (if there are changes that are accidentally spec-incompatible or bugs). In either case, we may not find out until there's a release and we add it to our test matrix.

Do you know if there's a way to surface tip failures more easily without treating them as failures? I think the trade-off is worth it, and I can merge this if no one objects, but I do think there are drawbacks here.

kr:
I am not much of a travis expert, so I don't know if there are alternatives.

This definitely has drawbacks, but all of the bad scenarios you just described are already possible if there's no lib/pq pull request activity before a go release. For example, that pretty travis badge on the github home page for this repo would say "failing" right now if travis hypothetically somehow reran the tests for master.

I think a more ideal way to address these problems would be some way to eagerly and automatically test betas and release candidates. Eagerly as in, it doesn't depend on some unrelated activity such as opening a pull request, it should happen when a pre-release version appears, because of that event. That way the noise from tip shenanigans during development won't bother pull requests here, but we'd have a good chance of catching any our-fault or their-fault problems with the database/sql interface during the bug-fixing phase leading up to a go release. I don't know if such a thing is possible without nontrivial engineering effort. I would guess travis doesn't have anything like this built in, but who knows.

johto:
> I think a more ideal way to address these problems would be some way to eagerly and automatically test betas and release candidates.

That would be a step forwards.  Even nicer would be testing every Go commit which changes anything under src/database/sql/, but I'm not sure either is possible.

> It's
> useful to see the results, but they shouldn't block a
> pull request from being landed.

I don't think they ever have.

For now, I'm holding the opinion that seeing red shouldn't make a committer completely ignore a pull request.  With status quo, we should be faster to react to changes in database/sql which break our test cases / our code, so I'd think this is better than ignoring errors in tip.

johto:
Oh, and I also committed the 1.5 part of this in ffe986aba3e6cfcded8b06615965941408891ef0, since there was an outstanding pull request for that part alone already.

uhoh-itsmaciek:
For what it's worth, I'm convinced by @kr's argument here (but of course I'm not going to force a merge through by myself, especially since I'm minimally active now).

johto:
> For what it's worth, I'm convinced by @kr's argument here (but of course I'm not going to force a merge through by myself, especially since I'm minimally active now).

Which part exactly?  It seems to me that the PR makes the tests run less often (or the failures invisible, which AIUI in practice is the same thing), yet he would actually like to run the tests *more* often.  The reasoning seems to be a straw man representing belief that committers won't look at PRs if they don't pass tests, which to me suggests a fundamental lack of faith in the committers.

The only actual issue I see with the current setup which this *would* improve is that a (possibly less experienced) developer wouldn't have to go around chasing tip failures for their PR.  But I don't think the tip is broken often enough for that to actually truly matter.

kr:
Yes, I was assuming a pretty strict policy of not merging pull requests that have failing travis status. If maintainers are ok with merging a pull request even when tip is broken (and living with a read "failing" badge on github) then there's much less motivation for this change.

kr:
I'll rebase this to get rid of the 1.5 part.

kr:
Rebased (and edited commit message).

tamird:
This can possibly be closed if #428 is merged.

tamird:
@johto probably should close this for now.

freeformz:
:+1: on closing this. I'd like to know when lib/pq is broken against tip.

PR #401, opened 2015-10-08
fixes the decoding of strings and int16/int32

jvshahid:
Without this fix scanning a `smallint` or `integer` into an `interface{}`
will cause the value to be of type `int64` igorning the type information
of the column. Same thing happens for `text`. It gets decoded as []byte.

### DIFF
--- a/conn_test.go
+++ b/conn_test.go
@@ -356,7 +356,9 @@ func TestEncodeDecode(t *testing.T) {
 			NULL::integer,
 			'2000-1-1 01:02:03.04-7'::timestamptz,
 			0::boolean,
-			123,
+			123::smallint,
+			123::integer,
+			123::bigint,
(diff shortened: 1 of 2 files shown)


johto:
> Without this fix scanning a smallint or integer into an interface{}
> will cause the value to be of type int64 igorning the type information
> of the column. Same thing happens for text. It gets decoded as []byte.

Of which both behaviors are correct.  Returning strings or < 64-bit ints from rows.Next() is not allowed.  See e.g. [driver.Value](https://golang.org/pkg/database/sql/driver/#Value).

jvshahid:
Please accept my apologies. I came across this bug and though it's easy to fix, didn't know the driver had a very strict set of types. I guess I should have read the readme. Thanks!

PR #405, opened 2015-10-15
Make copyin.Close() idempotent

freeformz:
This change is causing breakage on tip:
https://github.com/golang/go/issues/12798

I think it's a reasonable upstream change.

Looking at the commit that added the code that is failing on our side
the lack of error propagation is called out and is the reason we're now
failing.
https://github.com/lib/pq/commit/72f84510ec4899677ec603abbaec5ba17a438a8c

This change attempts to preserve old behavior while supporting the fact
that errors are now propagated properly by Stmt.Close().

Open to better ways to handle this though.

### DIFF
--- a/copy.go
+++ b/copy.go
@@ -215,9 +215,7 @@ func (ci *copyin) Exec(v []driver.Value) (r driver.Result, err error) {
 	}
 
 	if len(v) == 0 {
-		err = ci.Close()
-		ci.closed = true
-		return nil, err
+		return nil, ci.Close()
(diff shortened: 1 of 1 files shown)


uhoh-itsmaciek:
Yeah, I think this is legit.

freeformz:
@johto thoughts?

johto:
Yeah, I think given that our interface mandates the Exec(nil) call, this should be fine.  LGTM.

PR #408, opened 2015-11-24
Fix empty query `Columns()` handling

tamird:
A query which returns an empty dataset no longer discards its
`Columns()`.

Best to review this with whitespace suppressed.


johto:
Yeah, this is indeed a bug, though I didn't quite care for combining two unrelated test cases like that.  Pushed in commit 351f9635305934d7e58b3adbbe80c48b6ef4c9c9 (and a follow-up 11fc39a580a008f1f39bb3d11d984fb34ed778d9 since I missed another code path which your code exercised).

Thanks!

tamird:
Thanks!

PR #409, opened 2015-11-25
Allow transactions to begin mid-query

tamird:
This was discovered in some synthetic tests in
github.com/cockroachdb/cockroach where we are running raw SQL rather
than starting transactions in the normal Go way.

Still, this is a bug.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1358,7 +1358,7 @@ func (rs *rows) Next(dest []driver.Value) (err error) {
 		switch t {
 		case 'E':
 			err = parseError(&rs.rb)
-		case 'C', 'I':
+		case 'C', 'I', 'T':
 			continue
 		case 'Z':
 			conn.processReadyForQuery(&rs.rb)
--- a/conn_test.go
+++ b/conn_test.go
@@ -1118,6 +1118,21 @@ func TestCommit(t *testing.T) {
 	}
 }
 
+func TestTransactionMultipleRequest(t *testing.T) {
+	db := openTestConn(t)
+	defer db.Close()
+
+	rows, err := db.Query("SELECT 1; BEGIN; SELECT 2")
+	if err != nil {
+		t.Fatal(err)
+	}
+	for rows.Next() {
+	}
+	if err := rows.Err(); err != nil {
+		t.Fatal(err)
+	}
+}
+
 func TestErrorClass(t *testing.T) {
 	db := openTestConn(t)
 	defer db.Close()


johto:
> Still, this is a bug.

Nope.  The code in question is broken.  In particular, this:

> +	if _, err := db.Exec("COMMIT"); err != nil {
> +		t.Fatal(err)
> +	}

is not guaranteed to be sent to the same connection as the Begin() was.  You absolutely can not do anything like this.  Use *sql.DB.Begin().

tamird:
@johto you're right, but you're missing the point. The bug is exhibited when the two queries _are_ sent to the same connection. In fact, I could remove the `COMMIT` completely, and the bug would still trigger.

tamird:
@johto added a call to `SetMaxOpenConns(1)` to guarantee that the queries all go to the same connection.

cbandy:
> added a call to `SetMaxOpenConns(1)` to guarantee that the queries all go to the same connection.

iirc, the actual connection may be closed and reopened between the `Query` and `Exec`.

What do you expect to happen when a `BEGIN` occurs on a throwaway connection like this?

tamird:
You guys are correct that the semantics of what happens here are murky at best, but it is still a bug for this query to result in a protocol error. Do you disagree?

tamird:
OK, I've removed the COMMIT entirely, to reduce some of the confusion here. Also, for context, here's what happens in this test without the change to `conn.go`:
```
go test ./...
--- FAIL: TestTransactionMultipleRequest (0.00s)
	conn_test.go:1132: pq: unexpected message after execute: 'T'
FAIL
FAIL	github.com/lib/pq	1.023s
ok  	github.com/lib/pq/hstore	0.025s
?   	github.com/lib/pq/listen_example	[no test files]
?   	github.com/lib/pq/oid	[no test files]
```

johto:
OK, I've had a closer look at this, and even the BEGIN is a red herring.  The problem is easily reproduced with just:

```go
    _, _ = db.Query("SELECT 1; SELECT 2")
```

However, the suggested fix is not actually a fix at all.  Consider what the behavior should be, if you do:

```go
    rows, _ := db.Query("SELECT 1; SELECT 2, 'bar'")
    for rows.Next() {
        _ = rows.Scan(/* ??? */)
    }
```

Now suddenly a single "rows" object could contain rows from two completely different-looking result sets, which is obviously not going to work.

Maybe the error in this case could be improved to try and hint that there might be some abuse of db.Query() going on, but there's no actual new functionality to be gained here in any case.  You're using the library in an illegal way, period.

tamird:
Why does 'SELECT 1; SELECT 2;' result in the server sending back 'T' (which usually means pending transaction)?

johto:
> Why does 'SELECT 1; SELECT 2;' result in the server sending back 'T' (which usually means pending transaction)?

I'm not sure where you got that idea from.  If you look at http://www.postgresql.org/docs/9.4/static/protocol-message-formats.html, you'll see that it's a "RowDescription" message, which describes the data types in the data rows that might follow.

tamird:
Ah, I thought that `'T'` was coming from the next server ready message, where it would indicate the transaction status:
> Current backend transaction status indicator. Possible values are 'I' if idle (not in a transaction block); 'T' if in a transaction block; or 'E' if in a failed transaction block (queries will be rejected until block is ended).

tamird:
Thanks for looking into this, I'm going to close it for now, though a better error would certainly be appreciated.

PR #418, opened 2016-01-08
Add support for testing against 9.5

freeformz:


johto:
Is there a reason not to test with PQTEST_BINARY_PARAMETERS=no?

freeformz:
@johto Nope, added.

uhoh-itsmaciek:
So... looks good, but weird Travis failures. I restarted the two builds that failed, but at least one of them seemed to crash Postgres, which isn't great. I tried to repro locally, but I can't can't `gvm` to work correctly...

freeformz:
I also restarted one (unrelated) and it just worked. This is my experience though with the test suite as a whole. I haven't looked into it more though. @uhoh-itsmaciek I'm going to merge this unless there are any objections.

PR #423, opened 2016-01-27
Add verify-ca to error message in conn

richo:
There's a valid branch called "verify-ca" but it's not called out in the error message


uhoh-itsmaciek:
Thanks!

PR #425, opened 2016-02-02
Add support for retrieving multiple results.

petermattis:
Reworked how the msg-ready command (`Z`) is processed. Previously
execution of a query would look for the msg-ready command before
completing the operation. Now, when executing a query, the driver places
the connection into a state where it knows there may be more results. If
another query is subsequently executed, the driver waits for the
msg-ready command to arrive, discarding any other commands, before
sending the new query. But if the special `NEXT` query is executed, the
driver looks for another set of results on the connection. If no such
results are found, ErrNoMoreResults is returned.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -115,6 +115,9 @@ type conn struct {
 	// Whether to always send []byte parameters over as binary.  Enables single
 	// round-trip mode for non-prepared Query calls.
 	binaryParameters bool
+
+	// Whether the connection is ready to execute a query.
+	readyForQuery bool
 }
 
 // Handle driver-side settings in parsed connection string.
@@ -587,6 +590,8 @@ func (cn *conn) gname() string {
 }
 
 func (cn *conn) simpleExec(q string) (res driver.Result, commandTag string, err error) {
+	cn.waitReadyForQuery()
+
 	b := cn.writeBuf('Q')
 	b.string(q)
 	cn.send(b)
@@ -614,11 +619,16 @@ func (cn *conn) simpleExec(q string) (res driver.Result, commandTag string, err 
 func (cn *conn) simpleQuery(q string) (res *rows, err error) {
 	defer cn.errRecover(&err)
 
+	cn.waitReadyForQuery()
+
+	// Mark the connection has having sent a query.
+	cn.readyForQuery = false
 	b := cn.writeBuf('Q')
 	b.string(q)
 	cn.send(b)
 
 	for {
+
 		t, r := cn.recv1()
 		switch t {
 		case 'C', 'I':
@@ -742,6 +752,8 @@ func (cn *conn) Prepare(q string) (_ driver.Stmt, err error) {
 	}
 	defer cn.errRecover(&err)
 
+	cn.waitReadyForQuery()
+
 	if len(q) >= 4 && strings.EqualFold(q[:4], "COPY") {
 		return cn.prepareCopyIn(q)
 	}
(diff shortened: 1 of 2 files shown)


uhoh-itsmaciek:
This is very clever, and I can see the utility, but I'm really skeptical of adding hacks like this... I'd like to hear opinions from other committers or the community.

petermattis:
Besides adding new functionality, this fixes the processing of results in queries that contain multiple statements. Consider the query `SELECT 1; SELECT 2, 3;`. The `SELECT 1` statement will return a single row (`D`) followed by a command complete (`C`). The second statement (`SELECT 2, 3`) will then send a row description (`T`) which will cause the processing of rows in `rows.Next` to error (`unexpected message after execute`) and leave the connection in a bad state.

I'm not terribly thrilled with the usage of `NEXT` as an interface to retrieve multiple results, but I'm not seeing any other options. The `database/sql` package doesn't give any access to driver data structures AFAICT.

johto:
> I'm not terribly thrilled with the usage of NEXT as an interface to retrieve multiple results, but I'm not seeing any other options. The database/sql package doesn't give any access to driver data structures AFAICT.

I really would rather not see hacks like this.  If database/sql doesn't support something, why not suggest improvements over there?

petermattis:
There is an issue filed upstream about this: https://github.com/golang/go/issues/12382

johto:
> Besides adding new functionality, this fixes the processing of results in queries that contain multiple statements.

Now that might be worth fixing.  I'm not sure whether we should throw away result sets after the first one (the opposite of what e.g. libpq does), or whether to simply clean up correctly and return an error.

Has anyone checked what the behavior of libpq with the streaming interface is?

kardianos:
Support for multiple result sets has been merged into `database/sql` now.

tamird:
Note that such support will need to live behind an appropriate build tag, and has no business being merged before 1.8 is out. It'd be cool to see it working, though.

kardianos:
@tamird On the contrary, it may be implemented as additional methods and the sql package will detect support. No build tags are required.

Also, I have yet to make this general request, but I'll start with the request here:

There are many new features in `database/sql`. Please oh PLEASE start adding support for them November 1 when the go1.8 freeze starts and have some idea of your implementation before the first 1.8 pre-release comes out. Once go1.8 is released, this stuff is frozen solid. If a design mistake was made I want to know before the first pre-release. Thanks.

tamird:
Ah, of course you're right. It would be great if there was somewhere to look at a live list of those updates.

Here's the few that I've found:
https://github.com/golang/go/commit/e13df02e5fbf2c0cd8811b826a8c8567efa882dd
https://github.com/golang/go/commit/86b2f29676c52774d91dda96e0ba5d4d7bcd3b47
https://github.com/golang/go/commit/707a83341b8c7973f4e0fce731fa279c618f233b

kardianos:
@tamird I'll get a list together within a week or two and include it in a more general announcement. Maybe I'll start with opening a single issue on each database drivers that link to a living google docs that contains a list to all the different interfaces / descriptions.

freeformz:
That would be awesome.

c4milo:
Is this PR still relevant with the upcoming Go 1.8 release supporting multiple result sets?

mjibson:
No, this PR isn't needed. We previously merged this support with another implementation: https://github.com/lib/pq/pull/532

PR #428, opened 2016-02-07
fix build on go1.6

tamird:
`database/sql.(*Row).Scan()`'s error message has changed slightly.


johto:
This seems a bit over engineered, I think.  It seems that it would be enough to see that the string starts with `sql: Scan error`.

tamird:
@johto done.

tamird:
@johto this is green, PTAL

johto:
LGTM, thanks!

PR #430, opened 2016-02-08
Remove unnecessary allocation

tamird:
Submitting on behalf of @petermattis.

This commit is extracted from #425.

@johto


tamird:
This is green.

freeformz:
Doesn't this loose the Columns and other data on queries which return an empty data set ?

tamird:
@freeformz what do you mean? `st` never has anything useful initialized on it; there is no functional change here, except for saving an allocation.

freeformz:
You are totally correct. My bad.

johto:
LGTM, pushed in fde2b33cb1c820fbd3ff19eb9414be2d86b1bd12.  Thanks!

PR #431, opened 2016-02-08
sslmode=require compliance when root cert provided

### LINKED ISSUE #424: require tls mode
in `ssl`, it says the default is `"require"`, but inspection of the code reveals that `"require"` doesn't verify peer. Is this expected behaviour, or am I reading this wrong?

EDIT: Forgot to link to source: https://github.com/lib/pq/blob/master/conn.go#L972-L1013

tamird:
Fixes #424.


tamird:
cc @richo

conn.go
```diff
@@ -973,8 +973,21 @@ func (cn *conn) ssl(o values) {
...
+		// We must skip TLS's own verification since it requires full
+		// verification since Go 1.3.
 		tlsConf.InsecureSkipVerify = true
+
+		// From http://www.postgresql.org/docs/current/static/libpq-ssl.html:
+		// Note: For backwards compatibility with earlier versions of PostgreSQL, if a
+		// root CA file exists, the behavior of sslmode=require will be the same as
+		// that of verify-ca, meaning the server certificate is validated against the
+		// CA. Relying on this behavior is discouraged, and applications that need
+		// certificate validation should always use verify-ca or verify-full.
```
richo: `verify-full` is basically the only coherent option if you want security, since verify-ca will accept **any** certificate in a public CA chain (And they're super easy to get now that letsencrypt and friends exist).

Can this just say "verify-full" ?

richo:
One comment about the recommendation, but this looks good :+1:

Thanks!

conn.go
tamird: Per the discussion in #424, this is just emulating the behaviour of `libpq`.

tamird:
Hm, postgres source suggests that they also allow sslrootcert to point to a nonexistent file when using `sslmode=require`. See https://github.com/postgres/postgres/blob/c8642d909fdd57c36dd71e0b0bb4071523324794/src/test/ssl/t/001_ssltests.pl#L103:L108

That's going to be a more invasive change. Do we want that?

conn.go
johto: verify-ca with sslrootcert is still fine.

richo: To be clear, this comment doesn't have behaviour, right?

Inasfar as libpq has a fairly longwinded and somewhat convincing argument, unless we're planning to include the entire thing, and be explicit that you're only buying yourself security from a passive adversary, it seems ok to have an Opinion in this comment to me at least

tamird:
Ping

cbandy:
I don't like the vagueness of the docs regarding this; e.g., "if a root CA file exists", "If a root CA file is present". It looks like they mean a successful call to [`stat()`](https://github.com/postgres/postgres/blob/c8642d909fdd57c36dd71e0b0bb4071523324794/src/interfaces/libpq/fe-secure-openssl.c#L1208-L1216).

I think adding this check isn't so invasive. If empty string behaves okay, L988-L990 could change to:

```go
if _, err := os.Stat(o.Get("sslrootcert")); err == nil {
  verifyCaOnly = true
}
```

tamird:
@cbandy done.

tamird:
@johto could you take a look?

uhoh-itsmaciek:
@tamird this looks good to me. Any thoughts @freeformz or anyone else?

tamird:
👋

uhoh-itsmaciek:
Rebased and merged. Thanks!

PR #432, opened 2016-02-09
fix typo

shawnps:


johto:
According to https://en.wiktionary.org/wiki/acknowledgement both are correct.  I don't see the point of changing this.

PR #433, opened 2016-02-09
fix typo

shawnps:


shawnps:
https://en.wiktionary.org/wiki/precede

"This word is commonly misspelt as preceed."

johto:
Thanks!

Issue #404, opened 2015-10-13
Document mapping between postgres (oid) and golang types


hnsl:
Right now I had to read (uncommented) source code for example to understand that null values when scanned/read will map to a golang nil value. This is not an obvious choice. A design where you had chosen to map the null value to a corresponding NullTime, NullInt64 etc with Valid = false (preserving the oid type information) would have been just as likely from my perspective.

johto:
> A design where you had chosen to map the null value to a corresponding NullTime, NullInt64 etc with Valid = false (preserving the oid type information) would have been just as likely from my perspective.

That's not up to us to decide if we want to conform to the database/sql interface.  See e.g. [driver.Value](https://golang.org/pkg/database/sql/driver/#Value) and [driver.Rows.Next](https://golang.org/pkg/database/sql/driver/#Rows).

cbandy:
The `sql` package also mentions that `nil` represents `NULL`, but not very prominently: https://golang.org/pkg/database/sql/#Scanner

hnsl:
The "database/sql/driver" documentation documents the interface for drivers which is irrelevant for users of "lib/pq" and "database/sql". It doesn't document an interface "end users" actually use.

This is from the (rs *Rows) Scan documentation:
"If an argument has type *interface{}, Scan copies the value provided by the underlying driver without conversion."

The above means "lib/pq" needs to document what Postgres types map to what golang types. This is currently impossible to know without reading the source code of lib/pq which means there is a hole in the documentation. Currently you have to rely on undocumented behaviour if you want to scan data into an interface{} pointer.

hnsl:
I don't seem to have the permission to reopen a ticket so I hereby declare my intent to reopen this ticket if I could since this information would get lost otherwise.

johto:
> The above means "lib/pq" needs to document what Postgres types map to what golang types. This is currently impossible to know without reading the source code of lib/pq which means there is a hole in the documentation. Currently you have to rely on undocumented behaviour if you want to scan data into an interface{} pointer.

Yeah, that's an actual thing we could improve.

PR #436, opened 2016-02-11
encode: `decode` T_text into a string

tamird:
This allows the caller to distinguish strings from byte arrays.

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -83,6 +83,8 @@ func binaryDecode(parameterStatus *parameterStatus, s []byte, typ oid.Oid) inter
 
 func textDecode(parameterStatus *parameterStatus, s []byte, typ oid.Oid) interface{} {
 	switch typ {
+	case oid.T_char, oid.T_varchar, oid.T_text:
+		return string(s)
 	case oid.T_bytea:
 		return parseBytea(s)
 	case oid.T_timestamptz:
--- a/encode_test.go
+++ b/encode_test.go
@@ -575,6 +575,17 @@ func TestBinaryByteSliceToInt(t *testing.T) {
 	}
 }
 
+func TestTextDecodeIntoString(t *testing.T) {
+	input := []byte("hello world")
+	want := string(input)
+	for _, typ := range []oid.Oid{oid.T_char, oid.T_varchar, oid.T_text} {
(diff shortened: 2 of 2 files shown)


johto:
I had explained in some issue why we're not allowed to do this, but I can't find it anymore.  The only thing I could find was cbandy's comment at the end of #300.

tamird:
Well, per https://go-review.googlesource.com/#/c/19439/ this isn't going anywhere.

tamird:
Reopened as an alternative to #300 now that https://go-review.googlesource.com/#/c/19439/ is merged in https://github.com/golang/go/commit/7162c4d05c1fed9db517c92b9a5232243ec453a2.

tamird:
@johto ping!

cbandy:
Excellent. We should not forget the other [character types](http://www.postgresql.org/docs/current/static/datatype-character.html). As we make this transition (or maybe before?) we should probably tackle #404, so we can point to the docs if/when users notice this change.

Based on comments in the CL, it looks like the change to `database/sql` will be in Go1.7, yes? Is there any reason/desire for us to keep the current/pre-Go1.7 behavior available?

tamird:
I'll take a look at #404 when I have more time, but I don't see any reason
for it to precede this (in fact, doing this first will reduce churn).

As for persevering the old behavior - I don't think there's any reason to
do so. It's easy enough to work around by scanning directly into a byte
array or casting (and it's just as much work as changing existing user code
to use a legacy API).

tamird:
@uhoh-itsmaciek could you take a look at this one? the upstream change was a documentation change, so while it will be included in 1.7, it should be safe to merge this even for use in versions prior to 1.7.

uhoh-itsmaciek:
@tamird sure, but I won't have time for it till next week.

freeformz:
@tamird How about `T_varchar` & `T_char` ?

tamird:
seems reasonable to add those. Done.

tamird:
@uhoh-itsmaciek this is green and ready for another look, when you get a chance.

freeformz:
LGTM

uhoh-itsmaciek:
Looks good. Thanks!

PR #440, opened 2016-03-01
Add 1.6 to travis for testing

garrypolley:
Having an explicit test for 1.6 will help me out to know that 1.6 is working as desired


uhoh-itsmaciek:
Thanks, I'll merge this as soon as Travis passes. It's worth noting that our test matrix is getting gynormous, but we definitely do want to keep testing recent releases, so if there's paring down to be done, it should be independent from this.

PR #442, opened 2016-03-02
Drop antediluvian Postgres and Go versions

uhoh-itsmaciek:
This drops testing for the ancient Postgres 8.4 and not quite as ancient but also very old Go 1.1 and Go 1.2.  Postgres 9.0 and Go 1.3 and 1.4 are also unsupported upstream, but I figure we can drop the super-old stuff first. The project doesn't have an official versioning policy, but the Travis builds take an absurd amount of time due to the breadth of the matrix (#440 took something like eight hours including waiting for the Travis job queue), so it might be worth paring down the test matrix.


cbandy:
:thumbsup:

freeformz:
:+1: FWIW: I would also drop 1.3 at this time and then 1.4 sometime after 1.7 is out.

uhoh-itsmaciek:
Dropped go1.3 as well.

PR #447, opened 2016-03-16
Expose ParseTimestamp for parsing the postgres timestamp format

paperstreet:
Make it return errors where it used to panic, and add a couple tests for things that used to panic.

For https://github.com/cockroachdb/cockroach/issues/4872.

### DIFF
--- a/encode_test.go
+++ b/encode_test.go
@@ -72,31 +72,38 @@ var timeTests = []struct {
 	{"123456-02-03 04:05:06.1", time.Date(123456, time.February, 3, 4, 5, 6, 100000000, time.FixedZone("", 0))},
 }
 
-// Helper function for the two tests below
-func tryParse(str string) (t time.Time, err error) {
-	defer func() {
-		if p := recover(); p != nil {
-			err = fmt.Errorf("%v", p)
-			return
-		}
-	}()
-	i := parseTs(nil, str)
-	t, ok := i.(time.Time)
-	if !ok {
-		err = fmt.Errorf("Not a time.Time type, got %#v", i)
-	}
-	return
-}
-
 // Test that parsing the string results in the expected value.
 func TestParseTs(t *testing.T) {
 	for i, tt := range timeTests {
-		val, err := tryParse(tt.str)
+		val, err := ParseTimestamp(nil, tt.str)
 		if err != nil {
 			t.Errorf("%d: got error: %v", i, err)
 		} else if val.String() != tt.timeval.String() {
 			t.Errorf("%d: expected to parse %q into %q; got %q",
 				i, tt.str, tt.timeval, val)
+		}
+	}
+}
+
+var timeErrorTests = []string{
+	"2001",
+	"2001-2-03",
+	"2001-02-3",
+	"2001-02-03 ",
+	"2001-02-03 04",
+	"2001-02-03 04:",
+	"2001-02-03 04:05",
+	"2001-02-03 04:05:",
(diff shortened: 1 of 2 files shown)


encode_test.go
```diff
@@ -101,6 +85,29 @@ func TestParseTs(t *testing.T) {
...
+	"2001",
+	"2001-2-03",
+	"2001-02-3",
+	"2001-02-03 ",
+	"2001-02-03 04",
+	"2001-02-03 04:",
+	"2001-02-03 04:05",
+	"2001-02-03 04:05:",
+	"2001-02-03 04:05:6",
+	"2001-02-03 04:05:06.123 B",
```
uhoh-itsmaciek: Are these values that are ever returned by Postgres for these types? If not, what is the performance impact of these changes?

paperstreet: These are values which would cause a panic before. I originally made this change in the [cockroach fork](https://github.com/cockroachdb/pq/commit/17d569dbc13f29cbe32b50e64d2046648cd9684e) because it was being called from a place where panics were unacceptable and the overhead of wrapping it in a deferred recover was too much.

I ran all the benchmarks and nothing changed except this one:

    name                 old time/op    new time/op    delta
    DecodeTimestamptz-8     728ns ± 2%     760ns ± 1%  +4.36%        (p=0.000 n=13+13)
    
    name                 old alloc/op   new alloc/op   delta
    DecodeTimestamptz-8     64.0B ± 0%     64.0B ± 0%    ~     (all samples are equal)
    
    name                 old allocs/op  new allocs/op  delta
    DecodeTimestamptz-8      2.00 ± 0%      2.00 ± 0%    ~     (all samples are equal)

cbandy:
Rather than a function for this, I think it would be more idiomatic to have and object that implemented Scanner.

tamird:
@cbandy could that request not block this change? I agree with you in spirit, but it seems an onerous burden to place on @paperstreet.

paperstreet:
Ah, sorry for the delay in responding. I think that's a good idea @cbandy, but since this commit was just to turn errors into panics (you'll note that I tried to keep the structure as similar as possible), it feels a bit out of scope for this change.

cbandy:
> @cbandy could that request not block this change? I agree with you in spirit, but it seems an onerous burden to place on @paperstreet.

I'm not able to block anything, afaik; just giving my opinion.

It looks like cockroach is getting down to the postgres wire protocol. Have you considered using the binary format for temporal data? Converting it to a `time.Time` is some [arithmetic](https://github.com/jackc/pgx/blob/v2.8.0/values.go#L1391-L1393).

(See [_send and _recv functions](https://github.com/postgres/postgres/blob/a70e13a39eccf5fc944c66e0029004b6abcb3cae/src/backend/utils/adt/timestamp.c#L230-L291) on the backend. How to interpret the 8 bytes depends how the backend was [compiled with integer-datetimes](https://github.com/postgres/postgres/blob/98a64d0bd713cb89e61bef6432befc4b7b5da59e/configure.in#L130-L132), but it has been an option for packagers since 7.3 and enabled by default since 8.4.)

tamird:
Supporting the binary format for temporal data seems like a good idea (thanks for the pointer!), but doesn't obviate the need to also support the text format.

I suppose I was under the false impression that you were a collaborator.

@uhoh-itsmaciek could you take a look?

uhoh-itsmaciek:
I ran the benchmarks myself and  got the same result--not much change. Looks good, thanks for the patch!

PR #449, opened 2016-03-28
Export FormatTimestamp for emitting the postgres timestamp format

tamird:
This is a dual to ParseTimestamp exported in 3cd0097, and is needed
for CockroachDB.

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -336,6 +336,10 @@ func parseTs(currentLocation *time.Location, str string) interface{} {
 	return t
 }
 
+// ParseTimestamp parses Postgres' text format. It returns a time.Time in
+// currentLocation iff that time's offset agrees with the offset sent from the
+// Postgres server. Otherwise, ParseTimestamp returns a time.Time with the
+// fixed offset offset provided by the Postgres server.
 func ParseTimestamp(currentLocation *time.Location, str string) (time.Time, error) {
 	p := timestampParser{}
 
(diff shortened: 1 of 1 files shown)


encode.go
```diff
@@ -444,6 +444,10 @@ func formatTs(t time.Time) (b []byte) {
 			return []byte("infinity")
 		}
 	}
+	return FormatTimestamp(t)
+}
+
+func FormatTimestamp(t time.Time) []byte {
```
uhoh-itsmaciek: If this is to be a public function, it should probably be documented (as should `ParseTimestamp`--I missed that originally) and tested. The existing tests cover this okay I think, but do you mind adding docs?

tamird: Done! Also added a comment for `ParseTimestamp`.

freeformz:
Shouldn't the infinity bits also be part of the FormatTimestamp?

tamird:
I'm not sure. I was following the example of ParseDuration which doesn't deal with infinity.

uhoh-itsmaciek:
@tamird thanks for the doc comments. @freeformz how so?

freeformz:
@uhoh-itsmaciek Maybe not. I guess I don't understand why the exported functions don't handle it, yet EnableInfinityTs is exported.

tamird:
@uhoh-itsmaciek this good to go?

uhoh-itsmaciek:
@tamird any thoughts on @freeformz's comment?

uhoh-itsmaciek:
That makes sense to me. I'll merge in a bit.

tamird:
Ping!

uhoh-itsmaciek:
Rebased and merged. Thanks!

PR #453, opened 2016-04-02
Teach driver how to handle uint64 by  implementing ColumnConverter. 

giulioungaretti:
This allows one to actually store and retrieve uint64 values stored as numeric  in postgres.

The idea is taken from the default database/sql (https://golang.org/src/database/sql/driver/types.go#L213) .

I would like to write some tests for it but running any test (with master branch)  just returns this :
```
--- FAIL: TestPgpass (0.00s)       conn_test.go:163: , unexpected error <nil> 
```

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -17,6 +17,7 @@ import (
 	"os/user"
 	"path"
 	"path/filepath"
+	"reflect"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


arp242:
This was added in #1125 and #1238 with CheckNamedValue(), which does the same as the (now-deprecated) ColumnConverter().

PR #455, opened 2016-04-07
Check for postgres driver before registering

leetrout:
Please forgive me for not having a good test case for this on hand but I wanted to get this PR open for discussion as quickly as possible.

With Go 1.6 (I assume 1.5 with `GO15VENDOREXPERIMENT=1` too) we started experiencing issues where vendored code that used lib/pq would cause an error in our main application that also used lib/pq.

The error was `panic: sql: Register called twice for driver postgres`.

The full traceback is

```
goroutine 1 [running]:
panic(0x808ba0, 0xc820102400)
	/usr/local/go/src/runtime/panic.go:464 +0x3e6
database/sql.Register(0x9873e0, 0x8, 0x7f242d4812f0, 0xc24748)
	/usr/local/go/src/database/sql/sql.go:45 +0x181
github.com/yhat/repo/src/vendor/github.com/yhat/events/vendor/github.com/lib/pq.init.1()
	/home/ubuntu/.go_workspace/src/github.com/yhat/repo/src/vendor/github.com/yhat/events/vendor/github.com/lib/pq/conn.go:44 +0x71
github.com/yhat/repo/src/vendor/github.com/yhat/events/vendor/github.com/lib/pq.init()
	/home/ubuntu/.go_workspace/src/github.com/yhat/repo/src/vendor/github.com/yhat/events/vendor/github.com/lib/pq/user_posix.go:24 +0x4ee
github.com/yhat/repo/src/vendor/github.com/yhat/events/events.init()
	/home/ubuntu/.go_workspace/src/github.com/yhat/repo/src/vendor/github.com/yhat/events/events/events.go:383 +0x68
github.com/yhat/repo/src/app.init()
	/home/ubuntu/.go_workspace/src/github.com/yhat/repo/src/app/deploy_test.go:30 +0xd3
main.init()
	github.com/yhat/repo/src/app/_test/_testmain.go:56 +0x4a
exit status 2
FAIL	github.com/yhat/repo/src/app	0.041s
```

The directory structure we had in place looks like:

```
.
├── src
│   ├── app (imports github.com/lib/pq & github.com/yhat/events/events
│   ├── ...
└── ├── vendor
├── github.com
│   ├── lib
│   │   └── pq
│   └── yhat
│       ├── events
│       │   ├── events
│       │   │   ├── events.go (imports github.com/lib/pq)
│       │   ├── main.go
│       │   └── vendor
│       │       └── github.com
│       │           ├── lib
│       │           │   └── pq
```

Referencing the the structure above:
`src/app/db.go` imports `github.com/lib/pq`
`src/app/events.go` imports `github.com/yhat/events/events`
`vendor/github.com/yhat/events/events/events.go` imports `github.com/lib/pq` which is also vendored within events.

In this case it appears the `init()` was being invoked twice and producing the error.

@bajh suggested this change; it simply checks to see if `"postgres"` is already registered before calling register. I've rolled out the patch on our fork and it did, indeed, solve the problem.

I suspect most people are not using vendoring and haven't encountered this issue yet.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -41,7 +41,15 @@ func (d *drv) Open(name string) (driver.Conn, error) {
 }
 
 func init() {
-	sql.Register("postgres", &drv{})
+	register := true
+	for _, d := range sql.Drivers() {
+		if d == "postgres" {
+			register = false
+		}
+	}
+	if register {
+		sql.Register("postgres", &drv{})
+	}
 }
 
 type parameterStatus struct {


johto:
> vendor/github.com/yhat/events/events/events.go imports github.com/lib/pq which is also vendored within events.

But with this change, only one of them would be imported and I might be wrong, but it's not deterministic which one.  This effectively renders one of the vendored drivers unnecessary, so why do that at all?  Why not just vendor a single copy for the entire application?

leetrout:
I see what you're saying. Do you suspect that this could come up more often
as vendoring becomes more popular?

This would be an easy change for us since the issue involves 2 packages in
our control but does that address the larger pattern of nested vendored
libs going forward?

On Thursday, April 7, 2016, Marko Tiikkaja <notifications@github.com> wrote:

> vendor/github.com/yhat/events/events/events.go imports github.com/lib/pq
> which is also vendored within events.
>
> But with this change, only one of them would be imported and I might be
> wrong, but it's not deterministic which one. This effectively renders one
> of the vendored drivers unnecessary, so why do that at all? Why not just
> vendor a single copy for the entire application?
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/lib/pq/pull/455#issuecomment-206970166>
>


-- 
Sorry I fat thumbed this on an iPhone

freeformz:
it's my opinion that the tool that does the vendoring should flatten to a single top level vendor directory in most cases. Nested vendor folders just complicate reasoning about a code base and are unnecessary 99% of the time.

leetrout:
> the tool that does the vendoring

So we're just following the quasi-pattern established for `GO15VENDOREXPERIMENT` and using submodules**. So there's no tool that actually created or manages the vendored folders and the usage or consumption of those is just within Go itself.

Do you think this should be brought up on the Go mailing list? Perhaps identifying this pattern where register is being called within init and there appears to be no helpful way to identify / limit the scope of how this interaction takes place. I don't want to make a mountain out of a molehill, either, so acknowledging I'm really new to Go it seems like there's still a missing piece. (Perhaps submodules aren't appropriate.**)


**
https://docs.google.com/document/d/1Bz5-UB7g2uPBdOx-rw5t9MxJwkfpx90cqG9AFL0JAYo/edit
> The environment variable also enables fetching of git submodules during “go get”. This is meant to allow experiments to understand whether git submodules are an appropriate and useful way to vendor code.

freeformz:
I never noticed that the doc said that, possibly because I think git submodules are the worst and didn't want to afflict that on myself. :-(

FWIW: Myself and most gophers I know don't think that was is a good idea, but it *does* work.

cbandy:
From my time with Ruby and its flavor of dependency hell, I learned that things that get imported (gems, packages) should set almost no restrictions on their dependencies. It is the job of something that actually runs (an application) to coordinate the versions in use.

Do not vendor in packages; vendor at `main`.

leetrout:
That makes sense pattern wise.

I think this might be relevant, too https://groups.google.com/forum/#!topic/golang-dev/4FfTBfN2YaI

leetrout:
I'm responding to the Go mailing list and closing this. I appreciate everyone taking a look at this and offering some insight. Seems like the onus is clearly on us to set up / use the vendoring in a different matter.

PR #456, opened 2016-04-15
Basic GSSAPI support

ulfurinn:
Support for GSSAPI authentication using `github.com/apcera/gssapi`. This is pretty much a direct translation from the C client (consult `fe-auth.c` for reference).

This requires cgo and will be stubbed out when cgo is not available.

Two-step authentication is not implemented since I don't know how to trigger this scenario to test it. In my setup authentication succeeds after the first step if there's a valid ticket. It should be trivial to add, I just decided to be cautious.

I couldn't figure out how to write a good unit test for this (glossing over the fact that setting up a kerberized Postgres to run it against is a bit of a pain to begin with).

Test environment:
* server: Ubuntu with MIT Kerberos
* client: Ubuntu with MIT Kerberos, MacOS with system implementation

`github.com/apcera/gssapi` itself is tested against Heimdal.


ulfurinn:
OK, I have no idea how to make Travis happy.

gawth:
Hi, I was wondering if this is something that will ever make it in to master?  2 year old so perhaps not.  Are there any plans to add Kerberos support at any stage?

Thanks
Alan

violet1986:
Kerberos is very important for enterprise user. I implemented the same code to support krb in our own branch. I really hope this PR could go into master.

yydzero:
+1 for Kerberos support!

adam8157:
I suppose this is not going into master because the `apcera/gssapi` requires external libraries but this `lib/pq` is `A pure Go Postgres driver` as quoted.

arp242:
Kerberos/GSSAPI support was added in #970.

PR #458, opened 2016-04-26
Support rumprun GOOS

vsukhin:
Could you please add build tag "rumprun" in order we can build go applications for rump unikernel?


freeformz:
I checked the gorumprun source and it supports `os/user

PR #460, opened 2016-05-06
Support arbitrary network dialers in listeners

tessr:


uhoh-itsmaciek:
Given https://github.com/lib/pq/pull/287, I think this looks reasonable. I'll commit if no one objects. Thanks for the patch!

tessr:
Thanks! Added a comment for the one new exported function.

uhoh-itsmaciek:
Excellent. Merged.

PR #461, opened 2016-05-10
Implement `go vet` fix.

gmccue:
`go vet ./...` complains with the following message:
```go
vendor/github.com/lib/pq/encode.go:25: unreachable code
vendor/github.com/lib/pq/encode.go:78: unrecognized printf verb 'u'
```


freeformz:
the panic was there for older versions of go, but lib/pq doesn't test against them anymore, so +1 from me.

uhoh-itsmaciek:
Agreed; merging. Thanks for the patch!

uhoh-itsmaciek:
I added a PR running `go vet` to Travis in #462 to make sure things stay clean.

PR #462, opened 2016-05-10
Have Travis run go vet

uhoh-itsmaciek:

PR #463, opened 2016-05-10
Add go fmt to travis

uhoh-itsmaciek:
If we add `go vet` as per #462 this probably makes sense, too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,7 +59,7 @@ env:
     - PGVERSION=9.0 PQTEST_BINARY_PARAMETERS=no
 
 script:
- - go vet ./... && go test -v ./...
+ - result=$(gofmt -d -e $(find -name \*.go)); test -z "$result" || (echo "$result" && false) && go vet ./... && go test -v ./...
 
 before_script:
  - psql -c 'create database pqgotest' -U postgres
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,7 +59,7 @@ env:
     - PGVERSION=9.0 PQTEST_BINARY_PARAMETERS=no
 
 script:
- - go test -v ./...
+ - go vet ./... && go test -v ./...
 
 before_script:
  - psql -c 'create database pqgotest' -U postgres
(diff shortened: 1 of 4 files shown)


freeformz:
I would prefer goimports, so that imports are also sorted / ordered.

uhoh-itsmaciek:
Good call. Is goimports available by default on Travis? Or can I `go get` it in the `before_install` step?

freeformz:
it requires a `before_install` go get.

.travis.yml
```diff
@@ -59,7 +59,7 @@ env:
     - PGVERSION=9.0 PQTEST_BINARY_PARAMETERS=no
 
 script:
- - go test -v ./...
+ - result=$(gofmt -d -e $(find -name \*.go)); test -z "$result" || (echo "$result" && false) && go test -v ./...
```
cbandy: Rather than a variable and `test`, could pipe the output to `awk`:

```
gofmt -d -e $(find -name \*.go) | awk '{ print } END { exit NR == 0 ? 0 : 1 }' && go test -v ./...
```

uhoh-itsmaciek: I tried this but it doesn't seem to work.

uhoh-itsmaciek:
@freeformz moved to goimports and seems to be working

.travis.yml
uhoh-itsmaciek: @cbandy I'm going to merge as is since this seems to be working, but feel free to submit a PR with the awk mechanism. If you can get a PR to fail (by intentionally introducing an error) and then pass, I don't particularly care what the command looks like, and awk seems simpler here.

cbandy: Indeed! The pipe character is significant in YAML. You can use the [folded style](http://www.yaml.org/spec/1.2/spec.html#id2796251) for this:

```yaml
script: >
  goimports -d -e $(find -name \*.go) | awk '{ print } END { exit NR == 0 ? 0 : 1 }' && go test -v ./...
```

uhoh-itsmaciek: Ah, okay. I'm not going to tweak this further since it works and I don't feel like playing Travis golf ;) but if you want to refine it, I'll accept a PR.

Issue #464, opened 2016-05-17
Array type support, like that jackx/pgx


dxvgef:
jackc/pgx support query postgresql array type and json/jsonb type, hope lib/pq can support.
I really need this feature, thank you!

freeformz:
Dupe of #327 (which may be a dupe of #49) ?

uhoh-itsmaciek:
Agreed, although if pgx already supports arrays, it may be worth looking at their interface.

freeformz:
@uhoh-itsmaciek :+1: I was looking at those issues on Monday and trying to make heads or tails of the various proposals and their status.

johto:
json and jsonb have always worked.  You just get the value back as a string and feed that to `encoding/json`.  I just merged array support from #466.

Issue #49, opened 2012-10-23
Support Postgresql array types


github-dr:
Is it possible to support array types?
http://www.postgresql.org/docs/9.2/static/arrays.html

From a cursory look at encode.go, both encode() and decode() functions would need to be extended to take or return a slice of any of the current types they handle.  Perhaps other parts need changing as well?    

Do you think it is feasible to extend this library to support arrays, or would I be better to use the cgo wrapper around libpq for such native/non-standard SQL support? (though ARRAY is actually part of SQL99 standard, so maybe other DBs that don't support it should be called non-std :-)

Ideally, I'd like to do something like:

db.Exec("INSERT INTO t (k, v) VALUES ($1, $2)", "key", []string{"value1", "value2"})

Thanks,

ptman:
NULLable columns can be represented as pointers in Go. E.g. []*time.Time for array of nullable timestamps

kevinburke:
Should this issue be closed now? It looks like #302 adds support for this, and got merged.

kevinburke:
Sorry - the commit merged in #302 seems to be a red herring, this is still unsupported.

freeformz:
I think this can be closed since #466 was merged. Please re-open this issue if I am wrong.

PR #466, opened 2016-05-26
Array support with delimiter interface

cbandy:
This adds types that implement `driver.Valuer` and `sql.Scanner` for arrays and slices.

There are a handful of "native" types that efficiently handle one-dimensional slices (e.g. `StringArray` for `[]string`) and a "generic" type that uses reflection to traverse multi-dimensional arrays/slices.

The `ArrayDelimiter` interface can be implemented by a type that does not use comma as [its delimiter](https://www.postgresql.org/docs/current/static/arrays.html#ARRAYS-IO). For example,

```go
type Box struct { ... }
func (b Box) Value() (driver.Value, error) { ... }
func (b Box) ArrayDelimiter() string { return ";" }
```

The most convenient way to use these types is probably the `Array` function. It provides a consistent way to send and receive arrays and slices of various types. For example,

```go
db.Query(..., pq.Array([]string{ ... }))
db.Query(..., pq.Array([][]int{ ... }))

var x []sql.NullInt64
var y []string
db.QueryRow(...).Scan(pq.Array(&x))
db.QueryRow(...).Scan(pq.Array(&y))
```

See also #49, #130, #327, #353, #364, #464.

### DIFF
--- a/array.go
+++ b/array.go
@@ -3,8 +3,10 @@ package pq
 import (
 	"bytes"
 	"database/sql/driver"
+	"encoding/hex"
 	"fmt"
 	"reflect"
+	"strconv"
 )
 
 var typeByteSlice = reflect.TypeOf([]byte{})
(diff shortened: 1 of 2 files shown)


cbandy:
`GenericArray.Scan()` is artificially limited to handle only one-dimensional array/slice of `sql.Scanner` as discussed toward the end of #364. I simply ran out of time to handle all the possible combinations of types and dimensions. Fortunately, those limitations can be lifted (gradually, even) without changing the exported API in any way.

freeformz:
@johto Any thoughts on this?

johto:
I'll have a closer look at this over the weekend.

My memory isn't that great, but I thought there were serious issues discussed with these struct definitions in #364, which is why I suggested doing:

```Go
type Int64Array struct { A *[]int64 }
```

instead.  Have those issues been fixed here?

cbandy:
> My memory isn't that great, but I thought there were serious issues discussed with these types of struct definitions

> Have those issues been fixed here?

I think you're referring to things implementing both `Scanner` and `Valuer`. The `Array` function returns pointers to achieve this, e.g. `return (*BoolArray)(&a)`. The `GenericArray` struct behaves as you've described.

cbandy:
> My memory isn't that great, but I thought there were serious issues discussed with these types of struct definitions

> Have those issues been fixed here?

I think you're referring to things implementing both `Scanner` and `Valuer`. The `Array` function returns pointers to achieve this, e.g. `return (*BoolArray)(&a)`. The `GenericArray` struct behaves as you've described.

Yes, I'd say those issues have been fixed.

swsnider:
Any chance of this getting merged soon? I'm going to have to hold this patch in a private fork otherwise :(.

ptrimble:
+1 on merging. We would like to use this improvement in our internal projects.

johto:
OK, I'm working on this right now.

johto:
I've pushed this in

ptrimble:
Thank you @johto!

cbandy:
@johto, thank you for reviewing and merging this. A lot of people expressed interest, but you're the only one to comment on the actual patches here. Thank you for taking care of the issues with shadowing and readability.

It wasn't my intention to leave you with code that you find hard to maintain. I want my contributions to this project to positive for both users and maintainers alike. I'm sorry for leaving this branch open for so long and I'm sorry if I haven't seemed open to feedback.

Also, thank you to @erykwalder. This is absolutely based on your work in #353 I had forgotten how many of your suggestions in #364 I had incorporated here.

PR #468, opened 2016-06-09
Clean up .travis.yml and reduce build time

cbandy:
With four Go versions, six PostgreSQL versions, and one connection parameter, we are generating *48* jobs during every build in Travis. While each job takes only a minute or two, the vast majority of time is spent doing the setup. (`goimports`, `go vet`, and `go test` take only 3 to 4 seconds combined.)

Moving `PQTEST_BINARY_PARAMETERS` to the `script` step reduces the number of jobs to 24 and nearly halves the CPU time, from [72 minutes](https://travis-ci.org/lib/pq/builds/129461243) to [38 minutes](https://travis-ci.org/lib/pq/builds/136559221). 💡

This changes how the build behaves during a failure. Without this PR, `go test` only runs when both `goimports` and `go vet` succeed. Now, `goimports`, `go vet`, and `go test` run sequentially regardless of each other's result. [The build fails when any of them fail.](https://docs.travis-ci.com/user/customizing-the-build#Customizing-the-Build-Step)

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -1,0 +1,73 @@
+#!/bin/bash
+
+set -eu
+
+client_configure() {
+	sudo chmod 600 $PQSSLCERTTEST_PATH/postgresql.key
+}
+
+pgdg_repository() {
+	local sourcelist='sources.list.d/postgresql.list'
+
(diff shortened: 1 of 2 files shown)


freeformz:
:+1: LGTM.

freeformz:
I'll merge it later if no one dissents.

PR #479, opened 2016-08-07
Run goimports only when installed

cbandy:
As expected upstream in golang/go \#16434, `goimports` is incompatible with Go 1.4.

This [Travis build](https://travis-ci.org/cbandy/go-pq/builds/150367748) shows what happens with incorrectly formatted code.


freeformz:
Closing this in favor of #483, which I merged. If we really need to test again go1.4 again (I hope not) we can do this instead.

PR #483, opened 2016-08-10
Drop go1.4

### LINKED ISSUE #482: Drop Go 1.4 support
goimports, a tool we use in our .travis.yml setup no longer supports go1.4 because it imports the `runtime/trace` package w/o a build tag guard.

See also: https://github.com/golang/go/issues/16661

freeformz:
Fixes #482

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: go
 
 go:
-  - 1.4
   - 1.5
   - 1.6
   - tip
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: go
 go:
   - 1.5
   - 1.6
-  - tip
+  - 1.7
 
 env:
   global:


cbandy:
I have no need for integration tests against Go 1.4.

I would rather not add a release candidate to the list of build jobs, since we already have jobs for `tip`.

freeformz:
@cbandy I can drop the go1.7rc for now, but tip is going to be open again RSN and I was just prepping things for a final go1.7

freeformz:
I've updated this to include go1.7 (instead of tip). We used to have tip IIRC, which did discover a problem with lib/pq due to a change in database/sql which I fixed. I believe there are other changes coming in the 1.8 time frame, so at some point after the freeze we may want to start testing with tip again to make sure we're good.

freeformz:
@cbandy Doh. I re-read your commend and realized I read it wrong the first time. I'll re-add tip.

freeformz:
I restored tip on a11ab5d56294722de05a36a5082729a80be6eed0

PR #484, opened 2016-08-11
Use space to separate date and time portions of Time literals

cbandy:
Fixes the error in (and motivation for) #391:

> pq: invalid input syntax for type time: "0001-01-01T17:00:00Z BC"

I couldn't find a spec that called for space rather than `T`, but every version of Postgres I tested behaves this way.

I've included a test for the negative case in case Postgres changes and we want to revert this.

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -477,7 +477,7 @@ func FormatTimestamp(t time.Time) []byte {
 		t = t.AddDate((-t.Year())*2+1, 0, 0)
 		bc = true
 	}
-	b := []byte(t.Format(time.RFC3339Nano))
+	b := []byte(t.Format("2006-01-02 15:04:05.999999999Z07:00"))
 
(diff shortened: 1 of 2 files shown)


freeformz:
How did this ever work then?

cbandy:
From what I can tell, we've been using the `RFC3339Nano` format since c61875595da3a75b82799eeff37800db713c1656. I guess we've just never tested sending a `time.Time` into a `time` nor `timetz` type.

freeformz:
:+1:

cbandy:
Changed it to specify the format rather than `Replace` on the result.

freeformz:
even better

mjibson:
@cbandy mind rebasing this to master so we can retest since it's been so long? That should also fix the go 1.4 errors.

cbandy:
Rebased.

PR #497, opened 2016-09-03
Fix possible panic on malformed server response

kostya-sh:


johto:
I'm not sure how I feel about that.  If your server is so broken it sends back invalid responses, would you not want to panic to make sure someone notices?

kostya-sh:
In many (all?) other cases invalid responses do not trigger panics.

freeformz:
FWIW: `errorf` panics. So this doesn't `Fix possible panic` on malformed server responses, it just possibly provides a better panic message?

kostya-sh:
@freeformz, no, this fixes real panic. pq re-panics runtime errors such as bounds out of range: https://github.com/lib/pq/blob/165a3529e799da61ab10faed1fabff3662d6193f/error.go#L482

Issue #450, opened 2016-03-28
Support read timeout for queries


rajeshucsb:
Currently there is no way to specify a read timeout for queries.
The  sql client has an option for statement_timeout
http://www.postgresql.org/docs/9.3/static/runtime-config-client.html

And JDBC has socketTimeout
https://jdbc.postgresql.org/documentation/head/connect.html

Something similar for golang pq driver will be very useful.

johto:
> The sql client has an option for statement_timeout
> http://www.postgresql.org/docs/9.3/static/runtime-config-client.html

That's a server-side option and requires no support in the client.  You can use statement_timeout with `pq` just fine.

> And JDBC has socketTimeout
> https://jdbc.postgresql.org/documentation/head/connect.html

Which is an absolute abomination, and only causes problems in my opinion.

> Something similar for golang pq driver will be very useful.

What's the problem you're trying to solve?

bobbytables:
So I'm looking at using `statement_timeout`, and a go-nuts post has

> Controlling it at pool level is trivial; just set statement_timeout in the second argument to sql.Open().
See: https://groups.google.com/forum/#!topic/golang-nuts/l9vga1aAsHI

You can easily `Exec` the query to set the statement timeout option right when you connect but how would this behave with connection pooling?

johto:
> You can easily Exec the query to set the statement timeout option right when you connect but how would this behave with connection pooling?

It doesn't work.  The connection pool doesn't tell you when it's connecting or re-connecting, and you're not even guaranteed to get the same connection on two subsequent executions.

However, you could use a transaction and `SET LOCAL`.

bobbytables:
@johto Yeah that's what I figured. I took a gander at the connection options for postgres and it doesn't even seem possible at connection time to set this parameter.

johto:
> I took a gander at the connection options for postgres and it doesn't even seem possible at connection time to set this parameter.

Sure it is.  But because database/sql only exposes a pool of connections, you have to have it for all connections in the pool.  Just set it in the connection string like I explained in the post you linked to.

jmscott:
with regard to timing out all queries in a connection pool, i am not
convinced adding the option "statement_timeout=<ms>"  works.   for example,
statement_timeout=1 is not timeing out my pooled queries.  my version of pq
was built yesterday, pulled from github via "go get ...".  the database is
pg9.6beta3.

-j

peter-mogensen:
[...]

PR #498, opened 2016-09-08
Add read_timeout and write_timeout

pneisen:
I believe this addresses issue #450.

While doing some failure testing connected to an AWS HA RDS cluster, we noticed that our connections would hang on a net.(*conn).Read forever. The Postgres server was no longer responding so, statement_timeout had no effect. The stack trace for the affected goroutines looked like this:
```
#    0x43d4e8    net.runtime_pollWait+0x58                                    /usr/local/go/src/runtime/netpoll.go:160
#    0x5654f7    net.(*pollDesc).wait+0x37                                    /usr/local/go/src/net/fd_poll_runtime.go:73
#    0x565563    net.(*pollDesc).waitRead+0x33                                    /usr/local/go/src/net/fd_poll_runtime.go:78
#    0x566c00    net.(*netFD).Read+0x1a0                                        /usr/local/go/src/net/fd_unix.go:243
#    0x571d8f    net.(*conn).Read+0x6f                                        /usr/local/go/src/net/net.go:173
#    0x608490    crypto/tls.(*block).readFromUntil+0x90                                /usr/local/go/src/crypto/tls/conn.go:472
#    0x6089d3    crypto/tls.(*Conn).readRecord+0xc3                                /usr/local/go/src/crypto/tls/conn.go:574
#    0x60c805    crypto/tls.(*Conn).Read+0x115                                    /usr/local/go/src/crypto/tls/conn.go:1109
#    0x637b2b    bufio.(*Reader).fill+0x10b                                    /usr/local/go/src/bufio/bufio.go:97
#    0x63818b    bufio.(*Reader).Read+0x1bb                                    /usr/local/go/src/bufio/bufio.go:209
#    0x5e5393    io.ReadAtLeast+0xa3                                        /usr/local/go/src/io/io.go:307
#    0x5e54f7    io.ReadFull+0x57                                        /usr/local/go/src/io/io.go:325
#    0x591b06    myproject/vendor/github.com/lib/pq.(*conn).recvMessage+0x116    /home/pneisen/myproject/vendor/github.com/lib/pq/conn.go:892
#    0x591e28    myproject/vendor/github.com/lib/pq.(*conn).recv1Buf+0x38        /home/pneisen/myproject/vendor/github.com/lib/pq/conn.go:942
#    0x591f36    myproject/vendor/github.com/lib/pq.(*conn).recv1+0x86        /home/pneisen/myproject/vendor/github.com/lib/pq/conn.go:963
#    0x59981e    myproject/vendor/github.com/lib/pq.(*conn).readParseResponse+0x2e    /home/pneisen/myproject/vendor/github.com/lib/pq/conn.go:1594
#    0x59056b    myproject/vendor/github.com/lib/pq.(*conn).prepareTo+0x63b        /home/pneisen/myproject/vendor/github.com/lib/pq/conn.go:732
#    0x5912b5    myproject/vendor/github.com/lib/pq.(*conn).Query+0x3d5        /home/pneisen/myproject/vendor/github.com/lib/pq/conn.go:790
#    0x5406e6    database/sql.(*DB).queryConn+0x5d6                                /usr/local/go/src/database/sql/sql.go:1092
#    0x54007f    database/sql.(*DB).query+0xef                                    /usr/local/go/src/database/sql/sql.go:1079
#    0x53fdcf    database/sql.(*DB).Query+0x8f                                    /usr/local/go/src/database/sql/sql.go:1062
#    0x5409bf    database/sql.(*DB).QueryRow+0x6f                                /usr/local/go/src/database/sql/sql.go:1143
```

Since the connect_timeout only applies to the initial connection before it is placed in the pool, and statement_timeout is a Postgres thing that doesn't apply when Postgres is down, we decided we wanted more timeout options. So, I added read_timeout and write_timeout (in milliseconds) to the connection string parameters and use them to set deadlines on reads and writes to the database connection. If a read or write returns before the deadline, then the deadline is cleared so the connection can be returned to the pool. If the deadline is reached an error is returned and eventually the connection gets marked as bad.

This fixed the issue we were seeing, but we did notice that the old, hung, established connections were still around in netstat. Looking at pq.(*conn).Close(), when the sql layer called to close the bad connection, the code would just return driver.ErrBadConn and not try to close the connection. I changed that to close the connection and return the result of the close call, similar to what is done at the bottom of that function. This seemed to do the trick and the hung connections were cleaned up.

Any feedback on these changes would be greatly appreciated. We have tested them in our fork as part of our applications failure testing and have had good results.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -115,6 +115,11 @@ type conn struct {
 	// Whether to always send []byte parameters over as binary.  Enables single
 	// round-trip mode for non-prepared Query calls.
 	binaryParameters bool
+
+	// Timeouts for read and write operations against the database server.
+	// A duration of 0 indicates no timeout.
+	readTimeout  time.Duration
+	writeTimeout time.Duration
 }
 
 // Handle driver-side settings in parsed connection string.
@@ -132,6 +137,13 @@ func (c *conn) handleDriverSettings(o values) (err error) {
 		return nil
 	}
 
+	intSetting := func(key string) (int, error) {
+		if value := o.Get(key); value != "" {
+			return strconv.Atoi(value)
+		}
+		return 0, nil
+	}
+
 	err = boolSetting("disable_prepared_binary_result", &c.disablePreparedBinaryResult)
 	if err != nil {
 		return err
@@ -140,6 +152,21 @@ func (c *conn) handleDriverSettings(o values) (err error) {
 	if err != nil {
 		return err
 	}
+
+	rt, err := intSetting("read_timeout")
+	if err != nil {
+		return err
+	}
+	// read_timeout is specified in milliseconds.
+	c.readTimeout = time.Duration(rt) * time.Millisecond
+
+	wt, err := intSetting("write_timeout")
+	if err != nil {
+		return err
+	}
+	// write_timeout is specified in milliseconds.
+	c.writeTimeout = time.Duration(wt) * time.Millisecond
+
 	return nil
 }
 
@@ -750,7 +777,7 @@ func (cn *conn) Prepare(q string) (_ driver.Stmt, err error) {
 
 func (cn *conn) Close() (err error) {
 	if cn.bad {
-		return driver.ErrBadConn
+		return cn.c.Close()
 	}
 	defer cn.errRecover(&err)
 
@@ -836,6 +863,18 @@ func (cn *conn) Exec(query string, args []driver.Value) (res driver.Result, err 
 }
 
 func (cn *conn) send(m *writeBuf) {
+	defer func() {
+		if cn.writeTimeout != 0 {
+			// Clear the write deadline if we set one.
+			cn.c.SetWriteDeadline(time.Time{})
+		}
+	}()
+
+	// Set the write deadline if we have a write timeout set.
+	if cn.writeTimeout != 0 {
+		cn.c.SetWriteDeadline(time.Now().Add(cn.writeTimeout))
+	}
+
 	_, err := cn.c.Write(m.wrap())
 	if err != nil {
 		panic(err)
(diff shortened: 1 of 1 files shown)


conn.go
```diff
@@ -132,6 +137,13 @@ func (c *conn) handleDriverSettings(o values) (err error) {
 		return nil
 	}
 
+	intSetting := func(key string) (int, error) {
```
johto: Why the different interface compared to boolSetting()?

conn.go
```diff
@@ -750,7 +777,7 @@ func (cn *conn) Prepare(q string) (_ driver.Stmt, err error) {
 
 func (cn *conn) Close() (err error) {
 	if cn.bad {
-		return driver.ErrBadConn
+		return cn.c.Close()
```
johto: This doesn't look correct, and should be done in a separate commit.

conn.go
```diff
@@ -836,13 +863,37 @@ func (cn *conn) Exec(query string, args []driver.Value) (res driver.Result, err
 }
 
 func (cn *conn) send(m *writeBuf) {
+	defer func() {
```
johto: defer allocates.  At the very least don't make everyone pay a price for this feature.  But ideally don't defer at all.

johto:
I think this feature is inherently misguided, but since I'm not going to be working on anything better and it can work for some types of applications, I'm not going to object.

There's two patches pretending to be one here, though, and I had some comments on both of them.

cbandy:
It looks to me like `libpq` punts on this by letting the user set things using `PQsocket()`.

Our equivalent is the `Dialer`. Should the user be doing this in the `Read` and `Write` of their own `net.Conn`?

pneisen:
That is an interesting thought about creating a custom ```Dialer```.  I think over all that would be a cleaner way to do it and it I could take care of both of the issues we were seeing.  I'll give it an attempt and get back with you guys. 

Thanks for the feedback!

conn.go
pneisen: I didn't realize that. Thanks for the info.

pneisen:
I went down the road of creating a wrapper around lib/pq that adds a custom dialer when Open() is called: https://github.com/Kount/pq-timeouts

The wrapper is cleaner and thanks for the suggestion. I'm going to close this pull request. I still think Close() not attempting to close a connection that is marked bad is an issue, but I will issue another pull request for that.

Thanks.

PR #500, opened 2016-09-12
Support empty Exec and Prepare

### LINKED ISSUE #499: panic on empty query + RowsAffected
An empty query panics when calling RowsAffected (and probably LastInsertId).

```
package main

import (
	"database/sql"
	"fmt"

	_ "github.com/lib/pq"
)

func main() {
	db, err := sql.Open("postgres", "postgresql://postgres@localhost:5432/?sslmode=disable")
	if err != nil {
		panic(err)
	}
	defer db.Close()

	if res, err := db.Exec(``); err != nil {
		panic(err)
	} else {
		fmt.Println("RES", res)
		fmt.Println(res.RowsAffected()) // <- PANIC HERE
	}
}
```

```
RES {0xc420012230 <nil>}
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x45e0bf]

goroutine 1 [running]:
panic(0x5cd080, 0xc42000c140)
	/usr/local/go/src/runtime/panic.go:500 +0x1a1
database/sql.driverResult.RowsAffected(0x6d2e40, 0xc420012230, 0x0, 0x0, 0x0, 0x0, 0x0)
	/usr/local/go/src/database/sql/sql.go:1963 +0x7f
database/sql.(*driverResult).RowsAffected(0xc420017140, 0x2, 0x2, 0x19)
	<autogenerated>:20 +0x6d
main.main()
	/home/mjibson/src/github.com/cockroachdb/t/m.go:21 +0x1f6
exit status 2
```

mjibson:
Fixes #499

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -32,6 +32,10 @@ var (
 	ErrSSLNotSupported           = errors.New("pq: SSL is not enabled on the server")
 	ErrSSLKeyHasWorldPermissions = errors.New("pq: Private key file has group or world access. Permissions should be u=rw (0600) or less.")
 	ErrCouldNotDetectUsername    = errors.New("pq: Could not detect default username. Please provide one explicitly.")
+
+	errUnexpectedReady = errors.New("unexpected ReadyForQuery")
+	errNoRowsAffected  = errors.New("no RowsAffected available after the empty statement")
+	errNoLastInsertId  = errors.New("no LastInsertId available after the empty statement")
 )
 
 type drv struct{}
@@ -598,11 +602,16 @@ func (cn *conn) simpleExec(q string) (res driver.Result, commandTag string, err 
 			res, commandTag = cn.parseComplete(r.string())
 		case 'Z':
 			cn.processReadyForQuery(r)
+			if res == nil && err == nil {
+				err = errUnexpectedReady
+			}
 			// done
 			return
 		case 'E':
 			err = parseError(r)
-		case 'T', 'D', 'I':
+		case 'I':
+			res = emptyRows
+		case 'T', 'D':
 			// ignore any results
 		default:
 			cn.bad = true
@@ -664,6 +673,20 @@ func (cn *conn) simpleQuery(q string) (res *rows, err error) {
 			errorf("unknown response for simple query: %q", t)
 		}
 	}
+}
+
+type noRows struct{}
+
+var emptyRows noRows
+
+var _ driver.Result = noRows{}
+
+func (noRows) LastInsertId() (int64, error) {
+	return 0, errNoLastInsertId
+}
+
+func (noRows) RowsAffected() (int64, error) {
+	return 0, errNoRowsAffected
 }
 
 // Decides which column formats to use for a prepared statement.  The input is
@@ -1720,6 +1743,9 @@ func (cn *conn) readExecuteResponse(protocolState string) (res driver.Result, co
 			res, commandTag = cn.parseComplete(r.string())
 		case 'Z':
 			cn.processReadyForQuery(r)
+			if res == nil && err == nil {
+				err = errUnexpectedReady
+			}
 			return res, commandTag, err
 		case 'E':
 			err = parseError(r)
@@ -1727,6 +1753,9 @@ func (cn *conn) readExecuteResponse(protocolState string) (res driver.Result, co
 			if err != nil {
 				cn.bad = true
 				errorf("unexpected %q after error %s", t, err)
+			}
+			if t == 'I' {
+				res = emptyRows
 			}
 			// ignore any results
 		default:
--- a/conn_test.go
+++ b/conn_test.go
@@ -385,9 +385,15 @@ func TestEmptyQuery(t *testing.T) {
 	db := openTestConn(t)
 	defer db.Close()
 
-	_, err := db.Exec("")
+	res, err := db.Exec("")
 	if err != nil {
 		t.Fatal(err)
+	}
+	if _, err := res.RowsAffected(); err != errNoRowsAffected {
+		t.Fatalf("expected %s, got %v", errNoRowsAffected, err)
+	}
+	if _, err := res.LastInsertId(); err != errNoLastInsertId {
+		t.Fatalf("expected %s, got %v", errNoLastInsertId, err)
 	}
 	rows, err := db.Query("")
 	if err != nil {
@@ -411,9 +417,15 @@ func TestEmptyQuery(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	_, err = stmt.Exec()
+	res, err = stmt.Exec()
 	if err != nil {
 		t.Fatal(err)
+	}
+	if _, err := res.RowsAffected(); err != errNoRowsAffected {
+		t.Fatalf("expected %s, got %v", errNoRowsAffected, err)
+	}
+	if _, err := res.LastInsertId(); err != errNoLastInsertId {
+		t.Fatalf("expected %s, got %v", errNoLastInsertId, err)
 	}
 	rows, err = stmt.Query()
 	if err != nil {


conn.go
```diff
@@ -591,6 +591,8 @@ func (cn *conn) simpleExec(q string) (res driver.Result, commandTag string, err
 	b.string(q)
 	cn.send(b)
 
+	res = emptyRows
```
johto: This isn't really correct.  This should only be assigned after EmptyQueryResponse ('I').

conn.go
```diff
@@ -666,6 +668,20 @@ func (cn *conn) simpleQuery(q string) (res *rows, err error) {
...
 }
 
+type noRows struct{}
+
+var emptyRows noRows
+
+var _ driver.Result = noRows{}
+
+func (noRows) LastInsertId() (int64, error) {
+	return 0, errors.New("no LastInsertId available after empty statement")
```
johto: "after *the* empty statement"

conn.go
```diff
@@ -666,6 +668,20 @@ func (cn *conn) simpleQuery(q string) (res *rows, err error) {
...
+}
+
+func (noRows) RowsAffected() (int64, error) {
+	return 0, errors.New("no RowsAffected available after empty statement")
```
johto: I feel like this should either not return an error, or the error should be exported.  Otherwise it's difficult to deal with these if you don't fully control the queries you send.

conn_test.go
```diff
@@ -385,10 +385,15 @@ func TestEmptyQuery(t *testing.T) {
...
 	defer db.Close()
 
-	_, err := db.Exec("")
+	res, err := db.Exec("")
 	if err != nil {
 		t.Fatal(err)
 	}
+	if ra, err := res.RowsAffected(); err == nil {
+		t.Fatal("expected error")
+	} else if ra != 0 {
```
johto: Feels odd to look at "ra" if err != nil; per convention in Go that usually means "ra" must be ignored.  But see my comment about RowsAffected above.

freeformz:
Is an empty Exec an error? If not we could possibly do this: https://github.com/lib/pq/compare/499

johto:
> Is an empty Exec an error? If not we could possibly do this: https://github.com/lib/pq/compare/499

I don't see the point when we could just handle EmptyQueryResponse correctly instead.

freeformz:
@johto I guess the Q is, what is "correct". Ignore each implementation (which don't really differ that much in the end). Should db.Exec() return something on which (0, "An Error of some sort") is returned?

freeformz:
FWIW: I don't see why an empty Exec's `LastInsertId` and `RowsAffected` should return errors. What am I missing?

conn.go
mjibson: While I agree from a spec-compliant protocol perspective, `res` will be `nil` if, for some reason, a `Z` is sent before an `I`, yielding the same panic. It is possible we should set `err` to something and then clear it when either `I` or `C` occur. This would send an error in any case where res wasn't set. Agree?

conn.go
mjibson: Exporting is probably best. Returing an error instead of just `0, nil` is consistent with https://github.com/golang/go/blob/8f9671d11a219d8fc9a6176ddf9939e743982ffc/src/database/sql/driver/driver.go#L202.

conn_test.go
mjibson: Good point.

mjibson:
@freeformz I think LastInsertId and RowsAffected should return errors for the same reasons that `driver.ResultNoRows` does: https://github.com/golang/go/blob/8f9671d11a219d8fc9a6176ddf9939e743982ffc/src/database/sql/driver/driver.go#L202. A `0, nil` return implies that 0 rows were affected, when really there wasn't even a possibility that a row could be modified.

I've modified the code to handle the `I` case correctly and also return an error if neither res or err are set.

conn.go
```diff
@@ -666,6 +671,22 @@ func (cn *conn) simpleQuery(q string) (res *rows, err error) {
 	}
 }
 
+type noRows struct{}
+
+var emptyRows noRows
+
+var _ driver.Result = noRows{}
+
+var ErrNoRowsAffected = errors.New("no RowsAffected available after the empty statement")
```
johto: The exported symbol seems awfully general compared to how specific the error message is.

mjibson: Have a better suggestion? I don't think it needs to be exported myself. Many other uncommon errors aren't.

mjibson:
Added code + test for Prepare as well, which was similarly broken.

conn.go
freeformz: Another way to do this is to define a public method that a user defined interface can be asserted against (for those users who really need to detect this error).

freeformz:
@mjibson Okay, sounds reasonable.

mjibson:
I've unexported the error. Good?

mjibson:
The test failures here seem to be unrelated to this change. Recommendations on how to proceed?

abonec:
I have this piece of code:
`result, err := wa.db.Exec("UPDATE snapshot_road SET length_measure = $1, area_measure = $2 WHERE actuality = 0 AND issue_task_id = $3;", rand.Intn(100), rand.Intn(100), issueTaskId)`
It works normaly when I not use `rowsAffected, err := result.RowsAffected()`. But when I do and when postgres throw an error `deadlock detected` program exit with panic and I can't handle this error even with this patch.

mjibson:
That test failure went away. Can I get another review or merge?

conn.go
```diff
@@ -598,11 +598,16 @@ func (cn *conn) simpleExec(q string) (res driver.Result, commandTag string, err
 			res, commandTag = cn.parseComplete(r.string())
 		case 'Z':
 			cn.processReadyForQuery(r)
+			if res == nil && err == nil {
+				err = errors.New("unexpected ReadyForQuery")
```
tamird: can you extract a singleton for this? it's repeated in `readExecuteResponse`

conn_test.go
```diff
@@ -385,10 +385,13 @@ func TestEmptyQuery(t *testing.T) {
 	db := openTestConn(t)
 	defer db.Close()
 
-	_, err := db.Exec("")
+	res, err := db.Exec("")
 	if err != nil {
 		t.Fatal(err)
 	}
+	if _, err := res.RowsAffected(); err != errNoRowsAffected {
+		t.Fatal("expected error")
```
tamird: `t.Fatalf("expected %s, got %v", errNoRowsAffected, err`

conn_test.go
```diff
@@ -411,10 +414,13 @@ func TestEmptyQuery(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	_, err = stmt.Exec()
+	res, err = stmt.Exec()
 	if err != nil {
 		t.Fatal(err)
 	}
+	if _, err := res.RowsAffected(); err != errNoRowsAffected {
+		t.Fatal("expected error")
```
tamird: same as above.

conn_test.go
```diff
@@ -385,10 +385,13 @@ func TestEmptyQuery(t *testing.T) {
 	db := openTestConn(t)
 	defer db.Close()
 
-	_, err := db.Exec("")
+	res, err := db.Exec("")
```
tamird: no test for LastInsertId?

conn_test.go
```diff
@@ -411,10 +414,13 @@ func TestEmptyQuery(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	_, err = stmt.Exec()
+	res, err = stmt.Exec()
```
tamird: no test for LastInsertId?

mjibson:
Added error singletons and lastinsertid tests.

tamird:
LGTM @johto could you take a look?

tlvenn:
Hi @johto, any chance to get this PR merged soon ? Thanks!

conn.go
```diff
@@ -32,6 +32,10 @@ var (
 	ErrSSLNotSupported           = errors.New("pq: SSL is not enabled on the server")
 	ErrSSLKeyHasWorldPermissions = errors.New("pq: Private key file has group or world access. Permissions should be u=rw (0600) or less.")
 	ErrCouldNotDetectUsername    = errors.New("pq: Could not detect default username. Please provide one explicitly.")
+
+	errUnexpectedReady = errors.New("unexpected ReadyForQuery")
+	errNoRowsAffected  = errors.New("no RowsAffected available after the empty statement")
+	errNoLastInsertId  = errors.New("no LastInsertId available after the empty statement")
```
freeformz: How do you expect these errors to be used by users of database/sql?

mjibson: I don't expect them to be used in that way. The errors are named here for the tests. I do not think they need to be exported because these errors are very rare (as evidenced by the fact that it was previously a panic if they occurred and that went undetected until now). It is incredibly unlikely that another library cares if an empty string is the cause of an error instead of just caring that there was an error at all.

However if you disagree I'm happy to export them. But I think that adds some unused clutter to the exported API that doesn't need to be present.

freeformz:
It's less that I think they should be exported and more about how users
will need to interact with these (even if they are rare) errors. Having to
do string matches on errors is :-(. Are they Temporary() errors? Can there
be other behavior associated with them that would make them easy to
detect/classify?

tamird:
This LGTM as-is. If users find that they need to interact with these errors, we can always export them, but I would rather that we not do so here in the name of reducing API obligations.

PR #501, opened 2016-09-13
Changes to conn.Close()

pneisen:
While doing load and failover testing, we noticed that connections marked as bad (timeouts, disconnections, errors, etc) were never being closed and remained marked as ESTABLISHED in a netstat listing. 

Looking at the code, Close() is never called on the connection in the case of the connection being marked bad. Looking around at the code for other golang database/sql drivers, they seem to all call Close() on the connection regardless of the state of the connection. I think this is what lib/pq should be doing as well. Thoughts?

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -749,19 +749,24 @@ func (cn *conn) Prepare(q string) (_ driver.Stmt, err error) {
 }
 
 func (cn *conn) Close() (err error) {
-	if cn.bad {
-		return driver.ErrBadConn
-	}
 	defer cn.errRecover(&err)
 
 	// Don't go through send(); ListenerConn relies on us not scribbling on the
 	// scratch buffer of this connection.
-	err = cn.sendSimpleMessage('X')
-	if err != nil {
-		return err
+	if cn.c != nil {
+		err := cn.sendSimpleMessage('X')
+		if err != nil {
+			return err
+		}
+
+		err = cn.c.Close()
+		cn.c = nil
+		if err != nil {
+			return err
+		}
 	}
 
-	return cn.c.Close()
+	return nil
 }
 
 // Implement the "Queryer" interface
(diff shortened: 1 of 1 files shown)


conn.go
```diff
@@ -749,19 +749,24 @@ func (cn *conn) Prepare(q string) (_ driver.Stmt, err error) {
...
-		return driver.ErrBadConn
-	}
 	defer cn.errRecover(&err)
 
-	// Don't go through send(); ListenerConn relies on us not scribbling on the
-	// scratch buffer of this connection.
-	err = cn.sendSimpleMessage('X')
-	if err != nil {
-		return err
+	if cn.c != nil {
```
mjibson: Why all this business about setting `cn.c = nil`? This function is only every called once per connection, so we will never hit this if block.

conn.go
```diff
@@ -749,19 +749,24 @@ func (cn *conn) Prepare(q string) (_ driver.Stmt, err error) {
...
+		// Don't go through send(); ListenerConn relies on us not scribbling on the
+		// scratch buffer of this connection.
+		err := cn.sendSimpleMessage('X')
+		if err != nil {
+			return err
```
mjibson: Returning here will result in the same problem as this PR is trying to fix: the connection will remain open.

mjibson:
My mistake, guru led me astray: conn.Close is called externally. I've deleted the comment about that.

conn.go
```diff
@@ -749,19 +749,24 @@ func (cn *conn) Prepare(q string) (_ driver.Stmt, err error) {
...
+
+		err = cn.c.Close()
+		cn.c = nil
+		if err != nil {
+			return err
+		}
 	}
 
-	return cn.c.Close()
+	return nil
```
mjibson: Should this return an error if `cn.c` was already nil? Seems like in that case we'd want `errors.New("pq: cannot close already closed connection")`?

mjibson:
Reading through Go's net conn close code, I think this should just always call cn.c.Close() because if that is closed twice, it'll return an error correctly:

https://github.com/golang/go/blob/81b9af7cccce8234319551330cb6406469f32bab/src/net/net.go#L198

PR #507, opened 2016-09-26
Add null bytea support

kuznet1:


cbandy:
After a quick skim of `database/sql`, I can't tell: Is it the responsibility of `Scanner.Scan` to `copy` the slice or does something else do that already?

encode.go
```diff
@@ -587,3 +587,22 @@ func (nt NullTime) Value() (driver.Value, error) {
...
+	return nil
+}
+
+// Value implements the driver Valuer interface.
+func (nb NullBytea) Value() (driver.Value, error) {
+	if !nb.Valid {
+		return nil, nil
+	}
+	return nb.Bytea, nil
+}
```
cbandy: Needs a newline at end of file.

kuznet1:
You're right, slice did not copied by anybody else. There are two possible types: `[]byte` and `RawBytes`. First is copied, but second no. I think that copied version is not necessary, but need to say about this in doc like [this](https://golang.org/pkg/database/sql/#RawBytes).

encode_test.go
```diff
@@ -31,6 +31,29 @@ func TestScanNilTimestamp(t *testing.T) {
 	}
 }
 
+func TestScanBytea(t *testing.T) {
+	var nb NullBytea
+	bytes := []byte("byteatest")
+	nb.Scan(bytes)
+	if !nb.Valid {
+		t.Errorf("Expected Valid=false")
```
cbandy: The expectation is that `Valid` is `true`.

encode_test.go
```diff
@@ -31,6 +31,29 @@ func TestScanNilTimestamp(t *testing.T) {
...
+	}
+	for i, elem := range nb.Bytea {
+		if bytes[i] != elem {
```
cbandy: Usually, we compare slices using `reflect.DeepEqual`:

```go
if !reflect.DeepEqual(bytes, nb.Bytea) {
  t.Errorf("Expected %v, got %v", bytes, nb.Bytea)
}
```

encode_test.go
```diff
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/lib/pq/oid"
+	"reflect"
```
cbandy: Build is failing on formatting rules. Move this to the group of imports around L6.

arp242:
The Go standard library now has `sql.Null[T any]`, which can be used for this. Seems to work well in a quick test:

```
func TestNullBytea( Added in go1.22
    {
        var nb sql.Null[[]byte]
        bytes := []byte("byteatest")
        nb.Scan(bytes)
        if !nb.Valid {
            t.Fatal("want Valid=true")
        }
        if !reflect.DeepEqual(bytes, nb.V) {
            t.Fatalf("\nwant: %#v\nhave: %#v", bytes, nb.V)
        }
    }

    {
        var nb sql.Null[[]byte]
        nb.Scan(nil)
        if nb.Valid {
            t.Error("want Valid=false")
        }
    }
}
```

PR #508, opened 2016-09-28
TLS Renegotiation Support

dlsteuer:
Hi there, we are using lib/pq to connect to an amazon redshift databases, we've had to disable sslmode because otherwise the connection drops silently when the query returns large amounts of data. Support for this was added in Go 1.7, and here is how we've added the support which works in our use case. I'm completely open to any suggestions on how to better set the value on the tlsConfig struct as I'm not entirely happy with the current implementation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 *.test
 *~
 *.swp
+*.iml
--- a/bench_test.go
+++ b/bench_test.go
@@ -17,7 +17,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/lib/pq/oid"
+	"github.com/sendwithus/pq/oid"
 )
 
 var (
--- a/buf.go
+++ b/buf.go
@@ -4,7 +4,7 @@ import (
 	"bytes"
 	"encoding/binary"
 
-	"github.com/lib/pq/oid"
+	"github.com/sendwithus/pq/oid"
 )
 
 type readBuf []byte
--- a/conn.go
+++ b/conn.go
@@ -22,7 +22,7 @@ import (
 	"time"
 	"unicode"
 
-	"github.com/lib/pq/oid"
+	"github.com/sendwithus/pq/oid"
 )
 
 // Common error types
--- a/conn.go
+++ b/conn.go
@@ -998,6 +998,16 @@ func (cn *conn) ssl(o values) {
 		errorf(`unsupported sslmode %q; only "require" (default), "verify-full", "verify-ca", and "disable" supported`, mode)
 	}
 
+	switch o.Get("tls-renegotiation") {
+	case "RenegotiateNever":
+		tlsConf.Renegotiation = tls.RenegotiateNever
+	case "RenegotiateFreelyAsClient":
+		tlsConf.Renegotiation = tls.RenegotiateFreelyAsClient
+	case "RenegotiateOnceAsClient":
+		tlsConf.Renegotiation = tls.RenegotiateOnceAsClient
+	}
+	delete(o, "tls-renegotiation")
+
 	cn.setupSSLClientCertificates(&tlsConf, o)
 	cn.setupSSLCA(&tlsConf, o)
 
--- a/encode.go
+++ b/encode.go
@@ -13,7 +13,7 @@ import (
 	"sync"
 	"time"
 
-	"github.com/lib/pq/oid"
+	"github.com/sendwithus/pq/oid"
 )
 
 func binaryEncode(parameterStatus *parameterStatus, x interface{}) []byte {
--- a/encode_test.go
+++ b/encode_test.go
@@ -8,7 +8,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/lib/pq/oid"
+	"github.com/sendwithus/pq/oid"
 )
 
 func TestScanTimestamp(t *testing.T) {


cbandy:
Interesting. [PostgreSQL removed SSL renegotiation in 9.5](https://www.postgresql.org/message-id/E1ZKB9q-0004zo-12@gemulon.postgresql.org). What value do you use successfully with Redshift?

dlsteuer:
That certainly makes things interesting, I was able to use RenegotiateFreelyAsClient successfully with Redshift where prior to that, the connection would silently die, and I would only get the rows that had been retrieved up to that point.

johto:
Is there not a way to disable renegotiation in Redshit?  I don't really object to this change (though I do have some comments based on a quick review), but my understanding was that the industry is currently thinking that renegotiation is a bad idea in general.

dlsteuer:
I'm unable to really find much information on renegotiation in redshift, pretty much just https://forums.aws.amazon.com/thread.jspa?messageID=722247&#722247. We'd prefer to disable renegotiation if we could.

cbandy:
Here's what I found.

1. It's easy to reproduce:

  ```go
package main

import (
	"database/sql"
	"log"

	_ "github.com/lib/pq"
)

func main() {
	db, err := sql.Open("postgres", "host=yours.here.redshift.amazonaws.com port=5439 user=abc password=xyz sslmode=require database=dev")
	if err != nil {
		log.Fatalf("connect %v", err)
	}
	defer db.Close()

	rows, err := db.Query(`SELECT rpad('',1024*1024,'x') FROM generate_series(1,1024)`)
	if err != nil {
		log.Fatalf("query %v", err)
	}

	var x string
	var i int
	for rows.Next() {
		i++
		err = rows.Scan(&x)
		if err != nil {
			log.Fatalf("scan %v", err)
		}
	}
	log.Printf("%d rows", i)

	err = rows.Err()
	if err != nil {
		log.Fatalf("rows %v", err)
	}
}
```
  ```
% go run main.go
2016/09/29 15:55:27 511 rows
2016/09/29 15:55:27 rows driver: bad connection
exit status 1
```

2. I wouldn't say that it fails "silently." In the above test, `rows.Err()` does return an `ErrBadConn`.

3. It seems to happen around 512 MiB, which coincides with the default [`ssl_renegotiation_limit`](https://www.postgresql.org/docs/8.0/static/runtime-config.html#GUC-SSL-RENEGOTIATION-LIMIT) of PostgreSQL 8.0.

4. Redshift doesn't have a user-facing way to see or [change](http://docs.aws.amazon.com/redshift/latest/dg/cm_chap_ConfigurationRef.html) [something like] the `ssl_renegotiation_limit`.

5. I wasn't able to reproduce this with the Redshift JDBC driver nor `psql`:
  ```
% psql -XAt -h yours.here.redshift.amazonaws.com -p 5439 -d dev -U abc
psql (9.5.3, server 8.0.2)
SSL connection (protocol: TLSv1.2, cipher: ECDHE-RSA-AES256-GCM-SHA384, bits: 256, compression: on)
Type "help" for help.

dev=# \q

% psql -XAt -h yours.here.redshift.amazonaws.com -p 5439 -d dev -U abc \
  -c "SELECT rpad('',1024*1024,'x') FROM generate_series(1,1024)" | wc -l
    1024
```
  ```
% PGSSLCOMPRESSION=0 psql -XAt -h yours.here.redshift.amazonaws.com -p 5439 -d dev -U abc
psql (9.5.3, server 8.0.2)
SSL connection (protocol: TLSv1.2, cipher: ECDHE-RSA-AES256-GCM-SHA384, bits: 256, compression: off)
Type "help" for help.

dev=# \q

% PGSSLCOMPRESSION=0 psql -XAt -h yours.here.redshift.amazonaws.com -p 5439 -d dev -U abc \
  -c "SELECT rpad('',1024*1024,'x') FROM generate_series(1,1024)" | wc -l
    1024
```

6. From what I can tell, `libpq` uses the default OpenSSL settings for renegotiation (i.e. no related options are [set](https://github.com/postgres/postgres/blob/REL9_6_0/src/interfaces/libpq/fe-secure-openssl.c#L893).) If I understand the OpenSSL [docs](https://www.openssl.org/docs/manmaster/ssl/SSL_CTX_set_options.html#SECURE-RENEGOTIATION) correctly, the default is to allow renegotiations at any time.

7. Renegotiation was [added](https://github.com/golang/go/commit/af125a5193c75dd59307fcf1b26d885010ce8bfd) to `crypto/tls` in Go 1.7.

8. `crypto/tls` does no compression.
 - https://github.com/golang/go/blob/go1.7.1/src/crypto/tls/handshake_client.go#L60
 - https://github.com/golang/go/blob/go1.7.1/src/crypto/tls/handshake_client.go#L562-L565
 - https://github.com/golang/go/blob/go1.7.1/src/crypto/tls/common.go#L66-L69

freeformz:
@cbandy Does psql from 9.5 also disable renegotiation?

Maybe we include this with a default to `tls.RenegotiateNever` for go 1.7+

cbandy:
@freeformz From what I can tell, `psql` uses `libpq` directly which renegotiates at any time (see 6. above.) To behave similarly to `libpq`, we'd need to use `tls.RenegotiateFreelyAsClient`.

freeformz:
@cbandy doesn't that conflict with https://www.postgresql.org/message-id/E1ZKB9q-0004zo-12@gemulon.postgresql.org or am I missing something?

cbandy:
> am I missing something?

Maybe that there are two parts to the renegotiation. One side initiates it and the other side accepts it. I'll try to be more clear about the sides and settings (as I understand it so far):

From what I can tell, the OpenSSL call [`SSL_renegotiate`](https://github.com/openssl/openssl/blob/OpenSSL_1_1_0b/apps/s_client.c#L2418-L2422) initiates it. This function is not called by recent versions (since that commit) of PostgreSQL backend. This function is not called by any versions of `libpq`.

From what I can tell, no OpenSSL calls are needed to accept a renegotiation; the library does it for the user. There are no settings to disable this, but there are some [settings](https://www.openssl.org/docs/manmaster/ssl/SSL_CTX_set_options.html#SECURE-RENEGOTIATION) to ensure that OpenSSL follows [RFC 5746](https://tools.ietf.org/html/rfc5746) while doing it.

No versions of `crypto/tls` can initiate a renegotiation. Recent versions of `crypto/tls` (since 1.7) will accept one or more renegotiations [when configured](https://golang.org/pkg/crypto/tls/#RenegotiationSupport) to do so. The default configuration behaves the same as older versions: `Read` permanently returns [an error](https://github.com/golang/go/blob/go1.6.3/src/crypto/tls/conn.go#L676) after a renegotiation request.

freeformz:
Could this also be related to: https://go-review.googlesource.com/c/31268/

cbandy:
@dlsteuer have you had a chance to try last month's merge of #525?

dlsteuer:
Yup looks good to me, we've switch back to it. Thanks so much!

PR #517, opened 2016-10-12
Always close underlying connection

mjibson:
Closes #501

See that issue for original bug report and code.

---
This change is [Reviewable](https://reviewable.io/reviews/lib/pq/517)

### DIFF
--- a/conn_test.go
+++ b/conn_test.go
@@ -5,6 +5,7 @@ import (
 	"database/sql/driver"
 	"fmt"
 	"io"
+	"net"
 	"os"
 	"reflect"
 	"strings"
@@ -662,6 +663,40 @@ func TestBadConn(t *testing.T) {
 	}
 	if !cn.bad {
 		t.Fatalf("expected cn.bad")
+	}
+}
+
+// TestCloseBadConn tests that the underlying connection can be closed with
+// Close after an error.
+func TestCloseBadConn(t *testing.T) {
+	nc, err := net.Dial("tcp", "localhost:5432")
+	if err != nil {
+		t.Fatal(err)
+	}
+	cn := conn{c: nc}
+	func() {
+		defer cn.errRecover(&err)
+		panic(io.EOF)
+	}()
+	// Verify we can write before closing.
+	if _, err := nc.Write(nil); err != nil {
+		t.Fatal(err)
+	}
+	// First close should close the connection.
+	if err := cn.Close(); err != nil {
+		t.Fatal(err)
+	}
+	// Verify write after closing fails.
+	if _, err := nc.Write(nil); err == nil {
+		t.Fatal("expected error")
+	} else if !strings.Contains(err.Error(), "use of closed network connection") {
+		t.Fatalf("expected use of closed network connection error, got %s", err)
+	}
+	// Verify second close fails.
+	if err := cn.Close(); err == nil {
+		t.Fatal("expected error")
+	} else if !strings.Contains(err.Error(), "use of closed network connection") {
+		t.Fatalf("expected use of closed network connection error, got %s", err)
 	}
 }
 
--- a/conn.go
+++ b/conn.go
@@ -772,19 +772,21 @@ func (cn *conn) Prepare(q string) (_ driver.Stmt, err error) {
 }
 
 func (cn *conn) Close() (err error) {
-	if cn.bad {
-		return driver.ErrBadConn
-	}
+	// Skip cn.bad return here because we always want to close a connection.
 	defer cn.errRecover(&err)
+
+	// Ensure that cn.c.Close is always run. Since error handling is done with
+	// panics and cn.errRecover, the Close must be in a defer.
+	defer func() {
+		cerr := cn.c.Close()
+		if err == nil {
+			err = cerr
+		}
+	}()
 
 	// Don't go through send(); ListenerConn relies on us not scribbling on the
 	// scratch buffer of this connection.
-	err = cn.sendSimpleMessage('X')
-	if err != nil {
-		return err
-	}
-
-	return cn.c.Close()
+	return cn.sendSimpleMessage('X')
 }
 
 // Implement the "Queryer" interface


tamird:
:lgtm:

Rather than referencing the original PR, consider inlining the rationale into this new commit message to help future readers.

---

Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[conn.go, line 789 at r1](https://reviewable.io:443/reviews/lib/pq/517#-KTscfB6f2ieJrC4tuah:-KTscfB6f2ieJrC4tuai:bj97d50) ([raw file](https://github.com/lib/pq/blob/01f475f5e545a1c74299061b4d78e9be6ef30d43/conn.go#L789)):*
> ```Go
> 	// Don't go through send(); ListenerConn relies on us not scribbling on the
> 	// scratch buffer of this connection.
> 	err = cn.sendSimpleMessage('X')
> ```

`return cn.sendSimpleMessage('X')`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/lib/pq/517#-:-KTscox1g8VuzG1YSkSC:b-n75a5i)*

pneisen:
I submitted the original PR and I agree with this approach. Simple and it should take care of the issue we were seeing. Thanks!

mjibson:
@cbandy I've added tests to verify that the `net.Conn` is writeable and then not writeable after the close. I think this is simpler than the spy solution. Agree?

freeformz:
This does change behavior in that if the connection was detected as being bad users got a `driver.ErrBadConn` error.

mjibson:
Yes, that's true. Arguably that was incorrect behavior, though, since users expect Close to close and return an error related to the closing of the underlying connection instead of a generic "this is bad" error.

freeformz:
@mjibson According to the commit that added it(07745050691bbe88ce770ca38311c19be9445e37), this was to signal database/sql that the conn is bad. `driver.ErrBadConn` is checked in a lot of places inside of `database/sql`. With that said I didn't see anywhere (but I wasn't thorough) its being checked on the return value of a driver's Close().

mjibson:
Here's my research:

The only time database/sql/* checks the result of conn.Close is in the `finalClose` function (https://github.com/golang/go/blob/55ef67f2f85c51d415a030ae144a0b3301a097bd/src/database/sql/sql.go#L361). This method is only used where it is returned by removeDepLocked (https://github.com/golang/go/blob/55ef67f2f85c51d415a030ae144a0b3301a097bd/src/database/sql/sql.go#L450). This is called in 3 places:

1. closeDBLocked, which is invoked and the error accumulated (https://github.com/golang/go/blob/55ef67f2f85c51d415a030ae144a0b3301a097bd/src/database/sql/sql.go#L535) but returned to the caller uninspected. This PR would change the behavior here for bad connections to return a nil error on Close instead of the badConn error. I think that's a good change because users want to see the error of the close itself.
2. in driverConn.Close (https://github.com/golang/go/blob/55ef67f2f85c51d415a030ae144a0b3301a097bd/src/database/sql/sql.go#L350), but driverConn.Close's error is never checked (according to `guru referrers`)
3. in DB.removeDep, which is invoked in https://github.com/golang/go/blob/55ef67f2f85c51d415a030ae144a0b3301a097bd/src/database/sql/sql.go#L1872 but returned to the user since it's a Stmt.

Thus I have not been able to find any evidence that the database/sql package depends on Close returning ErrBadConn. I think we're safe. Agree?

freeformz:
@mjibson Thanks for doing more thorough research.

conn_test.go
```diff
@@ -665,6 +666,40 @@ func TestBadConn(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	// Verify write after closing fails.
+	if _, err := nc.Write(nil); err == nil {
+		t.Fatal("expected error")
+	} else if !strings.Contains(err.Error(), "use of closed network connection") {
+		t.Fatalf("expected use of closed network connection error, got %s", err)
+	}
+	// Verify second close fails.
+	if err := cn.Close(); err == nil {
```
freeformz: Why should the second Close() error ?

mjibson: Same behavior as os.File and net.Conn: closing something that's already been closed is an error.

tamird:
Reviewed 2 of 2 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[conn_test.go, line 672 at r2](https://reviewable.io:443/reviews/lib/pq/517#-KTujqPjOpxzxbkMDGfp:-KTujqPjOpxzxbkMDGfq:bbo6w0v) ([raw file](https://github.com/lib/pq/blob/4741650f3c4ff8849f62ccd81879b4cc28d8b969/conn_test.go#L672)):*
> ```Go
> // Close after an error.
> func TestCloseBadConn(t *testing.T) {
> 	nc, err := net.Dial("tcp", "localhost:5432")
> ```

consider using `net.Pipe` here. you could then try to read from the so-called "other" connection after calling Close.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/lib/pq/517)*

mjibson:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[conn_test.go, line 672 at r2](https://reviewable.io:443/reviews/lib/pq/517#-KTujqPjOpxzxbkMDGfp:-KTwmrCrU78_WGqsgGl8:b-2ivhym) ([raw file](https://github.com/lib/pq/blob/4741650f3c4ff8849f62ccd81879b4cc28d8b969/conn_test.go#L672)):*
<details><summary><i>Previously, tamird (Tamir Duberstein) wrote…</i></summary>
> consider using `net.Pipe` here. you could then try to read from the so-called "other" connection after calling Close.
</details>
I tried that and I'm not sure it's better. You have to start two io.Copy goroutines or the Write and Close calls fail (because the Close calls does a send). Without those I suppose it'd take a few go routines and channels to coordinate the sending. And I'm not sure what you mean by reading from the other connection after close. Why does it need to read? What is that testing?

Unless I'm missing something I think it's better as is.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/lib/pq/517)*

conn_test.go
freeformz: Fair enough.

PR #521, opened 2016-10-16
Show position in syntax errors

### LINKED ISSUE #194: Report position of syntax errors in queries
Currently the message when there is a syntax error does not include the position within the query. This makes larger queries quite difficult to debug. It looks like the Line and Where attributes of pq.Error don't return anything useful in these cases.

@johto tracked down the code that handles this in libpq to http://doxygen.postgresql.org/fe-protocol3_8c.html#a547f08283e1202c05d3039579dc38616

fxkr:
We now format them like psql, e.g.:

    pq: syntax error at or near "FROM"
    LINE 1: SELECT * FROM FROM items;
                          ^

Closes #194

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -621,7 +621,7 @@ func (cn *conn) simpleExec(q string) (res driver.Result, commandTag string, err 
 }
 
 func (cn *conn) simpleQuery(q string) (res *rows, err error) {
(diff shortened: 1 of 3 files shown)


arp242:
Fixed via #1219

Issue #489, opened 2016-08-18
Permission check on postgresql.key doesn't work on Windows


Superdominik:
I've run into a problem with client-side certificates on Windows. Even after changing the permissions on the key-file, I get 

> pq: Private key file has group or world access. Permissions should be u=rw (0600) or less.

I've investigated the problem a bit and found that [here](https://github.com/lib/pq/blob/55196ec83d60d789e27b2627549bf00974a70084/conn.go#L1099) in the code it checks for the `Mode()`of the file via the `os` package. This seems to return `-rw-rw-rw-` on Windows though, no matter how I change the permissions.

Is this a `pq` problem or maybe an upstream problem of the `os` package?

cbandy:
It looks like the `os` package is doing it: https://github.com/golang/go/blob/go1.7/src/os/types_windows.go#L31-L50

We should probably look to how `libpq` behaves on Windows.

cbandy:
It looks to me like `libpq` completely skips this check on Windows: https://github.com/postgres/postgres/blob/REL9_5_4/src/interfaces/libpq/fe-secure-openssl.c#L1192-L1200

freeformz:
It looks like it may work properly if the file is set to "Read Only". @Superdominik you said you changed the permissions, but I'm not sure if that's what you tried.

Superdominik:
@freeformz I've tried setting it to read-only. Go then thinks the file has `-r--r--r--` permissions. Which still doesn't allow it to be used with pq.
I've additionally removed inheritance and all principals from the file. Then added just the local user back and only gave read access. Changing anything in the Windows security model doesn't have any effect on the permissions read by `Mode()` though. So I don't think it's possible to work around this problem.

freeformz:
Ahhh Yeah, `0444` is > `0600`. :-( We should maybe skip the check on windows too then.

PR #525, opened 2016-10-23
TLS fixes

cbandy:
Moves the `crypto/tls` stuff out of `conn` and adds some files with build flags for #489 and #508.

I don't imagine users will ever want to disable TLS renegotiation, so I opted not to add a driver setting for it. If we do want a setting, then `sslRenegotiation` could change to look much like `sslClientCertificates` and `sslCertificateAuthority`.

### DIFF
--- a/ssl.go
+++ b/ssl.go
@@ -64,50 +64,37 @@ func ssl(o values) func(net.Conn) net.Conn {
 // in the user's home directory. The configured files must exist and have
 // the correct permissions.
 func sslClientCertificates(tlsConf *tls.Config, o values) {
-	var missingOk bool
-
 	sslkey := o.Get("sslkey")
 	sslcert := o.Get("sslcert")
-	if sslkey != "" && sslcert != "" {
-		// If the user has set an sslkey and sslcert, they *must* exist.
-		missingOk = false
-	} else {
-		// Automatically load certificates from ~/.postgresql.
-		user, err := user.Current()
+
+	var cinfo, kinfo os.FileInfo
+	var err error
+
+	if sslcert != "" && sslkey != "" {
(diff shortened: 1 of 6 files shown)


ssl.go
```diff
@@ -133,6 +120,34 @@ func sslCertificateAuthority(tlsConf *tls.Config, o values) {
...
+		// from where to load them.
+		return
+	}
+
+	cert = filepath.Join(user.HomeDir, ".postgresql", "postgresql.crt")
+	key = filepath.Join(user.HomeDir, ".postgresql", "postgresql.key")
+
+	cinfo, err = os.Stat(cert)
+	if err != nil {
+		cinfo = nil
```
mjibson: I don't think this if block or the one below are needed since if `err != nil` then `cinfo` is also set to nil.

cbandy: I don't see that documented or guaranteed for `os.Stat()`. The convention, as I understand it, is to not trust the other values returned when there is an error.

mjibson: Good point.

freeformz:
Why do all of the ssl paths panic, when a driver's Open can return an error ? This seems way saner to me and something we should fix. Since this touches all of `*conn.ssl` it seems like a decent time to do it. Thoughts?

cbandy:
> Why do all of the ssl paths panic, when a driver's Open can return an error ?

I'm not sure. `errRecover()` seems to have some logic related to TLS, e.g. `handshake failure`.

I considered returning errors but got lazy. I didn't try to grok the existing tests.

In fact, it occurs to me now that since I don't have a windows box on which to test this, I'm not _certain_ a windows build even compiles. How do we vet windows patches?

mjibson:
This successfully compiles on windows. I'm going to merge it.

Issue #494, opened 2016-08-30
race in COPY


mjibson:
The race detector complains about the following code:

```
package main

import (
    "database/sql"
    "log"

    "github.com/lib/pq"
)

func main() {
    db, err := sql.Open("postgres", "postgresql://postgres@localhost:5432?sslmode=disable")
    if err != nil {
        log.Fatal(err)
    }
    defer db.Close()

    if _, err := db.Exec(`
        CREATE TABLE t (
            i INT PRIMARY KEY
        );
    `); err != nil {
        log.Fatal(err)
    }

    txn, err := db.Begin()
    if err != nil {
        log.Fatal(err)
    }

    if _, err := txn.Prepare(pq.CopyIn("t", "i")); err != nil {
        log.Fatal(err)
    }

    if _, err := txn.Query("SELECT 1"); err == nil {
        log.Fatal("expected error")
    }
}
```

```
==================
WARNING: DATA RACE
Read at 0x00c420076208 by goroutine 8:
  bufio.(*Reader).Read()
      /usr/local/go/src/bufio/bufio.go:192 +0x10e
  io.ReadAtLeast()
      /usr/local/go/src/io/io.go:307 +0xb1
  io.ReadFull()
      /usr/local/go/src/io/io.go:325 +0x72
  github.com/lib/pq.(*conn).recvMessage()
      /go/src/github.com/lib/pq/conn.go:892 +0xdd
  github.com/lib/pq.(*copyin).resploop()
      /go/src/github.com/lib/pq/copy.go:143 +0x98

Previous write at 0x00c420076208 by main goroutine:
  bufio.(*Reader).fill()
      /usr/local/go/src/bufio/bufio.go:88 +0x553
  bufio.(*Reader).Read()
      /usr/local/go/src/bufio/bufio.go:209 +0x6ab
  io.ReadAtLeast()
      /usr/local/go/src/io/io.go:307 +0xb1
  io.ReadFull()
      /usr/local/go/src/io/io.go:325 +0x72
  github.com/lib/pq.(*conn).recvMessage()
      /go/src/github.com/lib/pq/conn.go:892 +0xdd
  github.com/lib/pq.(*conn).recv1Buf()
      /go/src/github.com/lib/pq/conn.go:942 +0x46
  github.com/lib/pq.(*conn).recv1()
      /go/src/github.com/lib/pq/conn.go:963 +0xaf
  github.com/lib/pq.(*conn).simpleQuery()
[...]

PR #527, opened 2016-10-27
Issue494

mjibson:


mjibson:
I ran this with the original tests in cockroach that found the bug and they now pass without complaining about a data race. I think it's time to clean this up and get it merged.

freeformz:
Fixes #494

freeformz:
Done.

tamird:
We should add `go test -race` to travis while you're fixing this.

mjibson:
```
--- FAIL: TestCopyRespLoopConnectionError (0.02s)
	copy_test.go:410: expected *pq.Error or driver.ErrBadConn, got &net.OpError{Op:"write", Net:"tcp", Source:(*net.TCPAddr)(0xc4201a7e60), Addr:(*net.TCPAddr)(0xc4201a7e90), Err:(*os.SyscallError)(0xc42021f740)}
```

One of the test failures occurred on go Tip, postgres 9.5. It appears to not always happen, so we may have some flaky tests.

freeformz:
Adding `go test -race` should be done separately IMO.

Issue #506, opened 2016-09-21
Add Context Support


freeformz:
This is to track support for this https://go-review.googlesource.com/c/29381/

freeformz:
golang/go#15123 is closed and this has landed in master as golang/go@e13df02e5fbf2c0cd8811b826a8c8567efa882dd

freeformz:
See also: https://go-review.googlesource.com/#/c/30166/

@johto @uhoh-itsmaciek You may want to take a look at that CL if interested.

itsjamie:
#531 is related

mjibson:
I'm actively working on this but don't have a full PR ready yet. If someone else has interest I'm happy to coordinate.

itsjamie:
@mjibson I was planning on working on one of the tickets I've linked to through #531, it looks like a lot of it runs through the new context-based API. I'd be happy to work with you if you've got something you can split off for me to work on.

mjibson:
I've just got this working so I'll send a PR soon. Need to write some tests first.

nathany:
With #535 merged, is this issue closed?

tamird:
Yep!

Issue #531, opened 2016-10-31
go1.8: support for additional features (by 2016-12-12)


kardianos:
Many new important features will land in go1.8. Please see document:
https://docs.google.com/document/d/1F778e7ZSNiSmbju3jsEWzShcb8lIO4kDyfKDNm4PNd8/edit?usp=sharing for the full list.

Because these features are new we want drivers to implement them before they are released. Ideally they should be implemented by 2016-12-12.

PR #535, opened 2016-11-07
Add context methods

mjibson:
Progress on #506.

---
This change is [Reviewable](https://reviewable.io/reviews/lib/pq/535)

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -98,6 +98,15 @@ type conn struct {
 	namei     int
 	scratch   [512]byte
 	txnStatus transactionStatus
+	txnClosed chan<- struct{}
+
+	// Save connection arguments to use during CancelRequest.
+	dialer Dialer
+	opts   values
+
+	// Cancellation key data for use with CancelRequest messages.
+	processID int
+	secretKey int
 
 	parameterStatus parameterStatus
 
@@ -307,7 +316,10 @@ func DialOpen(d Dialer, name string) (_ driver.Conn, err error) {
 		}
 	}
 
-	cn := &conn{}
+	cn := &conn{
+		opts:   o,
+		dialer: d,
+	}
 	err = cn.handleDriverSettings(o)
 	if err != nil {
 		return nil, err
@@ -529,7 +541,15 @@ func (cn *conn) Begin() (_ driver.Tx, err error) {
 	return cn, nil
 }
 
+func (cn *conn) closeTxn() {
+	if cn.txnClosed != nil {
+		close(cn.txnClosed)
+		cn.txnClosed = nil
+	}
+}
+
 func (cn *conn) Commit() (err error) {
+	defer cn.closeTxn()
 	if cn.bad {
 		return driver.ErrBadConn
 	}
@@ -565,6 +585,7 @@ func (cn *conn) Commit() (err error) {
 }
 
 func (cn *conn) Rollback() (err error) {
+	defer cn.closeTxn()
 	if cn.bad {
 		return driver.ErrBadConn
 	}
@@ -796,7 +817,11 @@ func (cn *conn) Close() (err error) {
 }
 
 // Implement the "Queryer" interface
-func (cn *conn) Query(query string, args []driver.Value) (_ driver.Rows, err error) {
+func (cn *conn) Query(query string, args []driver.Value) (driver.Rows, error) {
+	return cn.query(query, args)
+}
+
+func (cn *conn) query(query string, args []driver.Value) (_ *rows, err error) {
 	if cn.bad {
 		return nil, driver.ErrBadConn
 	}
@@ -1074,6 +1099,7 @@ func (cn *conn) startup(o values) {
 		t, r := cn.recv()
 		switch t {
 		case 'K':
+			cn.processBackendKeyData(r)
 		case 'S':
 			cn.processParameterStatus(r)
 		case 'R':
@@ -1301,6 +1327,7 @@ func (cn *conn) parseComplete(commandTag string) (driver.Result, string) {
 
 type rows struct {
 	cn       *conn
+	closed   chan<- struct{}
 	colNames []string
 	colTyps  []oid.Oid
 	colFmts  []format
@@ -1309,6 +1336,9 @@ type rows struct {
 }
 
 func (rs *rows) Close() error {
+	if rs.closed != nil {
+		defer close(rs.closed)
+	}
 	// no need to look at cn.bad as Next() will
 	for {
 		err := rs.Next(nil)
@@ -1511,6 +1541,11 @@ func (cn *conn) readReadyForQuery() {
 		cn.bad = true
 		errorf("unexpected message %q; expected ReadyForQuery", t)
 	}
+}
+
+func (c *conn) processBackendKeyData(r *readBuf) {
+	c.processID = r.int32()
+	c.secretKey = r.int32()
 }
 
 func (cn *conn) readParseResponse() {
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,11 @@
 language: go
 
 go:
-  - 1.5
-  - 1.6
-  - 1.7
-  - tip
+  - 1.5.x
+  - 1.6.x
+  - 1.7.x
+  - 1.8.x
+  - master
 
 sudo: true
 
--- a/conn_go18.go
+++ b/conn_go18.go
@@ -1,0 +1,92 @@
+// +build go1.8
+
+package pq
+
+import (
+	"context"
+	"database/sql/driver"
+	"errors"
+)
+
+// Implement the "QueryerContext" interface
+func (cn *conn) QueryContext(ctx context.Context, query string, args []driver.NamedValue) (driver.Rows, error) {
+	list := make([]driver.Value, len(args))
+	for i, nv := range args {
+		list[i] = nv.Value
+	}
+	var closed chan<- struct{}
+	if ctx.Done() != nil {
+		closed = watchCancel(ctx, cn.cancel)
+	}
+	r, err := cn.query(query, list)
+	if err != nil {
+		return nil, err
+	}
+	r.closed = closed
+	return r, nil
+}
+
+// Implement the "ExecerContext" interface
+func (cn *conn) ExecContext(ctx context.Context, query string, args []driver.NamedValue) (driver.Result, error) {
+	list := make([]driver.Value, len(args))
+	for i, nv := range args {
+		list[i] = nv.Value
+	}
+
+	if ctx.Done() != nil {
+		closed := watchCancel(ctx, cn.cancel)
+		defer close(closed)
+	}
+
+	return cn.Exec(query, list)
+}
+
+// Implement the "ConnBeginTx" interface
+func (cn *conn) BeginTx(ctx context.Context, opts driver.TxOptions) (driver.Tx, error) {
+	if opts.Isolation != 0 {
+		return nil, errors.New("isolation levels not supported")
+	}
+	if opts.ReadOnly {
+		return nil, errors.New("read-only transactions not supported")
+	}
+	tx, err := cn.Begin()
+	if err != nil {
+		return nil, err
+	}
+	if ctx.Done() != nil {
+		cn.txnClosed = watchCancel(ctx, cn.cancel)
+	}
+	return tx, nil
+}
+
+func watchCancel(ctx context.Context, cancel func()) chan<- struct{} {
+	closed := make(chan struct{})
+	go func() {
+		select {
+		case <-ctx.Done():
+			cancel()
+		case <-closed:
+		}
+	}()
+	return closed
+}
+
+func (cn *conn) cancel() {
+	var err error
+	can := &conn{}
+	can.c, err = dial(cn.dialer, cn.opts)
+	if err != nil {
+		return
+	}
+	can.ssl(cn.opts)
+
+	defer can.errRecover(&err)
+
+	w := can.writeBuf(0)
+	w.int32(80877102) // cancel request code
+	w.int32(cn.processID)
+	w.int32(cn.secretKey)
+
+	can.sendStartupPacket(w)
+	_ = can.c.Close()
+}
--- a/go18_test.go
+++ b/go18_test.go
@@ -2,7 +2,12 @@
 
 package pq
 
-import "testing"
+import (
+	"context"
+	"database/sql"
+	"testing"
+	"time"
+)
 
 func TestMultipleSimpleQuery(t *testing.T) {
 	db := openTestConn(t)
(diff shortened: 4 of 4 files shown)


rikonor:
Is this WIP?

mjibson:
Nope. Just waiting for reviews.

rikonor:
@mjibson I see.

Another question - assuming I provide a context with timeout, will an ongoing query be cancelled if the timeout is exceeded? Or is it only for queries that are currently queued for execution?

mjibson:
A cancelled context will prevent new queries and halt currently running
queries.

tamird:
To be clear, it won't cancel the queries on the server, but it will unblock
the client.

jhorowitz:
Is there a reason not to add something which would send a cancel backend signal to postgres? Unless I'm misunderstanding (which I probably am), you seem to already have the PID. From there, canceling seems pretty simple. As long as you're a superuser, you could call pg_cancel_backend or pg_terminate_backend.

mjibson:
No, it does indeed cancel queries on the server. See the conn.Cancel function in conn.go. It opens a new connection instructing the server to cancel the original connection. See https://www.postgresql.org/docs/current/static/protocol-flow.html#AEN112861

tamird:
Ah, I was recalling my experiences with MySQL, disregard.

conn.go
```diff
@@ -22,6 +22,8 @@ import (
 	"time"
 	"unicode"
 
+	"golang.org/x/net/context"
```
kr: Is this a new dependency? If we're only using this for the interface, it would also be possible to copy just the interface definition into this package, and then avoid depending on golang.org/x/net/context.

conn.go
```diff
@@ -1651,6 +1727,31 @@ func (cn *conn) readReadyForQuery() {
...
+
+	defer can.errRecover(&err)
+
+	w := can.writeBuf(0)
+	w.int32(80877102) // cancel request code
+	w.int32(cn.processID)
+	w.int32(cn.secretKey)
+
+	can.sendStartupPacket(w)
+	_ = can.c.Close()
```
kr: It seems wasteful to open a connection just to send one command, then immediately tear down the whole connection. Is there some way to save this conn for some time so it can potentially be used by the pool?

conn.go
```diff
@@ -806,46 +848,75 @@ func (cn *conn) Query(query string, args []driver.Value) (_ driver.Rows, err err
...
+		r = &rows{
 			cn:       cn,
 			colNames: st.colNames,
 			colTyps:  st.colTyps,
 			colFmts:  st.colFmts,
-		}, nil
+		}
+	}
+
+	if r != nil && ctx.Done() != context.Background().Done() {
```
kr: It looks like this is trying to avoid the overhead of an extra goroutine in the case where the context will never be canceled. It would be better to write it as:

```go
if r != nil && ctx.Done() != nil
```

Same for the other places that do this.

conn.go
```diff
@@ -806,46 +848,75 @@ func (cn *conn) Query(query string, args []driver.Value) (_ driver.Rows, err err
...
+		closed := make(chan struct{})
+		go func() {
+			select {
+			case <-ctx.Done():
+				cn.Cancel()
+			case <-closed:
+			}
+		}()
+		r.close = closed
```
kr: Maybe consider pulling this whole blob of code out into a function? It is a bit subtle, and it's repeated several times. Maybe something like:

```go
if r != nil && ctx.Done() != nil {
	r.close = watchCancel(ctx, cn.Cancel)
}
```

with 

```go
func watchCancel(ctx context.Context, cancel func()) chan<- struct{} {
	closed := make(chan struct{})
	go func() {
		select {
		case <-ctx.Done():
			cancel()
		case <-closed:
		}
	}()
	return closed
}
```

conn.go
cbandy: Nope! The [protocol for cancellation](https://www.postgresql.org/docs/current/static/protocol-flow.html#AEN112861) is to open a new connection that will immediately be closed.

context.go
```diff
@@ -0,0 +1,44 @@
...
+// contextInterface is named this way so that files tagged go1.8 can import
+// context without conflict.
+type contextInterface interface {
+	Deadline() (deadline time.Time, ok bool)
+	Done() <-chan struct{}
+	Err() error
+	Value(key interface{}) interface{}
+}
+
+type emptyCtx int
```
cbandy: As I recall, an empty struct is actually the smallest value to have lying around.

conn.go
```diff
@@ -1513,6 +1576,31 @@ func (cn *conn) readReadyForQuery() {
...
+	c.processID = r.int32()
+	c.secretKey = r.int32()
+}
+
+func (cn *conn) Cancel() {
+	var err error
+	can := &conn{}
+	can.c, err = dial(cn.dialer, cn.opts)
+	if err != nil {
+		return
```
cbandy: Is there really nothing to be done here? Retry doesn't seem appropriate...

Is this method exported for the driver package? I don't see it in the docs.

context.go
mjibson: You are correct, but in order to be a more direct copy of the context code I kept it as an int: https://tip.golang.org/pkg/context/?m=all#emptyCtx

conn.go
mjibson: I'm not sure what else we would do. The cancellation protocol itself says that the request may or may not be successful. And it's unclear what to do in the face of network error. It doesn't seem like an error here should set an error on cn either. This is currently my best idea.

I've unexported this method.

tamird:
Why do we need this "contextInterface"? If we could move all the context-related functionality under the 1.8 build tag, we could just use "context" instead.

Seems like we can achieve that by having an embedded structure in `conn` which is empty when `!go.18`.

---

Reviewed 3 of 4 files at r2.
Review status: 3 of 4 files reviewed at latest revision, 4 unresolved discussions.

---

*[conn_go18.go, line 32 at r2](https://reviewable.io:443/reviews/lib/pq/535#-Kbk8cTBR-lB9JuWkDTq:-Kbk8cTCKeTEoH1pRTjh:b-rrvgcl) ([raw file](https://github.com/lib/pq/blob/3b4b0f882cb5c99414be8887cf9cde39f5f90c42/conn_go18.go#L32)):*
> ```Go
> func (cn *conn) BeginTx(ctx context.Context, opts driver.TxOptions) (driver.Tx, error) {
> 	if opts.Isolation != 0 {
> 		return nil, errors.New("isolation levels not supported")
> ```

seems like we can support both this and the below, but i guess that's not required for this PR.

---

*[go18_test.go, line 81 at r2](https://reviewable.io:443/reviews/lib/pq/535#-Kbk8zQf8hvLl-dTufMc:-Kbk8zQf8hvLl-dTufMd:b-ye3zv3) ([raw file](https://github.com/lib/pq/blob/3b4b0f882cb5c99414be8887cf9cde39f5f90c42/go18_test.go#L81)):*
> ```Go
> 
> 	// Delay execution for just a bit until db.ExecContext has begun.
> 	go cancel()
> ```

seems inherently flaky.

---

*[go18_test.go, line 103 at r2](https://reviewable.io:443/reviews/lib/pq/535#-Kbk994bZmkNBSG3lBUu:-Kbk994bZmkNBSG3lBUv:bfz9c4y) ([raw file](https://github.com/lib/pq/blob/3b4b0f882cb5c99414be8887cf9cde39f5f90c42/go18_test.go#L103)):*
> ```Go
> 
> 	ctx, cancel := context.WithCancel(context.Background())
> 	tx, err := db.BeginTx(ctx, nil)
> ```

you need to clean up this `tx` to avoid goroutine leaks, I think.

---

*[go18_test.go, line 109 at r2](https://reviewable.io:443/reviews/lib/pq/535#-Kbk92X9zTSHUwr44XIr:-Kbk92X9zTSHUwr44XIs:b-7ropln) ([raw file](https://github.com/lib/pq/blob/3b4b0f882cb5c99414be8887cf9cde39f5f90c42/go18_test.go#L109)):*
> ```Go
> 
> 	// Delay execution for just a bit until tx.Exec has begun.
> 	go cancel()
> ```

ditto

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/lib/pq/535#-:-Kbk8FEmY1_HhUgCBBd9:b-i31uny)*

mjibson:
I've moved all the context stuff to a go 1.8-only file. I like this design much more. It also fleshed out a bug.

---

Review status: 3 of 4 files reviewed at latest revision, 4 unresolved discussions.

---

*[conn_go18.go, line 32 at r2](https://reviewable.io:443/reviews/lib/pq/535#-Kbk8cTBR-lB9JuWkDTq:-KbqIScnKsd7kgePiRfP:b-jw8e3c) ([raw file](https://github.com/lib/pq/blob/3b4b0f882cb5c99414be8887cf9cde39f5f90c42/conn_go18.go#L32)):*
<details><summary><i>Previously, tamird (Tamir Duberstein) wrote…</i></summary><blockquote>

seems like we can support both this and the below, but i guess that's not required for this PR.
</blockquote></details>

another pr for sure

---

*[go18_test.go, line 81 at r2](https://reviewable.io:443/reviews/lib/pq/535#-Kbk8zQf8hvLl-dTufMc:-KbqIMAsCgaOzIIbV78q:b-u9hgxu) ([raw file](https://github.com/lib/pq/blob/3b4b0f882cb5c99414be8887cf9cde39f5f90c42/go18_test.go#L81)):*
<details><summary><i>Previously, tamird (Tamir Duberstein) wrote…</i></summary><blockquote>

seems inherently flaky.
</blockquote></details>

I've added a 10ms sleep here. It's difficult to test this correctly.

---

*[go18_test.go, line 103 at r2](https://reviewable.io:443/reviews/lib/pq/535#-Kbk994bZmkNBSG3lBUu:-KbqELGsp3Uwhk_ZkPdl:bfha0m0) ([raw file](https://github.com/lib/pq/blob/3b4b0f882cb5c99414be8887cf9cde39f5f90c42/go18_test.go#L103)):*
<details><summary><i>Previously, tamird (Tamir Duberstein) wrote…</i></summary><blockquote>

you need to clean up this `tx` to avoid goroutine leaks, I think.
</blockquote></details>

I added leaktest to this test and it passed without any cleanup. I don't think a Tx starts any goroutines.

---

*[go18_test.go, line 109 at r2](https://reviewable.io:443/reviews/lib/pq/535#-Kbk92X9zTSHUwr44XIr:-KbqIaXDT9-nqBy0m7fG:b-896fix) ([raw file](https://github.com/lib/pq/blob/3b4b0f882cb5c99414be8887cf9cde39f5f90c42/go18_test.go#L109)):*
<details><summary><i>Previously, tamird (Tamir Duberstein) wrote…</i></summary><blockquote>

ditto
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/lib/pq/535#-:-KbqIV8yy1mBiuottwi9:b-n1scdi)*

tamird:
:lgtm:

Let's squash this down, too.

---

Reviewed 3 of 4 files at r3.
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

*[conn.go, line 101 at r3](https://reviewable.io:443/reviews/lib/pq/535#-KbqObfBgiwfKDCzR4Vl:-KbqObfBgiwfKDCzR4Vm:b3sclzh) ([raw file](https://github.com/lib/pq/blob/7738c14b8ec8df9cc97b85eacffab30011dba042/conn.go#L101)):*
> ```Go
> 	scratch   [512]byte
> 	txnStatus transactionStatus
> 	txnClose  chan<- struct{}
> ```

txnClosed for consistency?

---

*[conn.go, line 871 at r3](https://reviewable.io:443/reviews/lib/pq/535#-KbqLZygYLQIf8RWN38U:-KbqLZygYLQIf8RWN38V:bcrtz0c) ([raw file](https://github.com/lib/pq/blob/7738c14b8ec8df9cc97b85eacffab30011dba042/conn.go#L871)):*
> ```Go
> 	if len(args) == 0 {
> 		// ignore commandTag, our caller doesn't care
> 		res, _, err = cn.simpleExec(query)
> ```

unrelated change?

---

*[conn.go, line 889 at r3](https://reviewable.io:443/reviews/lib/pq/535#-KbqOWKVAw0itaiqUg9-:-KbqOWKVAw0itaiqUg90:b1pf97l) ([raw file](https://github.com/lib/pq/blob/7738c14b8ec8df9cc97b85eacffab30011dba042/conn.go#L889)):*
> ```Go
> 		// used.
> 		st := cn.prepareTo(query, "")
> 		res, err = st.Exec(args)
> ```

also unrelated?

---

*[conn.go, line 1330 at r3](https://reviewable.io:443/reviews/lib/pq/535#-KbqOS4P0TQCyZgAbqNd:-KbqOS4P0TQCyZgAbqNe:b-dczsg7) ([raw file](https://github.com/lib/pq/blob/7738c14b8ec8df9cc97b85eacffab30011dba042/conn.go#L1330)):*
> ```Go
> type rows struct {
> 	cn       *conn
> 	close    chan<- struct{}
> ```

consider calling this `closed` for consistency

---

*[conn_go18.go, line 18 at r3](https://reviewable.io:443/reviews/lib/pq/535#-KbqOLFo26sQrWEfhGPD:-KbqOLFo26sQrWEfhGPE:b-79dwbd) ([raw file](https://github.com/lib/pq/blob/7738c14b8ec8df9cc97b85eacffab30011dba042/conn_go18.go#L18)):*
> ```Go
> 	}
> 	var closed chan<- struct{}
> 	if ctx.Done() != nil {
> ```

move this below `r, err := ...` and then you can just set `r.close` directly instead of having `closed` in scope

---

*[go18_test.go, line 103 at r2](https://reviewable.io:443/reviews/lib/pq/535#-Kbk994bZmkNBSG3lBUu:-KbqL07R7inHX722MMI4:bdj3y9f) ([raw file](https://github.com/lib/pq/blob/3b4b0f882cb5c99414be8887cf9cde39f5f90c42/go18_test.go#L103)):*
<details><summary><i>Previously, mjibson (Matt Jibson) wrote…</i></summary><blockquote>

I added leaktest to this test and it passed without any cleanup. I don't think a Tx starts any goroutines.
</blockquote></details>

Evidence to the contrary: https://github.com/cockroachdb/cockroach/pull/13197/commits/9bf770cd8f6eaff5441b80d3aec1a5614e8747e1

Cancelling the contexts in these tests is likely what's making it work.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/lib/pq/535#-:-KbqOs3_Sv6emRfheVxt:b-9bjf2d)*

mjibson:
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

*[conn.go, line 101 at r3](https://reviewable.io:443/reviews/lib/pq/535#-KbqObfBgiwfKDCzR4Vl:-KbqPy4nGdSIUk51g6Gv:b-896fix) ([raw file](https://github.com/lib/pq/blob/7738c14b8ec8df9cc97b85eacffab30011dba042/conn.go#L101)):*
<details><summary><i>Previously, tamird (Tamir Duberstein) wrote…</i></summary><blockquote>

txnClosed for consistency?
</blockquote></details>

Done.

---

*[conn.go, line 871 at r3](https://reviewable.io:443/reviews/lib/pq/535#-KbqLZygYLQIf8RWN38U:-KbqQAM1jBQweDeQKIeK:bhjxxii) ([raw file](https://github.com/lib/pq/blob/7738c14b8ec8df9cc97b85eacffab30011dba042/conn.go#L871)):*
<details><summary><i>Previously, tamird (Tamir Duberstein) wrote…</i></summary><blockquote>

unrelated change?
</blockquote></details>

yes, removed

---

*[conn.go, line 889 at r3](https://reviewable.io:443/reviews/lib/pq/535#-KbqOWKVAw0itaiqUg9-:-KbqQB5SQ4VPOYxbx0tW:b-hfpv0l) ([raw file](https://github.com/lib/pq/blob/7738c14b8ec8df9cc97b85eacffab30011dba042/conn.go#L889)):*
<details><summary><i>Previously, tamird (Tamir Duberstein) wrote…</i></summary><blockquote>

also unrelated?
</blockquote></details>

removed

---

*[conn.go, line 1330 at r3](https://reviewable.io:443/reviews/lib/pq/535#-KbqOS4P0TQCyZgAbqNd:-KbqQk1SgmhRW_GB9tAD:b-896fix) ([raw file](https://github.com/lib/pq/blob/7738c14b8ec8df9cc97b85eacffab30011dba042/conn.go#L1330)):*
<details><summary><i>Previously, tamird (Tamir Duberstein) wrote…</i></summary><blockquote>

consider calling this `closed` for consistency
</blockquote></details>

Done.

---

*[conn_go18.go, line 18 at r3](https://reviewable.io:443/reviews/lib/pq/535#-KbqOLFo26sQrWEfhGPD:-KbqQsko94EMwbVEOWq8:bz7iaxo) ([raw file](https://github.com/lib/pq/blob/7738c14b8ec8df9cc97b85eacffab30011dba042/conn_go18.go#L18)):*
<details><summary><i>Previously, tamird (Tamir Duberstein) wrote…</i></summary><blockquote>

move this below `r, err := ...` and then you can just set `r.close` directly instead of having `closed` in scope
</blockquote></details>

That's how it was before, but it's a bug. We need to start the cancellation listening before the first row is returned. If this is moved below the cn.query line, then only when we have a rows object are we able to cancel a query.

---

*[go18_test.go, line 103 at r2](https://reviewable.io:443/reviews/lib/pq/535#-Kbk994bZmkNBSG3lBUu:-KbqRV8flJy908XZclrI:b6yte4n) ([raw file](https://github.com/lib/pq/blob/3b4b0f882cb5c99414be8887cf9cde39f5f90c42/go18_test.go#L103)):*
<details><summary><i>Previously, tamird (Tamir Duberstein) wrote…</i></summary><blockquote>

Evidence to the contrary: https://github.com/cockroachdb/cockroach/pull/13197/commits/9bf770cd8f6eaff5441b80d3aec1a5614e8747e1

Cancelling the contexts in these tests is likely what's making it work.
</blockquote></details>

Adding tx.Rollback() at the end produces an error: `sql: Transaction has already been committed or rolled back`. I'm not convinced adding a check to make sure tx.Rollback returns that error is useful. But I'm getting pretty convinced it doesn't have a leaky go routine if it already thinks it's been rolled back.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/lib/pq/535)*

tamird:
Can you add 1.8.x to the travis list? In fact, all the versions should end in `.x`, and `tip` is apparently now spelled `master`. https://docs.travis-ci.com/user/languages/go#Specifying-a-Go-version-to-use

---

Reviewed 2 of 2 files at r4.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[conn_go18.go, line 18 at r3](https://reviewable.io:443/reviews/lib/pq/535#-KbqOLFo26sQrWEfhGPD:-KbqTBz3A863v0BgyT9I:b-37ujzf) ([raw file](https://github.com/lib/pq/blob/7738c14b8ec8df9cc97b85eacffab30011dba042/conn_go18.go#L18)):*
<details><summary><i>Previously, mjibson (Matt Jibson) wrote…</i></summary><blockquote>

That's how it was before, but it's a bug. We need to start the cancellation listening before the first row is returned. If this is moved below the cn.query line, then only when we have a rows object are we able to cancel a query.
</blockquote></details>

Ah, thanks for explaining.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/lib/pq/535#-:-KbqTUm0CPK5fJF-2aTE:b-xukp9g)*

nathany:
thanks! ref #531

PR #541, opened 2016-11-16
Add rows.{Result,Tag}() to enable access to the result/tag in queries.

mjibson:
Enable access to the driver result (rows affected) and command
completion tag for statements executed via `conn.Query`.

This can be used in programs that access the connection mechanism directly
instead of through database/sql. CockroachDB is one such example.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -664,6 +664,9 @@ func (cn *conn) simpleQuery(q string) (res *rows, err error) {
 				res = &rows{
 					cn: cn,
 				}
+				if t == 'C' {
+					res.result, res.tag = cn.parseComplete(r.string())
+				}
 			}
 			res.done = true
 		case 'Z':
@@ -1333,6 +1336,8 @@ type rows struct {
 	colFmts  []format
 	done     bool
 	rb       readBuf
+	result   driver.Result
+	tag      string
 }
 
 func (rs *rows) Close() error {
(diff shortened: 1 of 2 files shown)


cbandy:
Needs tests. At first glance, I don't think these messages happen for queries that return result sets.

mjibson:
Added tests. Indeed SELECT messages don't produce tags and driver.Results, but if one is, say, running everything the user is sending via .Query because you are processing input from the commandline, then .Result and .Tag become very useful. Again, this is exactly what we are doing in cockroach. Merging this PR would allow us to remove our fork and use the main repo again. The cost to the user is an extra string and int and the time it takes to parse them per rows object.

knz:
LGTM

mjibson:
@tamird mind reviewing this?

tamird:
So the idea is that users of this functionality need to write their own ResultTag interface in order to get at these methods? This feel like feature creep to me - this library doesn't provide many postgres-specific facilities that aren't exposed by database/sql, and I think it's our responsibility to keep it that way.

There are other libraries that do not share this philosophy (such as https://github.com/jackc/pgx) - we should not be making changes to lib/pq that are motivated by our unique development goals in CRDB.

mjibson:
The argument against why this is feature creep is that this feature is useful to any package that needs the data in tag and result. This is not many packages, true, but relying on just database/sql doesn't allow this level of detail about the returned data. Cockroach uses this in its command line package to emulate psql, which prints tags after statements. Any other application attempting to create a command line similar to psql will also need this functionality. So the argument is that this is useful, just to a small group of applications. It is not a cockroach-specific feature (although cockroach is the first to need it).

tamird:
Can you generalize that rationale and include it in the commit message? LG following that.

PR #545, opened 2016-11-23
Handle nil and NULL values consistently for Array types

### LINKED ISSUE #540: StringArray Value method returns nil when slice is nil, I was expecting '{}' instead.
This was a surprise to me since usually Go's SQL types are not null and there are explicit types to allow inserting null values in the database.  Is this behavior the intended by design or likely a bug?

cbandy:
`Value()` of a nil slice is NULL, and `Scan()` of NULL sets a slice to nil.

Fixes #540

### DIFF
--- a/array.go
+++ b/array.go
@@ -70,6 +70,9 @@ func (a *BoolArray) Scan(src interface{}) error {
 		return a.scanBytes(src)
 	case string:
 		return a.scanBytes([]byte(src))
+	case nil:
+		*a = nil
+		return nil
 	}
 
 	return fmt.Errorf("pq: cannot convert %T to BoolArray", src)
@@ -80,7 +83,7 @@ func (a *BoolArray) scanBytes(src []byte) error {
 	if err != nil {
 		return err
 	}
-	if len(elems) == 0 {
+	if *a != nil && len(elems) == 0 {
 		*a = (*a)[:0]
 	} else {
 		b := make(BoolArray, len(elems))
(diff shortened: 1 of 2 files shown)


cbandy:
I just discovered that an empty array value is returned as `nil` when the scan destination is `nil`. For example,

```go
var x []string
db.QueryRow(`SELECT ARRAY[]::text[]`).Scan(pq.Array(&x))
// x == nil
```

Here's a snippet showing similar stuff: https://play.golang.org/p/-HGKEG3_s8

Since the intent of this PR is to distinguish NULL values as `nil`, I think an empty array value should be returned as initialized, empty slice. (GH won't let me review my own PR, but consider this rejected by me until this ^ is addressed.)

cbandy:
> I think an empty array value should be returned as initialized, empty slice.

Pushed in an amended commit. This was another case where Value had the expected behavior but Scan did not.

mjibson:
I can't see what changed to this changeset from the previous, but things still LGTM. Maybe in the future just push it as a second commit for review, and then do a squash before merge? We've been doing that here when we need to use github reviews and it works pretty well.

cbandy:
Thanks. Will do.

I added the ScanEmpty tests and guards against nil in hunks like these:

```diff
-	if len(elems) == 0 {
+	if *a != nil && len(elems) == 0 {
  		*a = (*a)[:0]
  	} else {
```

mikeatlas:
Looking for this to be patched soon!

mjibson:
@cbandy do you want to merge this? It's your PR and I think you have write access.

cbandy:
I don't have write access.

mjibson:
Sorry, thought you did.

Issue #555, opened 2016-12-24
Errors for unique constraint violations have empty error data


TheBeege:
When a unique constraint violation occurs, the additional data in the Error object is missing, specifically Constraint, Table, Column, and Schema. At the least, Constraint, Table, and Schema should be populated according to https://www.postgresql.org/docs/current/static/protocol-error-fields.html.

I'll submit a PR for the test cases. I'll also see if I can fix it myself and submit a separate PR.

This would be my first major project code contribution, so tips appreciated

TheBeege:
I cracked open ParseError and checked out the message output. It seems like PG isn't providing the necessary information. Looking at the docs again, the note at the bottom states the following: "Frontends should not assume that the presence of any of these fields guarantees the presence of another field." That _really_ sucks. Maybe this should be a "won't fix"? Could certainly add a whole bunch of junk to find and set that additional error info, but it may not be worth it. I also imagine it would cause some performance issues.

I'll keep poking around to see if I can find something that would help.

TheBeege:
I re-read the docs. This info is only populated for "SQLSTATE class 23 (integrity constraint violation)". Closing this issue.

PR #556, opened 2016-12-24
Added test cases for #555

TheBeege:


TheBeege:
Per #555, PG does not yet supply these fields to front-ends for unique constraint violations

PR #557, opened 2017-01-16
add IntArray type

gazoon:
Now you can use []int slice in the same way as []int64 before.


koalalorenzo:
I guess nobody cared :O

PR #564, opened 2017-02-11
Return last instead of first tag

mjibson:


johto:
Shouldn't we also do the same thing with rows here?  Effectively, always overwrite instead of checking whether res is null.

mjibson:
No, because a `C` comes after row data. That is, we always expect a `C`, and sometimes it will have been proceeded by some row data (which has already populated res). But in the case that there are no rows returned, it will be just a `C`.

mjibson:
I think my analysis above is slightly wrong, although the conclusion may be correct. I was wrong because if there's row data, then this function normally returns and Next takes over, so we could safely set res always in that case. I ran the tests with your suggestion and they failed on a query of specified columns but with empty results:

```
--- FAIL: TestEmptyResultSetColumns (0.00s)
	conn_test.go:464: unexpected number of columns 0 in response to an empty query
```

So although it normally works, in the case of empty results it doesn't. I've updated the PR and added a test that tests for this case.

tamird:
LGTM as a temporary measure to unblock downstream work.

As discussed in person, I'd like to see a refactoring makes the behaviour surrounding handling of multiple result sets a bit more rigorous, which would also allow you to retrieve all results and tags.

PR #565, opened 2017-02-13
Remove unneeded for loop from listen example.

icholy:
Since both cases in the select return, the for loop has no effect.


freeformz:
Thanks!

icholy:
that was quick!

PR #571, opened 2017-03-04
Add missing OID types from Postgres 9.6

jordanlewis:
Several new OID types were added since the last time types.go was
updated. This commit adds the missing types as of Postgres 9.6.2.


jordanlewis:
cc @mjibson

PR #573, opened 2017-03-06
ssl: only set "sslrootcert" once

### LINKED ISSUE #570: Concurrent map writes
Hi,

i have faced such issue today:

```
fatal error: concurrent map writes
goroutine 4155 [running]:
runtime.throw(0xb18b1d, 0x15)
	/usr/local/go/src/runtime/panic.go:596 +0x95 fp=0xc420672af0 sp=0xc420672ad0
runtime.mapassign(0xa80880, 0xc4204898c0, 0xc420672c08, 0x0)
	/usr/local/go/src/runtime/hashmap.go:589 +0x2a1 fp=0xc420672b90 sp=0xc420672af0
<PROJECT_PATH>/vendor/github.com/lib/pq.ssl(0xc4204898c0, 0x0)
	/go/src/<PROJECT_PATH>/vendor/github.com/lib/pq/ssl.go:34 +0x27d fp=0xc420672da0 sp=0xc420672b90
<PROJECT_PATH>/vendor/github.com/lib/pq.(*conn).ssl(0xc420a178c0, 0xc4204898c0)
	/go/src/<PROJECT_PATH>/vendor/github.com/lib/pq/conn.go:1033 +0x40 fp=0xc420672e40 sp=0xc420672da0
<PROJECT_PATH>/vendor/github.com/lib/pq.(*conn).cancel(0xc420b3a2c0)
	/go/src/<PROJECT_PATH>/vendor/github.com/lib/pq/conn_go18.go:81 +0xf2 fp=0xc420672ef0 sp=0xc420672e40
<PROJECT_PATH>/vendor/github.com/lib/pq.(*conn).(<PROJECT_PATH>/vendor/github.com/lib/pq.cancel)-fm()
	/go/src/<PROJECT_PATH>/vendor/github.com/lib/pq/conn_go18.go:19 +0x2a fp=0xc420672f08 sp=0xc420672ef0
<PROJECT_PATH>/vendor/github.com/lib/pq.watchCancel.func1(0xece220, 0xc42040f170, 0xc420493400, 0xc4203674a0)
	/go/src/<PROJECT_PATH>/vendor/github.com/lib/pq/conn_go18.go:67 +0xc4 fp=0xc420672fc0 sp=0xc420672f08
runtime.goexit()
```

**pq version**: `ba5d4f7a35561e22fbdf7a39aa0070f4d460cfc0`
**postgres version**: `9.2`
**go version**: `1.8`

tamird:
Fixes #570, though I have not been able to write a test for this.

Remove methods on `values`. These methods are not a useful abstraction
and only serve to obscure that the underlying structure is a map. Their
presence is particularly inconvenient in that they promote confusing
empty values with the lack of value.

---
This change is [Reviewable](https://reviewable.io/reviews/lib/pq/573)

### DIFF
--- a/ssl.go
+++ b/ssl.go
@@ -15,7 +15,7 @@ import (
 func ssl(o values) func(net.Conn) net.Conn {
 	verifyCaOnly := false
 	tlsConf := tls.Config{}
-	switch mode := o.Get("sslmode"); mode {
+	switch mode := o["sslmode"]; mode {
 	// "require" is the default.
 	case "", "require":
 		// We must skip TLS's own verification since it requires full
@@ -23,15 +23,19 @@ func ssl(o values) func(net.Conn) net.Conn {
 		tlsConf.InsecureSkipVerify = true
 
 		// From http://www.postgresql.org/docs/current/static/libpq-ssl.html:
-		// Note: For backwards compatibility with earlier versions of PostgreSQL, if a
-		// root CA file exists, the behavior of sslmode=require will be the same as
-		// that of verify-ca, meaning the server certificate is validated against the
-		// CA. Relying on this behavior is discouraged, and applications that need
-		// certificate validation should always use verify-ca or verify-full.
-		if _, err := os.Stat(o.Get("sslrootcert")); err == nil {
-			verifyCaOnly = true
-		} else {
-			o.Set("sslrootcert", "")
+		//
+		// Note: For backwards compatibility with earlier versions of
+		// PostgreSQL, if a root CA file exists, the behavior of
+		// sslmode=require will be the same as that of verify-ca, meaning the
+		// server certificate is validated against the CA. Relying on this
+		// behavior is discouraged, and applications that need certificate
+		// validation should always use verify-ca or verify-full.
+		if sslrootcert, ok := o["sslrootcert"]; ok {
+			if _, err := os.Stat(sslrootcert); err == nil {
+				verifyCaOnly = true
+			} else {
+				delete(o, "sslrootcert")
+			}
 		}
 	case "verify-ca":
 		// We must skip TLS's own verification since it requires full
@@ -39,7 +43,7 @@ func ssl(o values) func(net.Conn) net.Conn {
 		tlsConf.InsecureSkipVerify = true
 		verifyCaOnly = true
 	case "verify-full":
-		tlsConf.ServerName = o.Get("host")
+		tlsConf.ServerName = o["host"]
 	case "disable":
 		return nil
 	default:
@@ -64,37 +68,42 @@ func ssl(o values) func(net.Conn) net.Conn {
 // in the user's home directory. The configured files must exist and have
 // the correct permissions.
 func sslClientCertificates(tlsConf *tls.Config, o values) {
-	sslkey := o.Get("sslkey")
-	sslcert := o.Get("sslcert")
+	// user.Current() might fail when cross-compiling. We have to ignore the
+	// error and continue without home directory defaults, since we wouldn't
+	// know from where to load them.
+	user, _ := user.Current()
 
-	var cinfo, kinfo os.FileInfo
-	var err error
-
-	if sslcert != "" && sslkey != "" {
-		// Check that both files exist. Note that we don't do any more extensive
-		// checks than this (such as checking that the paths aren't directories);
-		// LoadX509KeyPair() will take care of the rest.
-		cinfo, err = os.Stat(sslcert)
-		if err != nil {
-			panic(err)
-		}
-
-		kinfo, err = os.Stat(sslkey)
-		if err != nil {
-			panic(err)
-		}
-	} else {
-		// Automatically find certificates from ~/.postgresql
-		sslcert, sslkey, cinfo, kinfo = sslHomeCertificates()
-
-		if cinfo == nil || kinfo == nil {
-			// No certificates to load
-			return
-		}
+	// In libpq, the client certificate is only loaded if the setting is not blank.
+	//
+	// https://github.com/postgres/postgres/blob/REL9_6_2/src/interfaces/libpq/fe-secure-openssl.c#L1036-L1037
+	sslcert := o["sslcert"]
+	if len(sslcert) == 0 && user != nil {
+		sslcert = filepath.Join(user.HomeDir, ".postgresql", "postgresql.crt")
+	}
+	// https://github.com/postgres/postgres/blob/REL9_6_2/src/interfaces/libpq/fe-secure-openssl.c#L1045
+	if len(sslcert) == 0 {
+		return
+	}
+	// https://github.com/postgres/postgres/blob/REL9_6_2/src/interfaces/libpq/fe-secure-openssl.c#L1050:L1054
+	if _, err := os.Stat(sslcert); os.IsNotExist(err) {
+		return
+	} else if err != nil {
+		panic(err)
 	}
 
-	// The files must also have the correct permissions
-	sslCertificatePermissions(cinfo, kinfo)
+	// In libpq, the ssl key is only loaded if the setting is not blank.
+	//
+	// https://github.com/postgres/postgres/blob/REL9_6_2/src/interfaces/libpq/fe-secure-openssl.c#L1123-L1222
+	sslkey := o["sslkey"]
+	if len(sslkey) == 0 && user != nil {
+		sslkey = filepath.Join(user.HomeDir, ".postgresql", "postgresql.key")
+	}
+
+	if len(sslkey) > 0 {
+		if err := sslKeyPermissions(sslkey); err != nil {
+			panic(err)
+		}
+	}
 
 	cert, err := tls.LoadX509KeyPair(sslcert, sslkey)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


ssl.go
```diff
@@ -15,39 +15,55 @@ import (
...
+				cert, err := ioutil.ReadFile(sslrootcert)
+				if err != nil {
+					panic(err)
+				}
+
+				if ok := tlsConf.RootCAs.AppendCertsFromPEM(cert); !ok {
+					errorf("couldn't parse pem in sslrootcert")
+				}
+			} else {
+				delete(o, "sslrootcert")
```
cbandy: iiuc, this has the same problem described in #570. We shouldn't be writing to `o`.

ssl.go
```diff
@@ -15,39 +15,55 @@ import (
...
-		tlsConf.ServerName = o.Get("host")
+		tlsConf.ServerName = o["host"]
 	case "disable":
 		return nil
 	default:
 		errorf(`unsupported sslmode %q; only "require" (default), "verify-full", "verify-ca", and "disable" supported`, mode)
 	}
 
 	sslClientCertificates(&tlsConf, o)
-	sslCertificateAuthority(&tlsConf, o)
```
cbandy: This function (or its contents) still need to be invoked when `sslmode` is `verify-ca` or `verify-full`.

cbandy:
I can't tell if it's appropriate to treat `o.Get(...) != ""` the same as `_, ok := o[...]; ok`.

ssl.go
cbandy: See failing tests `TestSSLVerifyCA` and `TestSSLVerifyFull`.

tamird:
All good points, thanks for your review. I'll update it shortly.

ssl.go
tamird: I don't think so, because it's now guarded by the presence of a non-empty value, which should only happen once at Dial and not during cancellation.

ssl.go
tamird: Done.

ssl.go
mjibson: I agree that this is safe and fixes the bug there because when this function is re-invoked during cancel, `sslrootcert` will have been deleted, and so this branch will not be entered.

tamird:
@cbandy addressed your concrete comments; anything else you'd like to see done before merging?

ssl.go
```diff
@@ -64,13 +68,13 @@ func ssl(o values) func(net.Conn) net.Conn {
...
-	sslkey := o.Get("sslkey")
-	sslcert := o.Get("sslcert")
+	sslkey, sslkeyOK := o["sslkey"]
+	sslcert, sslcertOK := o["sslcert"]
 
 	var cinfo, kinfo os.FileInfo
 	var err error
 
-	if sslcert != "" && sslkey != "" {
+	if sslkeyOK && sslcertOK {
```
cbandy: As I understand it, this change breaks compatibility with `libpq` connection strings.

If I'm reading `libpq` source correctly, these settings fallback to the user/home paths when the supplied value is blank:

- [`if (conn->sslcert && strlen(conn->sslcert) > 0) ... else ... "%s/%s", homedir, USER_CERT_FILE`](https://github.com/postgres/postgres/blob/REL9_6_2/src/interfaces/libpq/fe-secure-openssl.c#L1036-L1039)
- [`if (conn->sslkey && strlen(conn->sslkey) > 0) ... else ... "%s/%s", homedir, USER_KEY_FILE`](https://github.com/postgres/postgres/blob/REL9_6_2/src/interfaces/libpq/fe-secure-openssl.c#L1219-L1222)

ssl.go
```diff
@@ -105,16 +109,15 @@ func sslClientCertificates(tlsConf *tls.Config, o values) {
 
 // sslCertificateAuthority adds the RootCA specified in the "sslrootcert" setting.
 func sslCertificateAuthority(tlsConf *tls.Config, o values) {
-	if sslrootcert := o.Get("sslrootcert"); sslrootcert != "" {
+	if sslrootcert, ok := o["sslrootcert"]; ok {
```
cbandy: As I understand it, this change breaks compatibility with `libpq` connection strings.

If I'm reading `libpq` source correctly, the CA certificate is only loaded when this setting is not blank:

- [`if (conn->sslrootcert && strlen(conn->sslrootcert) > 0)`](https://github.com/postgres/postgres/blob/REL9_6_2/src/interfaces/libpq/fe-secure-openssl.c#L950-L957)

cbandy:
If compatibility with `libpq` connection strings is a goal, then maybe we should also support the home directory fallback for [`sslrootcert`](https://www.postgresql.org/docs/current/static/libpq-connect.html#LIBPQ-CONNECT-SSLROOTCERT), `~/.postgresql/root.crt`.

ssl.go
```diff
@@ -15,31 +15,35 @@ import (
...
-		// Note: For backwards compatibility with earlier versions of PostgreSQL, if a
-		// root CA file exists, the behavior of sslmode=require will be the same as
-		// that of verify-ca, meaning the server certificate is validated against the
-		// CA. Relying on this behavior is discouraged, and applications that need
-		// certificate validation should always use verify-ca or verify-full.
-		if _, err := os.Stat(o.Get("sslrootcert")); err == nil {
-			verifyCaOnly = true
-		} else {
-			o.Set("sslrootcert", "")
+		// From http://www.postgresql.org/docs/current/static/libpq- ssl.html:
```
tamird: Done.

ssl.go
tamird: The old code is also incompatible with that `libpq` logic in that `libpq` allows either the certificate or the key to be blank, while this only does the right thing if both are blank.

I'm inclined to "break" compatibility in this way until someone reports that they need it, at which time we should add tests and implement the actual `libpq` behaviour.

ssl.go
tamird: I think you're right, but as mentioned above, I think this is OK until we get a report of someone relying on this behaviour. WDYT?

tamird: Fixed and added a test.

ssl.go
tamird: Fixed this, I think, though I think tests are going to break and I've not yet explicitly tested the empty string case.

tamird:
Well, in general, `values` definitely promotes bad practices. That said, fixing the regressions you pointed out and adding tests.

tamird:
@cbandy OK, I think this is good to go with new tests exercising the treatment of empty `sslcert` and `sslkey` options.

mjibson:
Review status: 0 of 6 files reviewed at latest revision, 1 unresolved discussion.

---

*[ssl_test.go, line 44 at r1](https://reviewable.io:443/reviews/lib/pq/573#-KejMy7Voy0WoTQIKtYy:-KejMy7WFlWzxUfD8jO1:bqmtjok) ([raw file](https://github.com/lib/pq/blob/2df7fd4097b0afbbeb771dfd9b5a683177179ba9/ssl_test.go#L44)):*
> ```Go
> 
> func checkSSLSetup(t *testing.T, conninfo string) {
> 	_, err := openSSLConn(t, conninfo)
> ```

Why don't we need to close the db here?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/lib/pq/573)*

tamird:
Review status: 0 of 6 files reviewed at latest revision, 1 unresolved discussion.

---

*[ssl_test.go, line 44 at r1](https://reviewable.io:443/reviews/lib/pq/573#-KejMy7Voy0WoTQIKtYy:-KejTdFXpOc4bnQ9WosA:b-ula95q) ([raw file](https://github.com/lib/pq/blob/2df7fd4097b0afbbeb771dfd9b5a683177179ba9/ssl_test.go#L44)):*
<details><summary><i>Previously, mjibson (Matt Jibson) wrote…</i></summary><blockquote>

Why don't we need to close the db here?
</blockquote></details>

I mean, we do, but it doesn't really matter since a nil error will result in a test failure, and the leaked database handle won't turn up anywhere.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/lib/pq/573)*

mjibson:
Review status: 0 of 6 files reviewed at latest revision, 1 unresolved discussion.

---

*[ssl_test.go, line 44 at r1](https://reviewable.io:443/reviews/lib/pq/573#-KejMy7Voy0WoTQIKtYy:-KejTwgh7n7X7VghLf7T:b2q3orv) ([raw file](https://github.com/lib/pq/blob/2df7fd4097b0afbbeb771dfd9b5a683177179ba9/ssl_test.go#L44)):*
<details><summary><i>Previously, tamird (Tamir Duberstein) wrote…</i></summary><blockquote>

I mean, we do, but it doesn't really matter since a nil error will result in a test failure, and the leaked database handle won't turn up anywhere.
</blockquote></details>

Will it remain open on the server until a timeout? When running `go test -count 100 -race` trying to find race conditions I've had the pg server complain about too many connections, so I want to make sure we're closing things correctly.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/lib/pq/573)*

tamird:
Review status: 0 of 6 files reviewed at latest revision, 1 unresolved discussion.

---

*[ssl_test.go, line 44 at r1](https://reviewable.io:443/reviews/lib/pq/573#-KejMy7Voy0WoTQIKtYy:-KejUK-JxFp147trbwAq:bjtwhnk) ([raw file](https://github.com/lib/pq/blob/2df7fd4097b0afbbeb771dfd9b5a683177179ba9/ssl_test.go#L44)):*
<details><summary><i>Previously, mjibson (Matt Jibson) wrote…</i></summary><blockquote>

Will it remain open on the server until a timeout? When running `go test -count 100 -race` trying to find race conditions I've had the pg server complain about too many connections, so I want to make sure we're closing things correctly.
</blockquote></details>

There will never be a non-nil connection here in a passing test run.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/lib/pq/573)*

mjibson:
Well then LGTM, again. I agree that there is (was?) a problem since empty strings had their meaning changed in this PR. It seems like this code handles those in at least a well documented way, and if there are missing combinations we can add those as test cases.

tamird:
Thanks, @mjibson. I'll wait for final approval from @cbandy or else merge this tomorrow.

Issue #575, opened 2017-03-08
Change user_posix build tag


johanbrandhorst:
Hi! This is a bit of a weird issue but bear with me. It has a fairly simple solution.

## Synopsis: Ignoring the solaris build tag through `govendor` while using pq as a dependency means it cannot be built on linux.

Why is this a problem? The dependency management tool `govendor` is a popular choice in todays pre-blessed world of dependency management and it allows users to ignore/avoid downloading parts of dependant packages that it does not want or need, through ignoring build tags.

For example, if a package depends on the popular logger `logrus`, govendor will vendor 100k LOC of `golang.org/x/sys/unix` because of a little import in a Solaris specific part of logrus. Any customer of `logrus` not using solaris would be wise to add an ignore to the build tag `solaris`, and save themselves some dependencies they do not use.

The way build tag ignoring in `govendor` works seems to be by ignoring any files that contain the build tag that has been ignored in any dependencies. This means that when someone who has duly ignored `solaris` tries to install `pq`, it will fail to build.

This is because of the file https://github.com/lib/pq/blob/ca5bc43047f2138703da0f3d3ca89a59f3d597f1/user_posix.go, which of course contains the solaris build tag.

So what to do? I think it should be possible to revert the build tagging in user_posix.go and still achieve the desired result, e.g. change the tag to `!windows` and everything should continue on as it was, as as far as I can tell the only supported platforms are the posix and windows platforms. Since the user_windows.go build is only built for windows (implicitly via its name), the other file can simply build for anything _but_ windows.

tamird:
Seems like a bug in govendor to me.

johanbrandhorst:
Could very well be! I'm mostly trying to find a solution.

tamird:
Have you filed an issue in govendor?

johanbrandhorst:
No, but from experience it seems @kardianos does not have much time for govendor nowadays. I can certainly raise the issue there if this kind of a change is unaccaptable. Would be interesting to hear his opinion on the problem.

kardianos:
This is a bug in govendor. Most of govendor problems are around fetching code. This one (around correctly interpreting build tags) is something I would actually like to address in a near term solution.

I think this is issue https://github.com/kardianos/govendor/issues/289 .  I would recommend this issue be closed in favor of it.

@johanbrandhorst You are correct, I haven't had much time to maintain govendor recently. Let me see if I can fix this one quickly though, it is a correctness issue and somewhat of a blocker.

[...]

PR #576, opened 2017-03-08
Change user_posix build tag to !windows

johanbrandhorst:
Fixes issue #575


mjibson:
This LGTM because the corresponding windows file is tagged (well, named) to only build on windows. Someone else should look too, though.

johanbrandhorst:
@mjibson indeed, see my rationale in #575. It might be a problem that shouldn't be fixed here though, as stated by @tamird.

johanbrandhorst:
Closing this as it is not a bug in pq. See discussion in #575

PR #578, opened 2017-03-13
Fix race condition in query cancellation

### LINKED ISSUE #577: pq: canceling statement due to user request
Hello there,

Recently, I've been starting to use context queries, but not everywhere.

And in my CI I see some `pq: canceling statement due to user request` appearing where there was no problem before.

In those cases `ctx.Err()` is nil.

What is that error meaning globally ? Is this a problem in my code that was uncovered by the use of context ?? Or could the context support in pq do wrong things ?

EDIT: I'll try to make a more precise case

tamird:
It was previously possible for context cancellation to interleave with
query execution in such a way that the watcher goroutine would not be
scheduled until after the query which it was meant to watch was
complete, and would end up issueing a cancellation for the subsequent
query.

Fixes #577.

---
This change is [Reviewable](https://reviewable.io/reviews/lib/pq/578)

### DIFF
--- a/conn_go18.go
+++ b/conn_go18.go
@@ -6,6 +6,8 @@ import (
 	"context"
 	"database/sql/driver"
 	"errors"
+	"io"
+	"io/ioutil"
 )
 
 // Implement the "QueryerContext" interface
@@ -14,12 +16,12 @@ func (cn *conn) QueryContext(ctx context.Context, query string, args []driver.Na
 	for i, nv := range args {
 		list[i] = nv.Value
 	}
-	closed := cn.watchCancel(ctx)
+	finish := cn.watchCancel(ctx)
 	r, err := cn.query(query, list)
 	if err != nil {
 		return nil, err
 	}
-	r.closed = closed
+	r.finish = finish
 	return r, nil
 }
 
@@ -30,8 +32,8 @@ func (cn *conn) ExecContext(ctx context.Context, query string, args []driver.Nam
 		list[i] = nv.Value
 	}
 
-	if closed := cn.watchCancel(ctx); closed != nil {
-		defer close(closed)
+	if finish := cn.watchCancel(ctx); finish != nil {
+		defer finish()
 	}
 
 	return cn.Exec(query, list)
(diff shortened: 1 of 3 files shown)


conn_go18.go
```diff
@@ -49,41 +51,51 @@ func (cn *conn) BeginTx(ctx context.Context, opts driver.TxOptions) (driver.Tx,
...
-	defer can.errRecover(&err)
-
 	w := can.writeBuf(0)
 	w.int32(80877102) // cancel request code
 	w.int32(cn.processID)
 	w.int32(cn.secretKey)
 
 	can.sendStartupPacket(w)
-	_ = can.c.Close()
+	can.sendStartupPacket(w)
```
mjibson: sendStartupPacket panics on various errors: https://github.com/lib/pq/blob/e4af84aab01e159ca479940dbde304519e8e8811/conn.go#L896

I'm not sure removing the errRecover is safe.

conn_go18.go
```diff
@@ -49,41 +51,51 @@ func (cn *conn) BeginTx(ctx context.Context, opts driver.TxOptions) (driver.Tx,
...
+	// Read until EOF to ensure that the server received the cancel.
+	if _, err := io.Copy(ioutil.Discard, can.c); err != nil {
+		return err
```
mjibson: This will cause can.c.Close to not be run if io.Copy returns an error. Why is it better to do that than ignore the error from io.Copy and always run can.c.Close()?

conn_go18.go
tamird: Lovely, thanks for pointing that out; fixed by having it return an error (and panicking in the existing callers).

conn_go18.go
tamird: Good point, fixed.

conn_go18.go
```diff
@@ -49,41 +51,57 @@ func (cn *conn) BeginTx(ctx context.Context, opts driver.TxOptions) (driver.Tx,
...
-	w.int32(cn.processID)
-	w.int32(cn.secretKey)
+		w := can.writeBuf(0)
+		w.int32(80877102) // cancel request code
+		w.int32(cn.processID)
+		w.int32(cn.secretKey)
 
-	can.sendStartupPacket(w)
-	_ = can.c.Close()
+		if err := can.sendStartupPacket(w); err != nil {
```
mjibson: I don't see how the panics are caught here. `cancel()` is only invoked in a go routine, which means a defer'd recover in another go routine won't see it.

conn_go18.go
```diff
@@ -49,41 +51,57 @@ func (cn *conn) BeginTx(ctx context.Context, opts driver.TxOptions) (driver.Tx,
...
+		}
 	}
 	return nil
 }
 
-func (cn *conn) cancel() {
-	var err error
-	can := &conn{}
-	can.c, err = dial(cn.dialer, cn.opts)
+func (cn *conn) cancel() error {
```
mjibson: Why are we returning errors here at all? They are never used. In case we want them in the future and it's just good practice?

conn_go18.go
tamird: sendStartupPacket no longer panics, though.

conn_go18.go
tamird: Yes, that was my thinking.

conn_go18.go
mjibson: Ah, I didn't see that change cuz no reviewable.

tamird: Yeah, GH reviews are terrible. Pro tip though: you can just navigate to https://reviewable.io/reviews/lib/pq/578.

PR #585, opened 2017-03-20
Rename drv to Driver, so that the driver can be wrapped

luna-duclos:
This will allow packages like https://github.com/ExpansiveWorlds/instrumentedsql to wrap lib/pq and add instrumentation (logging, metrics, tracing, etc.)


luna-duclos:
Are there any remaining holdups to get this merged ?

mjibson:
I was hoping to have someone else voice an opinion, but I think it's safe.

PR #595, opened 2017-03-31
handlePgpass: warn about incorrect .pgpass permissions as psql does

martinatsynqfm:
this implements the comment

i got confused today when trying to connect my program to a password-protected database. manually copying the password into the `Connect` string worked, but storing it in `.pgpass` did not. when i checked with psql, i found my error, of course

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -177,7 +177,8 @@ func (c *conn) handlePgpass(o values) {
 	}
 	mode := fileinfo.Mode()
 	if mode&(0x77) != 0 {
-		// XXX should warn about incorrect .pgpass permissions as psql does
+		fmt.Fprintf(os.Stderr, "WARNING: password file %q has group or world access; "+
+			"permissions should be u=rw (0600) or less\n", filename)
 		return
 	}
 	file, err := os.Open(filename)


martinatsynqfm:
checks failed

https://travis-ci.org/lib/pq/jobs/217185835 looks like an intermittent issue

i don't understand https://travis-ci.org/lib/pq/jobs/217185858

johto:
I don't think it's kosher for a library to write to stderr like this.

If we want warnings there should probably be a separate interface for that (but the sql.DB interface makes that a bit cumbersome, here, again).

martinatsynqfm:
i agree that it's not nice for a library to log warning like this. i couldn't find any mechanism that was already in use for warnings, though

do you think this driver could either
- receive a callback function or a `Writer` or a `Logger` for sending warnings back, or
- offer a way to let library users specify that they want hard errors if pgpass can't be used

martinatsynqfm:
then again, this is exactly what the official libpq does:
```
$ cat main.c
#include <err.h>
#include <stdlib.h>

#include <postgresql/libpq-fe.h>

int main() {
  PGconn *pg;
  pg = PQconnectdb("dbname=postgres");

  if (PQstatus(pg) != CONNECTION_OK)
    errx(EXIT_FAILURE, "PostgreSQL connection failed: %s", PQerrorMessage(pg));

  fprintf(stderr, "success!\n");

  return EXIT_SUCCESS;
}
$ gcc main.c -lpq
$ chmod 640 ~/.pgpass
$ ./a.out
WARNING: password file "/home/mastensg/.pgpass" has group or world access; permissions should be u=rw (0600) or less
success!
$
```

matrosov-nikita:
@martinatsynqfm could you please take a look at this [issue](https://github.com/lib/pq/issues/940)? Approach with PGPASSFILE on windows doesn't work at all because of incorrect permissions check mentioned in this PR.

PR #598, opened 2017-04-10
Close conn if error occured(timeout error) during the startup.

c77cc:


arp242:
Looks like this was fixed in #698

PR #604, opened 2017-04-20
support 1.8  driver.RowsColumnTypeDatabaseTypeName interface

it512:
add driver.RowsColumnTypeDatabaseTypeName interface support 
ColumnType can return currect DatabaseTypeName


buf.go
```diff
@@ -4,7 +4,7 @@ import (
 	"bytes"
 	"encoding/binary"
 
-	"github.com/lib/pq/oid"
+	"github.com/it512/pq/oid"
```
stevenroose: I don't think this is meant to happen.

felixge:
@stevenroose @it512 I'm very interested in this feature as well, and willing to help implementing it.

I think the biggest problem with this PR is that it doesn't handle user defined types. IMO those are pretty important in postgres, so I'd love to get good support for them. Unfortunately the types will depend on the individual DB, so it's impossible to generate a static lookup table of them like this PR does.

My proposal would be to dynamically get the lookup table by querying `SELECT oid, typname FROM pg_type;` at runtime. The main question is when this should happen (when opening the db? or for each new conn) and if there could/should be a mechanism to refresh the types.

Maybe I'm also missing something, and there is an easy way to get the latest type names without the additional query.

Anyway, let me know what you think!

Cheers
Felix

sirwart:
FWIW, it512's implementation works great for my use case and I've been using his branch. I'd be happy to creating a new pull-request with cleaned up imports if that helps.

PR #605, opened 2017-04-28
committed failing encode/decode test

samuelkaufman:


johto:
This PR really needs some background information.  Why is this a problem?

samuelkaufman:
@johto whoops sorry was in progress writing up an issue with the reasoning behind the test case.

tamird:
Why don't you include the issue in this PR description (and commit message)? You're making it somewhat more difficult than it needs to be by making it necessary to chase down the issue link (particularly for a future source code reader).

samuelkaufman:
@tamird Hey, yeah I had written the pr first, then wrote up the issue, then edited the description of the PR to include the issue number (#606). I didn't go through all the formalities at first because I about 90% expected I was just doing it wrong and it wasn't an issue with lib/pq, so sorry for the sloppiness.

To clarify here as well:

## Description

Parsing the output of FormatTimestamp with ParseTimestamp is not possible.

## Expected

Based on the [documentation][1]:
> FormatTimestamp formats t into Postgres' text format for timestamps.
> ParseTimestamp parses Postgres' text format.

You would expect Parse to handle the output of Format

## Actual

It errors during parse.

I've also included a test that checks that parsed timestamp equals the input timestamp, however I am not sure that is possible and/or right, depending on timezones, rounding, etc.


[1]: https://godoc.org/github.com/lib/pq#FormatTimestamp

tamird:
Thanks, but I would prefer if you could include the information in the commit message. It can be pretty frustrating, when reading the output of `git blame` to be directed to a Github issue, requiring an additional context switch.

samuelkaufman:
@tamird got it, I just amended the commit message to have all the info in it as well.

arp242:
This was fixed in #1073, which includes a similar test.

PR #607, opened 2017-05-02
Add support for tcp keep alives

kvigen:


kingcr7:
why cancel it? how to set keep alive duration？i can not set keep alive param in client when connet to pg.

PR #612, opened 2017-05-20
Implement basic support for numeric range types

Ragnis:
- The `Int32Range` and `Int64Range` can be used for reading/writing `int4range` and `int8range` values, respectively
- The `Float64Range` is useful for reading/writing `numrange` values, when possible loss of precision is not of concern.

I'm not yet sure what would be a good way of handling the `numrange` type. Maybe by treating the min/max values as strings?

### DIFF
--- a/ranges/float64range.go
+++ b/ranges/float64range.go
@@ -22,7 +22,7 @@ func (r *Float64Range) Scan(val interface{}) error {
 		r.MaxInclusive = false
 		return nil
 	}
-	minIncl, maxIncl, min, max, err := readFloatRange(val.([]byte))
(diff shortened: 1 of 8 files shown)


buyology:
nice! would be great to also add support for `tsrange` and `tstzrange` while at it? 

ping if you want me to take a stab at that

UmanShahzad:
Can we look into merging this at some point?

cliedeman:
was also implementing my own until I found this

aszecowka:
What it the status of this PR? I am interested in strange support, does this library support that?

arp242:
I'm closing this in favour of #1196, which seems a better solution using type parameters.

PR #618, opened 2017-05-30
Avoid goroutine leak on error in query with context

### LINKED ISSUE #617: Goroutine leak on context queries
Revision: `2704adc878c21e1329f46f6e56a1c387d788ff94`

Bug: after a long time I notice that goroutine number in my application goes up. The goroutine profile shows:

    35886 @ 0x42c76a 0x42c84e 0x403dcd 0x403add 0x72819e 0x458d51
    #	0x72819d	github.com/lib/pq.(*conn).watchCancel.func1+0xcd	.../vendor/src/github.com/lib/pq/conn_go18.go:65

It's this code:
```
select {
case <-done:
	_ = cn.cancel()
	finished <- struct{}{} // <<< stuck here
case <-finished:
}
```

Some investigation leads me to believe that this is the reason:
```
// Implement the "QueryerContext" interface
func (cn *conn) QueryContext(ctx context.Context, query string, args []driver.NamedValue) (driver.Rows, error) {
	list := make([]driver.Value, len(args))
	for i, nv := range args {
		list[i] = nv.Value
	}
	finish := cn.watchCancel(ctx)
	r, err := cn.query(query, list)
	if err != nil {
		return nil, err
	}
	r.finish = finish
	return r, nil
}
```

If `cn.query` returns an error, the `finish` function is never called. I believe, it should read
```
	if err != nil {
		if finish != nil {
			finish()
		}
		return nil, err
	}
```

ainar-g:
Fixes #617. Finish watching the context in case of error, thus preventing a goroutine leak on every failed QueryContext call.

### DIFF
--- a/go18_test.go
+++ b/go18_test.go
@@ -5,6 +5,7 @@ package pq
 import (
 	"context"
 	"database/sql"
+	"runtime"
 	"testing"
 	"time"
 )
@@ -152,6 +153,36 @@ func TestContextCancelQuery(t *testing.T) {
 		} else if err := rows.Close(); err != nil {
 			t.Fatal(err)
 		}
+	}
+}
+
+// TestIssue617 tests that a failed query in QueryContext doesn't lead to a
+// goroutine leak.
+func TestIssue617(t *testing.T) {
+	db := openTestConn(t)
+	defer db.Close()
+
+	const N = 10
+
+	numGoroutineStart := runtime.NumGoroutine()
+	for i := 0; i < N; i++ {
+		func() {
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+			_, err := db.QueryContext(ctx, `SELECT * FROM DOESNOTEXIST`)
+			pqErr, _ := err.(*Error)
+			// Expecting "pq: relation \"doesnotexist\" does not exist" error.
+			if err == nil || pqErr == nil || pqErr.Code != "42P01" {
+				t.Fatalf("expected undefined table error, got %v", err)
+			}
+		}()
+	}
+	numGoroutineFinish := runtime.NumGoroutine()
+
+	// We use N/2 and not N because the GC and other actors may increase or
+	// decrease the number of goroutines.
+	if numGoroutineFinish-numGoroutineStart >= N/2 {
+		t.Errorf("goroutine leak detected, was %d, now %d", numGoroutineStart, numGoroutineFinish)
 	}
 }
 
--- a/conn_go18.go
+++ b/conn_go18.go
@@ -19,6 +19,9 @@ func (cn *conn) QueryContext(ctx context.Context, query string, args []driver.Na
 	finish := cn.watchCancel(ctx)
 	r, err := cn.query(query, list)
 	if err != nil {
+		if finish != nil {
+			finish()
+		}
 		return nil, err
 	}
 	r.finish = finish


ainar-g:
@tamird Done.

go18_test.go
```diff
@@ -155,6 +156,32 @@ func TestContextCancelQuery(t *testing.T) {
...
+func TestIssue617(t *testing.T) {
+	db := openTestConn(t)
+	defer db.Close()
+
+	const N = 10
+
+	numGoroutineStart := runtime.NumGoroutine()
+	for i := 0; i < N; i++ {
+		query := `SELECT * FROM DOESNOTEXIST`
+		ctx, _ := context.WithTimeout(context.Background(), time.Second)
```
tamird: heh, this also leaks a goroutine. I understand that what you want here is a cancellable context, so please write this as:

```
func() {
  ctx, cancel := context.WithCancel(context.Background())
  defer cancel()
  rows, err := db.QueryContext(ctx, `SELECT * FROM DOESNOTEXIST`)
  ....
}()

File: go18_test.go [hunk 2]
```diff
@@ -155,6 +156,32 @@ func TestContextCancelQuery(t *testing.T) {
...
+// goroutine leak.
+func TestIssue617(t *testing.T) {
+	db := openTestConn(t)
+	defer db.Close()
+
+	const N = 10
+
+	numGoroutineStart := runtime.NumGoroutine()
+	for i := 0; i < N; i++ {
+		query := `SELECT * FROM DOESNOTEXIST`
```
tamird: no need for this local, just inline it (see suggestion below)

go18_test.go
```diff
@@ -155,6 +156,32 @@ func TestContextCancelQuery(t *testing.T) {
...
+		ctx, _ := context.WithTimeout(context.Background(), time.Second)
+		rows, err := db.QueryContext(ctx, query)
+		if rows != nil || err == nil {
```
tamird: this is simultaneously testing too much and too little. There's no need to check that rows is nil - that's database/sql's API surface, not ours. However, we _should_ confirm that the error we get is the one we expect - it's not enough to check that the error is non-nil.

go18_test.go
```diff
@@ -155,6 +156,32 @@ func TestContextCancelQuery(t *testing.T) {
...
+			t.Fatal("expected nil rows and a non-nil error, got rows %v and error %v", rows, err)
+		}
+	}
+	numGoroutineFinish := runtime.NumGoroutine()
+
+	// We use N/2 and not N because the GC and other actors may increase or
```
tamird: If you address my comment about leaking context goroutines, I think you'll be able to remove this (but please correct me if I'm wrong).

go18_test.go
ainar-g: `WithTimeout` doesn't seem to leak goroutines: https://play.golang.org/p/76pOH9R44j.

But I'll do it your way anyway.

go18_test.go
ainar-g: Nope, the number of goroutines still fluctuates from launch to launch.

go18_test.go
ainar-g: Done.

go18_test.go
ainar-g: Done.

ainar-g:
@tamird Done.

go18_test.go
```diff
@@ -155,6 +156,36 @@ func TestContextCancelQuery(t *testing.T) {
...
+func TestIssue617(t *testing.T) {
+	db := openTestConn(t)
+	defer db.Close()
+
+	const N = 10
+
+	numGoroutineStart := runtime.NumGoroutine()
+	for i := 0; i < N; i++ {
+		func() {
+			ctx, cancel := context.WithTimeout(context.Background(), time.Second)
```
tamird: WithCancel, not WithTimeout

ainar-g: :man_facepalming: Sorry about that. Done.

tamird:
Thanks! Will merge on green.

tamird:
Looks like there's some flakiness in Travis, but I'm going to go ahead and merge. Thank you @ainar-g!

ainar-g:
@tamird You're welcome! It has been nice working with you.

PR #619, opened 2017-05-31
Add support for sql.TxOptions

thatguystone:


go18_test.go
```diff
@@ -239,3 +239,79 @@ func TestContextCancelBegin(t *testing.T) {
...
+			}
+
+			var isRO string
+			err = tx.QueryRow("select current_setting('transaction_read_only')").Scan(&isRO)
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			if ro != (isRO == "on") {
+				t.Errorf("read/[write,only] not set: %t != %t for level %s",
```
mjibson: The second `%t` is used with isRO, which is a string, not a bool.

thatguystone: Fixed. Good catch.

conn_go18.go
```diff
@@ -44,13 +45,30 @@ func (cn *conn) ExecContext(ctx context.Context, query string, args []driver.Nam
...
 func (cn *conn) BeginTx(ctx context.Context, opts driver.TxOptions) (driver.Tx, error) {
-	if opts.Isolation != 0 {
-		return nil, errors.New("isolation levels not supported")
+	mode := ""
+
+	switch sql.IsolationLevel(opts.Isolation) {
+	case sql.LevelDefault:
+		// Don't touch mode: use the server's default
+	case sql.LevelReadUncommitted:
+		mode += " ISOLATION LEVEL READ UNCOMMITTED"
```
tamird: there's no need to do concatenation here (which forces an allocation); you can just `var mode string` and assign  it here.

conn_go18.go
```diff
@@ -44,13 +45,30 @@ func (cn *conn) ExecContext(ctx context.Context, query string, args []driver.Nam
...
+		mode += " ISOLATION LEVEL SERIALIZABLE"
+	default:
+		return nil, fmt.Errorf("isolation level not supported: %d", opts.Isolation)
 	}
+
 	if opts.ReadOnly {
-		return nil, errors.New("read-only transactions not supported")
+		mode += " READ ONLY"
+	} else {
+		mode += " READ WRITE"
```
tamird: you can omit this as well, since it's the default.

go18_test.go
```diff
@@ -239,3 +239,79 @@ func TestContextCancelBegin(t *testing.T) {
...
+
+			tx.Rollback()
+		}
+	}
+
+	_, err := db.BeginTx(ctx, &sql.TxOptions{
+		Isolation: sql.LevelLinearizable,
+	})
+	if err == nil {
+		t.Errorf("expected LevelLinearizable to fail")
```
tamird: please check for the exact error you expect; it's not enough to check for a non-nil error.

go18_test.go
```diff
@@ -239,3 +239,79 @@ func TestContextCancelBegin(t *testing.T) {
...
+			isolation: "repeatable read",
+		},
+		{
+			level:     sql.LevelSerializable,
+			isolation: "serializable",
+		},
+	}
+
+	for _, test := range tests {
+		for i := 0; i < 2; i++ {
```
tamird: nit: `for _, ro := range []bool{true, false} {`?

conn_go18.go
thatguystone: This can be changed via server config (`default_transaction_read_only`) and set per-session, so it's probably best if the library makes no assumptions about defaults.

tamird: TIL, thanks.

conn_go18.go
thatguystone: Yeah, it's a bit cleaner without the concat.

As a note: go doesn't perform an alloc when concating with an empty string, see: https://play.golang.org/p/_Xp6RjpYZW

tamird: Neat, thanks for the pointer.

thatguystone:
@tamird @mjibson Can this be merged?

mjibson:
The failing tests need to be investigated first. They are occurring in the master branch of go, so it is possibly a fluke, but we need to know why they are failing before we can merge.

thatguystone:
The same tests are also failing in pq/master. I've pushed a fix.

PR #624, opened 2017-06-09
Implement support ColumnConverter  interface

kshvakov:
Hi, this is implementation of [ValueConverter interface](https://golang.org/pkg/database/sql/driver/#ValueConverter) for providing the ConvertValue method

### DIFF
--- a/conn_go18.go
+++ b/conn_go18.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"reflect"
 )
 
 // Implement the "QueryerContext" interface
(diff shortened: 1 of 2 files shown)


conn_go18.go
```diff
@@ -126,3 +127,63 @@ func (cn *conn) cancel() error {
 		return err
 	}
 }
+
+var _converter = &converter{}
+
+func (st *stmt) ColumnConverter(idx int) driver.ValueConverter {
+	return _converter
```
ainar-g: Why not just return `&converter{}`?

conn_go18.go
```diff
@@ -126,3 +127,63 @@ func (cn *conn) cancel() error {
 		return err
 	}
 }
+
+var _converter = &converter{}
```
ainar-g: Golint will probably error here. Why not call it `defaultConverter`? This would be clearer.

kshvakov: Yes, you are right, I changed this variable name to defaultConverter

conn_go18.go
kshvakov: Only for performance

conn_go18.go
```diff
@@ -126,3 +127,63 @@ func (cn *conn) cancel() error {
...
+	case uint:
+		return int64(value), nil
+	case uint8:
+		return int64(value), nil
+	case uint16:
+		return int64(value), nil
+	case uint32:
+		return int64(value), nil
+	case uint64:
+		if value >= 1<<63 {
```
ainar-g: Sorry, I've only now noticed, but you could use `value > math.MaxInt64` here.

kshvakov: Done

kshvakov:
Hello, branch is rebased

arp242:
This was added in #1125 and #1238 with CheckNamedValue(), which does the same as the (now-deprecated) ColumnConverter().

Issue #622, opened 2017-06-07
Support Go 1.8 ColumnTypes


johanbrandhorst:
Is this on the roadmap? It appears currently unsupported. It would require implementing the following interfaces on `rows`:

https://golang.org/pkg/database/sql/driver/#RowsColumnTypeDatabaseTypeName
https://golang.org/pkg/database/sql/driver/#RowsColumnTypeLength
https://golang.org/pkg/database/sql/driver/#RowsColumnTypeNullable
https://golang.org/pkg/database/sql/driver/#RowsColumnTypePrecisionScale
https://golang.org/pkg/database/sql/driver/#RowsColumnTypeScanType

ainar-g:
That's odd. [This](https://play.golang.org/p/P1GSVsf7yI) code, which uses `ScanType` works for me already.

johanbrandhorst:
`database/sql` returns `interface{}` as the `ScanType` by default, so it'll work, but it won't be very good.

arp242:
Looks like this was added in #647.

PR #628, opened 2017-06-27
Implements RowsColumnTypeScanType

wingyplus:
Hi, I just implements https://golang.org/pkg/database/sql/driver/#RowsColumnTypeScanType by map from postgres oid (in oid package). If someone have any suggestions, please comments or review this code.

Ref #622


arp242:
Looks like this was added in #647.

PR #634, opened 2017-07-06
correctly close rows objects if there are multiple result sets

mjibson:


conn_test.go
```diff
@@ -1583,3 +1583,25 @@ func TestRowsResultTag(t *testing.T) {
...
+	rows, err := db.Query("SELECT 1; SELECT 2;")
+	if err != nil {
+		t.Fatal(err)
+	}
+	if err := rows.Close(); err != nil {
+		t.Fatal(err)
+	}
+
+	var id int
+	if err := db.QueryRow("SELECT 3").Scan(&id); err != nil {
```
cbandy: Should these calls be on a `Tx` to definitely use the same connection?

mjibson: Yes. Done.

tamird: I think you missed this one, which is still a call on `db`.

mjibson: ergh. yes.

PR #640, opened 2017-07-24
Add support for NamedValueChecker interface

cyberdelia:
Add support for the new [NamedValueChecker](https://tip.golang.org/pkg/database/sql/driver/#NamedValueChecker) interface available in go 1.9.

This will allow to pass an array or a slice directly as an argument.

### DIFF
--- a/conn_go19.go
+++ b/conn_go19.go
@@ -1,0 +1,27 @@
+// +build go1.9
+
+package pq
+
+import (
+	"database/sql/driver"
+	"reflect"
+)
+
+func (c *conn) CheckNamedValue(nv *driver.NamedValue) (err error) {
+	if _, ok := nv.Value.(driver.Valuer); ok {
+		// Ignore Valuer, for backward compatiblity with pq.Array()
+		return driver.ErrSkip
+	}
+
+	// Ignoring []byte / []uint8
+	if _, ok := nv.Value.([]uint8); ok {
+		return driver.ErrSkip
+	}
+
+	if k := reflect.ValueOf(nv.Value).Kind(); k == reflect.Array || k == reflect.Slice {
+		nv.Value, err = Array(nv.Value).Value()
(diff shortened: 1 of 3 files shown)


conn_go19.go
```diff
@@ -0,0 +1,28 @@
...
+		return driver.ErrSkip
+	}
+
+	v := reflect.ValueOf(nv.Value)
+	switch v.Kind() {
+	case reflect.Array, reflect.Slice:
+		// Ignoring []byte / []uint8
+		if v.Type().Elem().Kind() == reflect.Uint8 {
+			return driver.ErrSkip
+		}
```
ainar-g: Any reason not to do
```go
if _, ok := nv.Value.([]unit8); ok {
    return driver.ErrSkip
}
```
above? That would spare an indentation level, simplify the condition, and would probably perform better (although it isn't important here). Besides, then you could replace `switch` with a simple
```go
if k := v.Kind(); k == reflect.Array || k == reflect.Slice {
    nv.Value, err = Array(nv.Value).Value()
    return err
}
```

conn_go19_test.go
```diff
@@ -0,0 +1,37 @@
...
+		t.Fatal("expected row")
+	}
+
+	var i []int64
+	if err := r.Scan(Array(&i)); err != nil {
+		t.Fatal(err)
+	}
+
+	if !reflect.DeepEqual(i, expected) {
+		t.Errorf("expect %v, got %v", expected, i)
```
ainar-g: `s/expect/expected/`? I personally prefer `exp`, since it is three letters, like `got`, but that's nitpicking.

cyberdelia:
@ainar-g Good points, made the appropriate changes.

ainar-g:
@cyberdelia Thanks. I hope @tamird and @mjibson will approve it.

PR #641, opened 2017-08-04
adding failed array ilike test

ptravis:
db.Exec("select * from temp where 'a' ilike any($1::text[])", Array([]string{"\\"}))

fails with

        conn_test.go:1637: pq: LIKE pattern must not end with escape character


arp242:
This does not seem like a bug in pq, but just what PostgreSQL returns. You can use `escape '$'` to set a different escape character in a quick check.

PR #643, opened 2017-08-06
Use go generate to generate oid types information

cyberdelia:
Replace ``gen.go`` by the installable ``oid`` command that can be run by ``go generate``.
Running ``go generate ./...`` will now re-create the ``types.go`` file.

This will also allow to generate ``types.go`` even if ``types.go`` is missing, as long as you have the ``oid`` binary installed.


tamird:
I understand moving this tool to its own package, but why require that it be installed? Seems like you could keep leave the `go generate` line as `go run ...`.

cyberdelia:
The installation step allow to avoid the cycle import of ``lib/pq`` itself when running the generation, so you can always generate the file even if the file is missing (as long as you have the old binary around). 

It was fairly useful to have when making changes to it, but I'm not too attached to not using to ``go run``.

cyberdelia:
@tamird @mjibson You seem to be the most recent maintainer on this repo, any chance to see some movement on this? I'll also gladly give you a hand for other maintenance tasks.

tamird:
I appreciate what you're trying to achieve, here, but I'm not sure I agree with the method. Ideally, we could extract a package or tool that can generate types.go without depending on knowledge of OIDs.

I have to think about this more than I have to know if that's possible or not, and this isn't super high on my priorities, so I probably won't get to that any time soon. If you'd like to do that exploration, I'd be more than happy to review it.

cyberdelia:
That could be explored for sure, but I don't think this zero sum change should be block for this.

I could use ``go run`` if you feel it should be kept, but this PR provides zero changes to the library except using the widely used and built-in go tool ``go generate`` and some small improvement to the generation code (while keeping the same output).

tamird:
> but this PR provides zero changes to the library except using the widely used and built-in go tool `go generate` and some small improvement to the generation code (while keeping the same output).

We're already using `go generate`, so that is not new. What is new is requiring that the `oid` binary be installed, and that's not obviously better. Having random executables of unknown versions lying around results in surprises (for developers of this library, in this case).

PR #646, opened 2017-08-08
Fix some linting issues

cyberdelia:
Fixed some of the "issues" raised by the [golint](https://github.com/golang/lint) tool on this package.

### DIFF
--- a/encode_test.go
+++ b/encode_test.go
@@ -370,17 +370,17 @@ func TestInfinityTimestamp(t *testing.T) {
 		t.Errorf("Scanning -infinity, expected time %q, got %q", y1500, resultT.String())
 	}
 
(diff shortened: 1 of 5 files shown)


encode_test.go
```diff
@@ -370,17 +370,17 @@ func TestInfinityTimestamp(t *testing.T) {
 		t.Errorf("Scanning -infinity, expected time %q, got %q", y1500, resultT.String())
 	}
 
-	y_1500 := time.Date(-1500, time.January, 1, 0, 0, 0, 0, time.UTC)
+	ym1500 := time.Date(-1500, time.January, 1, 0, 0, 0, 0, time.UTC)
```
cbandy: This name isn't clear to me, but I can't think of anything better.

cyberdelia: I was thinking ``m`` as is ``minus``.

PR #647, opened 2017-08-10
Add ColumnType support

cyberdelia:
This allow this driver to return valid information for [``ColumnType``](https://golang.org/pkg/database/sql/#ColumnType), at the exception of ``Nullable`` which, as far as I know, is not determinable with the data available in a [``RowDescription``](https://www.postgresql.org/docs/9.6/static/protocol-message-formats.html).

Related to #628 and #604.


tamird:
Looks like there's a missing type - the new tests are failing on Go 1.8 (which is the only time they're enabled, it would seem).

cyberdelia:
@tamird I changed the type name last minute, I fixed the last reference.

tamird:
Reviewed 3 of 5 files at r1.
Review status: 3 of 5 files reviewed at latest revision, 7 unresolved discussions.

---

*[rows.go, line 26 at r1](https://reviewable.io:443/reviews/lib/pq/647#-KrCc9c-7sarIy_0f13Q:-KrCc9c-7sarIy_0f13R:bks9dcw) ([raw file](https://github.com/lib/pq/blob/1dc07bfcfaee861523593b55ff045ba9e168f8bf/rows.go#L26)):*
> ```Go
> func (fd fieldDesc) Type() reflect.Type {
> 	switch fd.OID {
> 	case oid.T_int8:
> ```

i'm surprised this mapping of oid to go type doesn't already exist. doesn't it?

---

*[rows.go, line 41 at r1](https://reviewable.io:443/reviews/lib/pq/647#-KrCcG3AHZo9z6bMyCPB:-KrCcG3AHZo9z6bMyCPC:bv70ltc) ([raw file](https://github.com/lib/pq/blob/1dc07bfcfaee861523593b55ff045ba9e168f8bf/rows.go#L41)):*
> ```Go
> 		return reflect.TypeOf(time.Time{})
> 	case oid.T_bytea:
> 		return reflect.TypeOf([]byte{})
> ```

nit: `[]byte(nil)`

---

*[rows.go, line 43 at r1](https://reviewable.io:443/reviews/lib/pq/647#-KrCcKomCym0DYesae8-:-KrCcKomCym0DYesae80:b-7u0jk8) ([raw file](https://github.com/lib/pq/blob/1dc07bfcfaee861523593b55ff045ba9e168f8bf/rows.go#L43)):*
> ```Go
> 		return reflect.TypeOf([]byte{})
> 	default:
> 		return reflect.TypeOf(new(interface{})).Elem()
> ```

`reflect.TypeOf(interface{}(nil))`?

---

*[rows_test.go, line 1 at r1](https://reviewable.io:443/reviews/lib/pq/647#-KrCbtIb5s7DYvNxgYtC:-KrCbtIb5s7DYvNxgYtD:b-j5utk8) ([raw file](https://github.com/lib/pq/blob/1dc07bfcfaee861523593b55ff045ba9e168f8bf/rows_test.go#L1)):*
> ```Go
> // +build go1.8
> ```

why is this behind a build tag? did you mean to test the new database/sql functionality here?

---

*[rows_test.go, line 105 at r1](https://reviewable.io:443/reviews/lib/pq/647#-KrCbfPAPKSjJWzl0ucM:-KrCbfPAPKSjJWzl0ucN:b-jsj1t0) ([raw file](https://github.com/lib/pq/blob/1dc07bfcfaee861523593b55ff045ba9e168f8bf/rows_test.go#L105)):*
> ```Go
> 	for i, tt := range tts {
> 		dt := fieldDesc{OID: tt.typ, Mod: tt.mod}
> 		if p, s, k := dt.PrecisionScale(); k != tt.ok || p != tt.precision || s != tt.scale {
> ```

it would be good to separate these assertions for readability:
```
p, s, k := ...
if p != .. {
..
}
if s != .. {
...
}
...
```

---

*[oid/gen.go, line 18 at r1](https://reviewable.io:443/reviews/lib/pq/647#-KrCbO-AEYKQ0LvTWcdC:-KrCbO-AEYKQ0LvTWcdD:bn6setw) ([raw file](https://github.com/lib/pq/blob/1dc07bfcfaee861523593b55ff045ba9e168f8bf/oid/gen.go#L18)):*
> ```Go
> )
> 
> // OID represent a postgres Object Identifier Type.
> ```

adding this type doesn't seem worth it to me, especially since there's already an Oid type elsewhere in this package. otherwise changes to this file look good to me.

---

*[oid/gen.go, line 79 at r1](https://reviewable.io:443/reviews/lib/pq/647#-KrCacTrmVtPICmAy49G:-KrCacTrmVtPICmAy49H:bbyo81p) ([raw file](https://github.com/lib/pq/blob/1dc07bfcfaee861523593b55ff045ba9e168f8bf/oid/gen.go#L79)):*
> ```Go
> 		log.Fatal(err)
> 	}
> 	fmt.Fprintln(w, "// generated by 'go run gen.go'; do not edit")
> ```

While you're here, can you make this follow the golang convention for generated code? https://github.com/golang/go/issues/13560#issuecomment-277804473

that would be `fmt.Println(w, "// Code generated by gen.go. DO NOT EDIT.")`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/lib/pq/647)*

cyberdelia:
@tamird I tried to use Reviewable, but I'm uncomfortable with allowing it to access all my orgs public repos just to get one PR reviewed.

I addressed most of the feedback at a few exception, I'll try to answer the rest inline.

rows.go
```diff
@@ -0,0 +1,95 @@
...
+	// Note that negative values denote variable-width types.
+	Len int
+	// The type modifier (see pg_attribute.atttypmod).
+	// The meaning of the modifier is type-specific.
+	Mod int
+}
+
+func (fd fieldDesc) Type() reflect.Type {
+	switch fd.OID {
+	case oid.T_int8:
```
cyberdelia: I haven't found a mapping of OID to type elsewhere, but I'll be glad to use it if someone point to an existing one.

rows.go
```diff
@@ -0,0 +1,95 @@
...
+	case oid.T_bool:
+		return reflect.TypeOf(false)
+	case oid.T_numeric:
+		return reflect.TypeOf(float64(0))
+	case oid.T_date, oid.T_time, oid.T_timetz, oid.T_timestamp, oid.T_timestamptz:
+		return reflect.TypeOf(time.Time{})
+	case oid.T_bytea:
+		return reflect.TypeOf([]byte(nil))
+	default:
+		return reflect.TypeOf(new(interface{})).Elem()
```
cyberdelia:
```go
reflect.TypeOf(interface{}(nil))
```

This is invalid go, but what did you have in mind?

oid/gen.go
```diff
@@ -10,10 +10,22 @@ import (
 	"log"
 	"os"
 	"os/exec"
+	"strings"
 
 	_ "github.com/lib/pq"
 )
 
+// OID represent a postgres Object Identifier Type.
+type OID struct {
```
cyberdelia: I could un-export it or rename it, but it helps with only using one query to generate the different structure we need in ``types.go``.

tamird:
Thanks. Did you mean to add a test of http://godoc.org/database/sql#Rows.ColumnTypes?

---

Reviewed 2 of 4 files at r2.
Review status: 3 of 5 files reviewed at latest revision, 4 unresolved discussions.

---

*[rows.go, line 43 at r2](https://reviewable.io:443/reviews/lib/pq/647#-KrCa4yLl4lnqsDInSfF-r2-43:-KrDAtXHOMr7sbPgo1cC:bprgjg2) ([raw file](https://github.com/lib/pq/blob/8650f17f5277888c1bc51710d544513ba170b76a/rows.go#L43)):*
<details><summary><i>Previously, cyberdelia (Timothée Peignier) wrote…</i></summary><blockquote>

```go
reflect.TypeOf(interface{}(nil))
```

This is invalid go, but what did you have in mind?
</blockquote></details>

Well it's valid, it just doesn't do the right thing. I guess this is fine as-is.

---

*[rows_test.go, line 105 at r2](https://reviewable.io:443/reviews/lib/pq/647#-KrD90-lQhsK6HVB8EIB:-KrD90-lQhsK6HVB8EIC:b1v4oic) ([raw file](https://github.com/lib/pq/blob/8650f17f5277888c1bc51710d544513ba170b76a/rows_test.go#L105)):*
> ```Go
> 		p, s, k := dt.PrecisionScale()
> 		if k != tt.ok {
> 			t.Errorf("(%d)  got: %t want: %t", i, k, tt.ok)
> ```

double space here

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/lib/pq/647#-:-KrDBCc6SxS9b2ZY0Akh:b1g5wse)*

cyberdelia:
I don't think testing ``Rows.ColumnTypes`` provides much here, it just does a interface check and call the said interface methods: https://golang.org/src/database/sql/sql.go#L2304

tamird:
Understood, it's the only thing we care about here - nobody is calling
these new methods directly, so if we're not testing the interface our users
will use, what are we doing?

cyberdelia:
@tamird I added a test.

tamird:
Looks good modulo some comments, the biggest one being that testing `fieldDesc` directly seems useless to me; I would just remove all those tests and the methods on `fieldDesc` can all be inlined into the relevant methods on `*rows`.

---

Reviewed 1 of 4 files at r2.
Review status: 4 of 5 files reviewed at latest revision, 7 unresolved discussions.

---

*[rows.go, line 24 at r3](https://reviewable.io:443/reviews/lib/pq/647#-KrXRqqiWXn3PcOodYlm:-KrXRqqiWXn3PcOodYln:b61g047) ([raw file](https://github.com/lib/pq/blob/ddceb65e4099b8f418b5a380bbc1bfa0d28507fc/rows.go#L24)):*
> ```Go
> }
> 
> func (fd fieldDesc) Type() reflect.Type {
> ```

it's not clear to me why we need all these methods; their bodies could just go into the relevant methods on `*rows`.

---

*[rows_test.go, line 63 at r3](https://reviewable.io:443/reviews/lib/pq/647#-KrXRe5f-Cg3NekqMEeP:-KrXRe5f-Cg3NekqMEeQ:b-w9pal) ([raw file](https://github.com/lib/pq/blob/ddceb65e4099b8f418b5a380bbc1bfa0d28507fc/rows_test.go#L63)):*
> ```Go
> 	for i, tt := range tts {
> 		dt := fieldDesc{OID: tt.typ}
> 		if kind := dt.Type().Kind(); kind != tt.kind {
> ```

why do we care about the Kind? again, doesn't seem like a useful test (we should test this end to end by querying for these types from the database).

---

*[rows_test.go, line 69 at r3](https://reviewable.io:443/reviews/lib/pq/647#-KrXREcOxnsD7MwIBpEp:-KrXREcOxnsD7MwIBpEq:b-2d5ucj) ([raw file](https://github.com/lib/pq/blob/ddceb65e4099b8f418b5a380bbc1bfa0d28507fc/rows_test.go#L69)):*
> ```Go
> }
> 
> func TestDataTypeLength(t *testing.T) {
> ```

also not clear what this is testing; it doesn't seem valuable to test behaviour of manually constructed `fieldDesc`s, since it's possible to construct nonsense values.

---

*[rows_test.go, line 91 at r3](https://reviewable.io:443/reviews/lib/pq/647#-KrXQiG6uKwHSRM-PfnW:-KrXQiG6uKwHSRM-PfnX:b-wjfo40) ([raw file](https://github.com/lib/pq/blob/ddceb65e4099b8f418b5a380bbc1bfa0d28507fc/rows_test.go#L91)):*
> ```Go
> }
> 
> func TestDataTypePrecisionScale(t *testing.T) {
> ```

what is this testing? reading this I don't know why `precision=9, scale=2` is correct for `mod=589830`.

---

*[rows_test.go, line 118 at r3](https://reviewable.io:443/reviews/lib/pq/647#-KrXPlE2GwjpV5rU1knc:-KrXPlE3h4GSlne5bqHS:b-divalj) ([raw file](https://github.com/lib/pq/blob/ddceb65e4099b8f418b5a380bbc1bfa0d28507fc/rows_test.go#L118)):*
> ```Go
> }
> 
> var columnTypesTests = []struct {
> ```

pretty odd to define this outside the test case, since it's only applicable to the query in the test. also, this "table drive" style is very hard to read because of how far away from the query these definitions are.

this is also super verbose because of the nested structs (meaning that you have to repeat the definition in each case; perhaps it's worth naming the types).

the method `Nullable` is also not tested: http://godoc.org/database/sql#ColumnType.Nullable

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/lib/pq/647#-:-KrXRy0UBe8aKU8QInOg:b-cyod7q)*

rows_test.go
```diff
@@ -0,0 +1,241 @@
...
+			t.Errorf("(%d) wrong precision got: %d want: %d", i, p, tt.precision)
+		}
+		if s != tt.scale {
+			t.Errorf("(%d) wrong scale got: %d want: %d", i, s, tt.scale)
+		}
+	}
+}
+
+func TestRowsColumnTypes(t *testing.T) {
+	columnTypesTests := []struct {
```
cyberdelia: I moved them inside the test function. ``Nullable`` isn't test since it's not implemented.

rows_test.go
```diff
@@ -0,0 +1,241 @@
...
+
+	for i, tt := range tts {
+		dt := fieldDesc{OID: tt.typ, Len: tt.len, Mod: tt.mod}
+		if l, k := dt.Length(); k != tt.ok || l != tt.length {
+			t.Errorf("(%d) got: %d, %t want: %d, %t", i, l, k, tt.length, tt.ok)
+		}
+	}
+}
+
+func TestDataTypePrecisionScale(t *testing.T) {
```
cyberdelia: It test the code that essentially does: https://github.com/postgres/postgres/blob/master/src/backend/utils/adt/numeric.c#L962-L968

rows_test.go
```diff
@@ -0,0 +1,241 @@
...
+	for i, tt := range tts {
+		dt := fieldDesc{OID: tt.typ}
+		if kind := dt.Type().Kind(); kind != tt.kind {
+			t.Errorf("(%d) got: %s want: %s", i, kind, tt.kind)
+		}
+	}
+}
+
+func TestDataTypeLength(t *testing.T) {
+	tts := []struct {
```
cyberdelia: This test we are actually doing the correct math and logic for this different field.

rows_test.go
```diff
@@ -0,0 +1,241 @@
...
+		{oid.T_time, reflect.Struct},
+		{oid.T_timetz, reflect.Struct},
+		{oid.T_timestamp, reflect.Struct},
+		{oid.T_timestamptz, reflect.Struct},
+		{oid.T_bytea, reflect.Slice},
+	}
+
+	for i, tt := range tts {
+		dt := fieldDesc{OID: tt.typ}
+		if kind := dt.Type().Kind(); kind != tt.kind {
```
cyberdelia: We use ``Kind`` here, mostly because it's shorter to write and easier to compare.

rows.go
```diff
@@ -0,0 +1,95 @@
...
+	OID oid.Oid
+	// The data type size (see pg_type.typlen).
+	// Note that negative values denote variable-width types.
+	Len int
+	// The type modifier (see pg_attribute.atttypmod).
+	// The meaning of the modifier is type-specific.
+	Mod int
+}
+
+func (fd fieldDesc) Type() reflect.Type {
```
cyberdelia: I preferred to contain the logic around field type and size at the field level, and not the row level, it also makes it easier to test said logic.

cyberdelia:
@tamird I addressed some of your feedback inline.

torkelo:
hope to see this merged soon as it would enable Postres support in https://github.com/grafana/grafana (https://grafana.com)

tamird:
@mjibson care to have a look? I think this is probably fine.

mjibson:
Review status: 4 of 5 files reviewed at latest revision, 13 unresolved discussions.

---

*[rows.go, line 36 at r4](https://reviewable.io:443/reviews/lib/pq/647#-KslClX5e_TLDrw23J3x:-KslClX6GfUaa0Z1Zw1P:b-ygnkic) ([raw file](https://github.com/lib/pq/blob/f8eb458b02624cae56a61a3d4d3e97cb0e2bfd0c/rows.go#L36)):*
> ```Go
> 	case oid.T_bool:
> 		return reflect.TypeOf(false)
> 	case oid.T_numeric:
> ```

Why are float4 and float8 missing? I'm also uncomfortable with mapping decimals to floats since they don't have a 1-1 mapping. Can we use a string instead?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/lib/pq/647)*

conn.go
```diff
@@ -1695,31 +1695,37 @@ func (cn *conn) readExecuteResponse(protocolState string) (res driver.Result, co
...
 	colNames = make([]string, n)
 	colFmts = make([]format, n)
-	colTyps = make([]oid.Oid, n)
+	colTyps = make([]fieldDesc, n)
 	for i := range colNames {
 		colNames[i] = r.string()
 		r.next(6)
-		colTyps[i] = r.oid()
-		r.next(6)
+		colTyps[i] = fieldDesc{
```
cbandy: Field assignments rather than struct expression here, please.

rows.go
```diff
@@ -0,0 +1,95 @@
...
+	case oid.T_int4:
+		return reflect.TypeOf(int32(0))
+	case oid.T_int2:
+		return reflect.TypeOf(int16(0))
+	case oid.T_varchar, oid.T_text:
+		return reflect.TypeOf("")
+	case oid.T_bool:
+		return reflect.TypeOf(false)
+	case oid.T_numeric:
+		return reflect.TypeOf(float64(0))
```
cbandy: Numeric has a much larger range than float64. To match other parts of this package, this case should probably be omitted.

cyberdelia:
@cbandy @mjibson I addressed your feedback, I'll gladly rebase it too.

mjibson:
LGTM. Will wait for another approval before merging.

mjibson:
Moving forward due to lack of requested change comments.

PR #654, opened 2017-08-30
Add support for BC dates

### LINKED ISSUE #653: BC dates break rows.Next()
If a date is stored in the database as `0001-12-31 BC` rows.Next() returns an error indicating `invalid timestamp`.  Support already exists for BC datetimes.

jnewmano:
Existing tests check for support for BC datetimes, adds support for BC dates.

Fixes #653

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -367,8 +367,15 @@ func ParseTimestamp(currentLocation *time.Location, str string) (time.Time, erro
 	timeSep := daySep + 3
 	day := p.mustAtoi(str, daySep+1, timeSep)
 
+	minLen := monSep + len("01-01") + 1
+
+	isBC := strings.HasSuffix(str, " BC")
+	if isBC {
+		minLen += 3
+	}
+
 	var hour, minute, second int
-	if len(str) > monSep+len("01-01")+1 {
+	if len(str) > minLen {
 		p.expect(str, ' ', timeSep)
 		minSep := timeSep + 3
 		p.expect(str, ':', minSep)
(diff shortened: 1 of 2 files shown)


jnewmano:
This broke other tests. Will reopen after they are fixed.

jnewmano:
@tamird How do we get this merged?

tamird:
@jnewmano could you squash these down to one commit? I don't see why all 3 commits are interesting separately.

encode_test.go
```diff
@@ -37,6 +37,7 @@ var timeTests = []struct {
 }{
 	{"22001-02-03", time.Date(22001, time.February, 3, 0, 0, 0, 0, time.FixedZone("", 0))},
 	{"2001-02-03", time.Date(2001, time.February, 3, 0, 0, 0, 0, time.FixedZone("", 0))},
+	{"0001-12-31 BC", time.Date(0, time.December, 31, 0, 0, 0, 0, time.FixedZone("", 0))},
```
mjibson: Add a test using a higher year.

encode.go
```diff
@@ -367,8 +367,14 @@ func ParseTimestamp(currentLocation *time.Location, str string) (time.Time, erro
 	timeSep := daySep + 3
 	day := p.mustAtoi(str, daySep+1, timeSep)
 
+	isBC := str[len(str)-3:] == " BC"
```
mjibson: This will panic if `len(str) < 3`. The previous code correctly checked for that.

jnewmano:
@mjibson Thank you for looking at this PR. I will work on the requested changes and get a squashed commit up.

encode.go
```diff
@@ -367,8 +367,14 @@ func ParseTimestamp(currentLocation *time.Location, str string) (time.Time, erro
 	timeSep := daySep + 3
 	day := p.mustAtoi(str, daySep+1, timeSep)
 
+	isBC := len(str) > 3 && str[len(str)-3:] == " BC"
```
tamird: isn't this `strings.HasSuffix(str, " BC")`? perhaps write:
```golang
minLen := monSep+len("01-01")+1

const suffixBC = " BC"
isBC := strings.HasSuffix(str, suffixBC)
if isBC {
  minLen += len(suffixBC)
}
var ... 
if len(str) > minLen {
...
```

jnewmano: great recommendation, I have update the PR to follow this approach. I opted not to use the const to follow the pattern in the surrounding code.

PR #658, opened 2017-09-21
refactor(Dialer):accept non timeout dialer and create a global variable DefaultDialer to be eventually replaced

eclivus:
Hello !
I needed to connect appengine standard environment to google cloud sql postgres so I made a slight modification to the Dialier to accept overloading.

I'd be happy to discuss about it and to accept comments.

Cheers


arp242:
Won't using `pq.DialOpen(cloudSqlDial{}, "...")` be the same? Or what am I missing?

Since this is quite an old PR and the submitter no longer seems active on GitHub I'll just go ahead and close it, but feel free to come back with more details if you're still having problems and we can re-open.

PR #661, opened 2017-10-07
Fix some issues in tests

ainar-g:
These issues were found by megacheck. Mostly unused values and unchecked
errors.

### DIFF
--- a/conn_test.go
+++ b/conn_test.go
@@ -935,12 +935,14 @@ func TestParseErrorInExtendedQuery(t *testing.T) {
 	db := openTestConn(t)
 	defer db.Close()
 
-	rows, err := db.Query("PARSE_ERROR $1", 1)
-	if err == nil {
-		t.Fatal("expected error")
+	_, err := db.Query("PARSE_ERROR $1", 1)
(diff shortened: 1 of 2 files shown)


conn_test.go
```diff
@@ -935,12 +935,12 @@ func TestParseErrorInExtendedQuery(t *testing.T) {
 	db := openTestConn(t)
 	defer db.Close()
 
-	rows, err := db.Query("PARSE_ERROR $1", 1)
+	_, err := db.Query("PARSE_ERROR $1", 1)
 	if err == nil {
```
tamird: :( this should check for a specific error rather than any non-nil error

conn_test.go
```diff
@@ -1393,35 +1393,29 @@ func TestParseOpts(t *testing.T) {
 
 func TestRuntimeParameters(t *testing.T) {
 	type RuntimeTestResult int
```
tamird: Isn't this type now unused?

ainar-g:
When did GitHub start to close PRs after a rebase?

Either way, @tamird, @mjibson, PTAL?

PR #664, opened 2017-10-10
Check code with megacheck

### LINKED ISSUE #662: Make megacheck pass (mostly) cleanly
In this issue I suggest additional changes, that will make [megacheck](https://github.com/dominikh/go-tools) pass mostly cleanly. I say "mostly" because there is SA1019 check, which raises an error when a deprecated type is used, and we need to use both `driver.Execer` and `driver.Queryer` for compatibility reasons. I've already made some changes in #661. The changes that I want to make are mostly simplifications, e.g.
```diff
-       var s sql.Scanner
-
-       s = Array(&[]bool{})
+       var s sql.Scanner = Array(&[]bool{})
```
and
```diff
-       err = boolSetting("binary_parameters", &cn.binaryParameters)
-       if err != nil {
-               return err
-       }
-       return nil
+       return boolSetting("binary_parameters", &cn.binaryParameters)
```
I can also do this to `.travis.yml`:
```diff
   - ./.travis.sh client_configure
-  - go get golang.org/x/tools/cmd/goimports
+  - go get golang.org/x/tools/cmd/goimports honnef.co/go/tools/cmd/megacheck
```
```diff
   - go vet ./...
+  # Ignore SA1019 in conn_test.go because we need compatibility with Go 1.5, so
+  # we have to use driver.Execer and driver.Queryer.
+  - megacheck -ignore 'github.com/lib/pq/conn_test.go:SA1019'  ./...
   - PQTEST_BINARY_PARAMETERS=no  go test -race -v ./...
```

What do you think?

ainar-g:
Fix everything we can without losing Go 1.5 compatibility. Add
megacheck, locked at 2017.1 release, and enabled for any version of Go
except Go 1.5, to the Travis scripts to check every PR.

Fixes #662.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -149,11 +149,7 @@ func (cn *conn) handleDriverSettings(o values) (err error) {
 	if err != nil {
 		return err
 	}
-	err = boolSetting("binary_parameters", &cn.binaryParameters)
-	if err != nil {
-		return err
-	}
-	return nil
+	return boolSetting("binary_parameters", &cn.binaryParameters)
 }
 
 func (cn *conn) handlePgpass(o values) {
--- a/.travis.sh
+++ b/.travis.sh
@@ -70,4 +70,19 @@ postgresql_uninstall() {
 	sudo rm -rf /var/lib/postgresql
 }
 
+megacheck_install() {
+	# Megacheck is Go 1.6+, so skip if Go 1.5.
+	if [[ "$(go version)" =~ "go1.5" ]]
+	then
+		echo "megacheck not supported, skipping installation"
+		return 0
+	fi
+	# Lock megacheck version at $MEGACHECK_VERSION to prevent spontaneous
+	# new error messages in old code.
+	go get -d honnef.co/go/tools/...
+	git -C $GOPATH/src/honnef.co/go/tools/ checkout $MEGACHECK_VERSION
+	go install honnef.co/go/tools/cmd/megacheck
+	megacheck --version
+}
+
 $1
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ env:
     - PQGOSSLTESTS=1
     - PQSSLCERTTEST_PATH=$PWD/certs
     - PGHOST=127.0.0.1
+    - MEGACHECK_VERSION=2017.1
   matrix:
     - PGVERSION=9.6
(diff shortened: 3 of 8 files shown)


conn.go
```diff
@@ -149,11 +149,7 @@ func (cn *conn) handleDriverSettings(o values) (err error) {
 	if err != nil {
 		return err
 	}
-	err = boolSetting("binary_parameters", &cn.binaryParameters)
-	if err != nil {
-		return err
-	}
-	return nil
+	return boolSetting("binary_parameters", &cn.binaryParameters)
```
johto: I don't quite agree with this change.  Now the last setting is different from the rest just to save a couple of lines of code.

ainar-g: Why? Isn't writing
```go
err = foo()
if err != nil {
    return err
}
return nil
```
the same as
```go
booleanVar = foo()
if booleanVar == true {
    return true
}
return false
```
? Would you agree that the latter is bad style?

johto: > Would you agree that the latter is bad style?

That's not quite the same thing.  What I'm saying (to try and change your analogy to what I think would be comparable) is that when you go from this:

```Go
if foo() {
    return true
}
if bar() {
    return true
}
return false
```

to this:

```Go
if foo() {
    return true
}
return bar()
```

you're making the last check "special".  If somebody wants to add one more check and it feels natural to do it at the end, they suddenly have to change the structure of the code:

```Go
if foo() {
    return true
}
if bar() {
    return true
}
return baz()
```


And in this context I think it actually makes a big difference because the code is essentially just a list of options.

tamird: FWIW, I don't think this is a bit worth picking - do what the linter says and don't overthink it.

ainar-g: @johto I thought like that before, but I've been using megacheck at work (with quite a few bugs found by it), and I have to say that I've noticed two things:
1. Code like that doesn't really change that often.
2. When it does, it isn't as painful as I had anticipated it to be.

@tamird I tend to agree. It is kinda like arguing over gofmt, just with a third-party tool.

tamird:
Thank you, @ainar-g!

conn.go
johto: OK.  Consider my complaint overruled.  I can see the value in having the codebase be completely lint clean.

PR #666, opened 2017-10-18
Avoid potential runtime panic in connection initialization

necaris:
Attempt to use `$HOME` to find the user's home directory to avoid calling `user.Current`, which can runtime panic due to a known glibc bug. This emulates the fix in https://go-review.googlesource.com/c/oauth2/+/34175, to work around http://golang.org/issue/13470 .


geek1011:
This does solve go-gitea/gitea#2728 based on my testing.

PR #668, opened 2017-10-21
Add PostgreSQL 10 to the list of versions to check

ainar-g:
This change adds PostgreSQL 10 support for Travis builders.


tamird:
Thanks!

PR #675, opened 2017-11-06
Add sql.Pinger implementation.

### LINKED ISSUE #533: Implement sql.Pinger interface
Looking at issue #531, I was thinking about implementing and submitting an implementation of Pinger. 

I read a few posts around, and it might be worthwhile to put a value on the context for determining if you're communicating with a pgBouncer instance or Postgres directly?

I did find a bug that was mentioned in the Perl database handler...
[Post](http://blog.endpoint.com/2015/01/dbdpg-one-ping-to-rule-them-all.html) 
[Commit related](http://www.nntp.perl.org/group/perl.dbd.pg.changes/2015/01/msg1355.html)

Essentially, implementation of ping similar to the Perl DBH package by executing `/* libpq ping test */` would destroy the state of any running COPY statements. Would this be something to consider?

itsjamie:
Executes a query `SELECT 'lib/pq ping test'` against the configured database.

Closes #533.

### DIFF
--- a/conn_go18.go
+++ b/conn_go18.go
@@ -83,10 +83,14 @@ func (cn *conn) Ping(ctx context.Context) error {
 
 	rows, err := cn.simpleQuery(`SELECT 'lib/pq ping test';`)
 	if err != nil {
-		return err
+		return driver.ErrBadConn
 	}
 
-	return rows.Close()
+	if err := rows.Close(); err != nil {
+		return driver.ErrBadConn
+	}
+	
+	return nil
 }
 
 func (cn *conn) watchCancel(ctx context.Context) func() {
--- a/conn_go18.go
+++ b/conn_go18.go
@@ -76,6 +76,19 @@ func (cn *conn) BeginTx(ctx context.Context, opts driver.TxOptions) (driver.Tx, 
 	return tx, nil
 }
 
+func (cn *conn) Ping(ctx context.Context) error {
+	if finish := cn.watchCancel(ctx); finish != nil {
+		defer finish()
+	}
+
+	rows, err := cn.simpleQuery(`SELECT 'lib/pq ping test';`)
+	if err != nil {
+		return err
+	}
+
+	return rows.Close()
+}
+
 func (cn *conn) watchCancel(ctx context.Context) func() {
 	if done := ctx.Done(); done != nil {
 		finished := make(chan struct{})
--- a/go18_test.go
+++ b/go18_test.go
@@ -317,3 +317,22 @@ func TestTxOptions(t *testing.T) {
 		t.Errorf("Expected error to mention isolation level, got %q", err)
 	}
 }
+
+func TestPing(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+
+	db := openTestConn(t)
+	defer db.Close()
+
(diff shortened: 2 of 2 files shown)


itsjamie:
The failed tests here just likely need to be re-ran, it looks like googlesource didn't respond for one of the test runs.

Any comment on the PR?

Thanks.

uhoh-itsmaciek:
I restarted the failing test job for you. The code looks good to me, but I'm not an active maintainer so I'll let someone else review and merge.

cbandy:
Is it possible or reasonable to send a Flush message instead of a query?

itsjamie:
@cbandy I'm not sure, I did the initial research on how others we're accomplishing this and found a few notes from other libraries that we're simply accomplishing it with SELECTs, see the #533 ticket for any discussion points. That's not to say there isn't a better method, 

@uhoh-itsmaciek Thank you for running the tests again.

conn_go18.go
```diff
@@ -76,6 +76,19 @@ func (cn *conn) BeginTx(ctx context.Context, opts driver.TxOptions) (driver.Tx,
 	return tx, nil
 }
 
+func (cn *conn) Ping(ctx context.Context) error {
```
mjibson: The godoc for the Ping function (https://golang.org/pkg/database/sql/driver/#Pinger) says it should return `driver.ErrBadConn`. I would like that error to be explicitly returned instead of returning whatever it is simpleQuery and rows.Close return (which I think is indeed driver.ErrBadConn from a quick look). We should hard code it here because that's what the API requires).

go18_test.go
```diff
@@ -317,3 +317,22 @@ func TestTxOptions(t *testing.T) {
...
+	db := openTestConn(t)
+	defer db.Close()
+
+	if err := db.PingContext(ctx); err != nil {
+		t.Fatal("expected success")
+	}
+
+	cancel()
+
+	if err := db.PingContext(ctx); err == nil {
```
mjibson: Can we come up with more tests that don't use context cancellation? I'd like a test where the conn has actually died and the query itself fails.

itsjamie: Is there a method in Postgres via Travis that you're aware of to temporarily prevent access?

Alternatively, could grab one of the test SQL drivers that returns a good conn, and then mock the Ping call to return an error. It looks though like the driver is trying to stay using actual Postgres as much as possible.

conn_go18.go
itsjamie: Will update this tonight. Thanks.

itsjamie: This should be solved with latest commit.

go18_test.go
itsjamie: Here's a potential solution, what do you think..

Start a transaction, execute a query to grab the pid that postgres has given that open connection.

Grab a new connection, and execute `SELECT pg_terminate_backend(pid)` from the in-progress transaction to kill the backing process. This should cause the connection to die, and the commit should fail.

dmmulroy:
Any updates on this issue?

andradei:
Why was this PR closed? Not having Pinger properly implemented is a big limitation for trying to test database connectivity.

andradei:
Just for reference to those who get here. #737 is the preferred PR to solve the problem.

PR #690, opened 2017-12-20
support 'true', 'false', '1', '0' for bool arrays

nuggien:


iangudger:
My interpretation of that documentation is that Postgres with accept all of those value, but they get encoded the same and thus only the 't' and 'f' values get returned.

arp242:
> My interpretation of that documentation is that Postgres with accept all of those value, but they get encoded the same and thus only the 't' and 'f' values get returned.

Yes, it's fairly explicit about that:

> The datatype output function for type boolean always emits either t or f

---

Given the lack of feedback and that this is a few years old I'll just go ahead and close it, but feel free to come back with more details if you're still having problems and we can re-open.

PR #698, opened 2018-01-18
Fix connection leak on conn.ssl or conn.startup failure

### LINKED ISSUE #697: A connection is leaked on ssl or startup failure
DialOpen leaks a connection allocated [here](https://github.com/lib/pq/blob/master/conn.go#L338) if `conn.ssl` or `conn.startup` panic, which they do on failure. I will send a PR to fix this shortly.

iangudger:
Fixes #697

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -339,6 +339,15 @@ func DialOpen(d Dialer, name string) (_ driver.Conn, err error) {
 	if err != nil {
 		return nil, err
 	}
+
+	// cn.ssl and cn.startup panic on error. Make sure we don't leak cn.c.
+	panicking := true
+	defer func() {
+		if panicking {
+			cn.c.Close()
+		}
+	}()
+
 	cn.ssl(o)
 	cn.buf = bufio.NewReader(cn.c)
 	cn.startup(o)
@@ -347,6 +356,7 @@ func DialOpen(d Dialer, name string) (_ driver.Conn, err error) {
 	if timeout, ok := o["connect_timeout"]; ok && timeout != "0" {
 		err = cn.c.SetDeadline(time.Time{})
 	}
+	panicking = false
 	return cn, err
 }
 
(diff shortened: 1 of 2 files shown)


conn.go
```diff
@@ -339,6 +339,15 @@ func DialOpen(d Dialer, name string) (_ driver.Conn, err error) {
...
+
+	// cn.ssl and cn.startup panic on error. Make sure we don't leak cn.c.
+	panicking := true
+	defer func() {
+		if panicking {
+			cn.c.Close()
+		}
+	}()
+
 	cn.ssl(o)
```
hugelgupf: It would be ideal if these functions returned errors instead of panicking... but this works as a hot fix. Would leave that to maintainers.

https://github.com/golang/go/wiki/CodeReviewComments#dont-panic

conn.go
```diff
@@ -339,6 +339,15 @@ func DialOpen(d Dialer, name string) (_ driver.Conn, err error) {
 	if err != nil {
 		return nil, err
 	}
+
+	// cn.ssl and cn.startup panic on error. Make sure we don't leak cn.c.
+	panicking := true
+	defer func() {
+		if panicking {
```
mjibson: Since `err` is a named return value I *think* you can just do `if err != nil` and drop `panicking`.

conn.go
mjibson: Indeed. Ideally `cn.startup` would return an error. But that's a larger change, so I'm fine with this now.

mjibson:
The build errors aren't your fault: due to the golang tools repo using stuff not present in older Go versions. We should fix that, or more probably just stop testing unsupported versions of Go.

conn.go
hugelgupf: I thought so, too, initially, but you can't. This defer runs before the recover-defer, so err will still be nil here.

mjibson: Ugh. I'm not aware of a project that is more popular and more un-Goish than lib/pq.

PR #700, opened 2018-01-19
Fix deadlock in Listener.resync

### LINKED ISSUE #699: Listener.resync can spin forever due to race condition
It is possible for both resync and resync.func1 to block indefinitely:

- Listener.connect creates a new connection, and schedules ListenerConn.listenerConnMain
- listenerConnMain detects an error and closes l.notificationChan
- connect calls resync
- resync schedules resync.func1, which is a closure over notificationChan. However func1 is not immediately run
- resync detects that notificationChan is closed, and sets it to nil (race happens here)
- resync now waits on doneChan
- func1 is scheduled and cn.Listen returns false, error
- func1 attemps to range over notificationChan, which is nil by now
- range over a nil chan blocks
- doneChan is never written to
- both resync and resync.func1 are deadlocked

lmb:
Listener.resync can execute in a fashion where notificationChan equals nil in the anonymous inner function. This leads to a deadlock if the function tries to range over notificationChan.

Fixes #699

### DIFF
--- a/notify.go
+++ b/notify.go
@@ -637,7 +637,7 @@ func (l *Listener) disconnectCleanup() error {
 // after the connection has been established.
 func (l *Listener) resync(cn *ListenerConn, notificationChan <-chan *Notification) error {
 	doneChan := make(chan error)
-	go func() {
+	go func(notificationChan <-chan *Notification) {
 		for channel := range l.channels {
 			// If we got a response, return that error to our caller as it's
 			// going to be more descriptive than cn.Err().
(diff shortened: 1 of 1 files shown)


lmb:
This is a minimally invasive fix. It feels like the whole inner goroutine isn't strictly necessary, but then I don't understand the code well enough to make that call.

johto:
> It feels like the whole inner goroutine isn't strictly necessary

Consider resyncing with two channels, "chan_foo" and "chan_bar":

1. cn.Listen("chan_foo") finishes
2. The servers sends more than [32](https://github.com/lib/pq/blob/27ea5d92de30060e7121ddd543fe14e9a327e0cc/notify.go#L690) notifications for chan_foo
3. cn.Listen("chan_bar") starts

With just a single goroutine you'd deadlock there.  That's why another one is set up to only read from the notification channel.

I guess another approach would be to make ListenerConn discard all notifications until the caller specifically asks for them to be delivered, but that would slow down the normal path since you'd have to check an atomic variable every time to know whether to discard a notification or not.

johto:
Merged.  Thanks for the report and the fix!

PR #704, opened 2018-01-23
Stop testing with go 1.5

mjibson:
The correct fix for this would involve removing all the Go 1.5 stuff in the `script` block, but I'm not going to allocate time to that currently. For now, this should unbreak all of the PR build failures.


mjibson:
I'm going to merge this in the next few days unless someone has an objection.

PR #709, opened 2018-02-03
Fix Listen() hanging indefinitely when listener is closed

nvx:
If a Listener is in the middle of reconnecting (eg, because the server name is incorrect) and Listener.Close() is called, any outstanding calls to Listener.Listen() will hang indefinitely as it waits for Listener.reconnectCond.

This simple PR triggers a reconnectCond broadcast at the end of Listener.Close() which unblocks any outstanding calls to Listener.Listen() to allow proper cleanup to occur.

### DIFF
--- a/notify.go
+++ b/notify.go
@@ -725,6 +725,9 @@ func (l *Listener) Close() error {
 	}
 	l.isClosed = true
 
+	// Unblock calls to Listen()
+	l.reconnectCond.Broadcast()
+
 	return nil
 }
 


wI2L:
+1
Let's revive this PR.
I ran into the same issue and found an existing PR for this.

nvx:
Yeah I'm not sure what the hold up is on this?

johto:
The fix looks correct to me.  Listen() already seems to be assuming this code to be here.

wI2L:
Just to give some insight about this. After reading the `Listen()`
documentation, I ended up using a `time.AfterFunc` callback to  close the
listener after a timeout to avoid blocking on it. The timer returned is
stopped if the call returns no error.

I tried with and without the change in this PR and I can confirm it works
perfectly with the broadcast on the cond in `Listen()`.

Le mer. 22 août 2018 à 12:10, Marko Tiikkaja <notifications@github.com> a
écrit :

johto:
I've merged this into the master branch.  Thanks!

Issue #426, opened 2016-02-03
Support parameter markers on `COMMENT` queries.


homme:
Currently the following produces an error:
```
err = tx.Exec("COMMENT ON TABLE mytable IS $1", "my comment")
```

`err` is a `*pq.Error` with the following fields set:

```
code=42601 
message=syntax error at or near "$1" 
file=scan.l 
line=1082 
position=72 
routine=scanner_yyerror 
severity=ERROR
```

An alternative solution would be to export a `pq.EscapeString(input string) (output string)` type function.

uhoh-itsmaciek:
Our parameter support is based on what Postgres itself supports, and I don't believe `COMMENT` queries are parameterizable. Postgres has something like the `EscapeString` you propose via [quote_literal](https://github.com/postgres/postgres/blob/ee943004466418595363d567f18c053bae407792/src/backend/utils/adt/quote.c#L77); adding that could be useful.

uhoh-itsmaciek:
To clarify, `quote_literal` is the internal Postgres function; it is exposed in the libpq client library via http://www.postgresql.org/docs/9.1/static/libpq-exec.html#LIBPQ-EXEC-ESCAPE-STRING

homme:
If `COMMENT` queries can't be parameterised then yes, an escaping/quoting function would be great to replace the hack I currently have.  Maybe `pq.QuoteString` to complement `pq.QuoteIdentifier`?  I'm not sure how the difference between the `PQescapeStringConn` and `PQescapeString` C functions in the above link would apply to the Go package?

johto:
> If COMMENT queries can't be parameterised then yes, an escaping/quoting function would be great to replace the hack I currently have. Maybe pq.QuoteString to complement pq.QuoteIdentifier?

Please don't deviate from the standard terminology.  Call it pq.QuoteLiteral.

> I'm not sure how the difference between the PQescapeStringConn and PQescapeString C functions in the above link would apply to the Go package?

You should be looking at PQescapeLiteral, specifically this part:

```C
    /*
     * If we are escaping a literal that contains backslashes, we use the
     * escape string syntax so that the result is correct under either value
     * of standard_conforming_strings.  We also emit a leading space in this
     * case, to guard against the possibility that the result might be
     * interpolated immediately following an identifier.
     */
    if (!as_ident && num_backslashes > 0)
    {
        *rp++ = ' ';
        *rp++ = 'E';
    }
```

[...]

PR #718, opened 2018-03-09
Add QuoteLiteral function

nickuraltsev:
This PR adds `QuoteLiteral` function that can be useful with SQL statements that don't support parametrization. The problem is discussed here: #426.

### DIFF
--- a/conn_test.go
+++ b/conn_test.go
@@ -1549,6 +1549,27 @@ func TestQuoteIdentifier(t *testing.T) {
 	}
 }
 
+func TestQuoteLiteral(t *testing.T) {
+	var cases = []struct {
+		input string
+		want  string
+	}{
+		{`foo`, `'foo'`},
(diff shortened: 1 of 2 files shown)


nickuraltsev:
Ping

conn_test.go
```diff
@@ -1550,6 +1550,27 @@ func TestQuoteIdentifier(t *testing.T) {
 	}
 }
 
+func TestQuoteLiteral(t *testing.T) {
```
mjibson: I would like this test to just be a slice of strings. The test calls QuoteLiteral, puts that string in a select statement, and verifies that the return value is the same is the original string. The current test doesn't test that postgres itself parses the strings correctly. Add entries with non-ascii runes too. Like `\x00` and others.

arp242:
Looks like this was added in #852

PR #719, opened 2018-03-10
Fix failing builds

nickuraltsev:
Builds are [failing](https://travis-ci.org/lib/pq/jobs/351414509) now because of the following issue: `TestInfinityTimestamp` expects the error message to start with `sql: Scan error on column index 0: unsupported`, but in the latest version of Go, the message is `sql: Scan error on column index 0, name "timestamp": unsupported`.

Please see this commit: https://github.com/golang/go/commit/8da7706bb4dc609689fd69e196ddaa800908c31f

This PR updates `TestInfinityTimestamp` to support both error messages.

### DIFF
--- a/encode_test.go
+++ b/encode_test.go
@@ -304,11 +304,9 @@ func TestInfinityTimestamp(t *testing.T) {
 	var err error
 	var resultT time.Time
 
-	expectedErrorStrRegexp, err := regexp.Compile(
+	expectedErrorStrRegexp := regexp.MustCompile(
 		`^sql: Scan error on column index 0(, name "timestamp(tz)?"|): unsupported`)
-	if err != nil {
-		t.Fatal("Error parsing expected error regexp")
-	}
+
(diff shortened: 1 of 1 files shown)


encode_test.go
```diff
@@ -304,24 +304,29 @@ func TestInfinityTimestamp(t *testing.T) {
 	var err error
 	var resultT time.Time
 
-	expectedErrorStrPrefix := `sql: Scan error on column index 0: unsupported`
+	expectedErrorStrRegexp, err := regexp.Compile(
```
mjibson: regexp.MustCompile is fine here since it should always succeed

nickuraltsev: Sure, I'll update the PR shortly

nickuraltsev:
@mjibson I've just updated the PR to use `MustCompile`. Can you do Squash and Merge?

mjibson:
Sure, I can do that.

PR #722, opened 2018-03-15
Close bad connection to solve CLOSE_WAIT stuck

wmin0:
Somehow server side would close connection actively (timeout, server close, etc...). Therefore, client should ack close in this situation, otherwise client process will full of CLOSE_WAIT and stuck finally.


wmin0:
https://golang.org/src/database/sql/sql.go#L1045 only close errBadConn. If lib/pq doesn't close connection when set conn.bad flag, the close of the bad connection will not be called.

domac:
> https://golang.org/src/database/sql/sql.go#L1045 only close errBadConn. If lib/pq doesn't close connection when set conn.bad flag, the close of the bad connection will not be called.

Hi, I would like to ask if the repair of this code works normally and can reduce the establishment of time_wait?

domac:
> https://golang.org/src/database/sql/sql.go#L1045 only close errBadConn. If lib/pq doesn't close connection when set conn.bad flag, the close of the bad connection will not be called.

I would like to ask if the repair of this code works normally and can reduce the establishment of close_wait stuck?

domac:
> This pr I worked with was at golang 10, you can find golang source code I referred here https://github.com/golang/go/blob/release-branch.go1.10/src/database/sql/sql.go#L1045.
> And @domac your question is yes, or I don't propose this pr ._.
> Maybe I should add some stress test in bench_test.go.

Thank you your reply.... This means that if my go version is 1.9, I need to modify it in the way of your previous pr (https://github.com/lib/pq/commit/d6b89069109a8b1fa68f423541057f1eb45477fd)  to avoid the close_wait problem?

wmin0:
https://github.com/golang/go/blob/release-branch.go1.9/src/database/sql/sql.go#L946
In golang 1.9, the leak seems to be worse because of it only closes connection on timeout. This pr should work on it.

wmin0:
My ci still fails because of race condition but it's not related to this pr ....
https://travis-ci.org/lib/pq/jobs/472262943

arp242:
It seems conn.Close() is always called on ErrBadConn by database/sql? I tried to replicate this with the pqtest.Fake server, but I can't really reproduce it: conn.Close() is always called. 

The logic in conn.Close() where it tries to send a Terminate message to the server is probably better than outright closing the connection?

Since this is a few years old I'll just go ahead and close it, but let me know if I'm mistaken and we can re-open.

Issue #422, opened 2016-01-27
Should `net.OpError` not result in an `ErrBadConn`?


jenanwise:
Hi there!

In [`errRecover`](https://github.com/lib/pq/blob/8ad2b298cadd691a77015666a5372eae5dbfac8f/error.go#L489), `net.OpError`s are turned into `driver.ErrBadConn`. My understanding is that drivers should never return `driver.ErrBadConn` if work may been completed on the server. The `net.OpError` handling was added with b48e1a27f71c015f27fcc99051af2b0e5014aa0f, but I don't see an explanation for the original intent of the current behavior. I assume it was for handling initial connection errors, but isn't it reasonable that a `net.OpError` may be triggered on an existing connection, potentially in the middle of a query?

My particular use case is: I've implemented a custom driver that wraps `pq` and sets a connection read and write deadline on every `Read()` and `Write()`. If the timeout hits, it will return a `net.OpError`. In this case, the connection should be marked bad, but the query not retried.

Am I missing something about the way pq handles this error?

johto:
> Am I missing something about the way pq handles this error?

Probably the fact that pq implements ErrBadConn incorrectly in this regard.  I've previously stated that this should be changed, but that school of thought didn't gain much support.

jenanwise:
Aren't you the maintainer? :grin: 

Changing

```go
	case *net.OpError:
		*err = driver.ErrBadConn
```

to

```go
	case *net.OpError:
		c.bad = true
		*err = v
```

doesn't seem to affect the test suite... probably because that particular path needs some tests. I'd be happy to submit a patch if I'm on the right path.

johto:
> Aren't you the maintainer?

I am not and never have been "the maintainer".  I'm one of the maintainers, thought I was probably the most active maintainer a while ago.  But I can't decide things like this on behalf of the project.

In any case, I think I've made my position clear.  Other maintainers are free to chime in.

uhoh-itsmaciek:
I merge straightforward patches here and there but I'm not actively using Go these days, so my opinion shouldn't hold much weight here. @johto do you mean #219? For what it's worth, I think the change is reasonable. Any thoughts on this @freeformz?

freeformz:
To quote the docs:

> To prevent duplicate operations, ErrBadConn should NOT be returned if there's a possibility that the database server might have performed the operation. Even if the server sends back an error, you shouldn't return ErrBadConn.

[...]

PR #728, opened 2018-03-19
Fix handling timeout errors

vijaykarthik-rubrik:
The following comment in `database/sql/driver/driver.go` mentions
```
// To prevent duplicate operations, ErrBadConn should NOT be returned
// if there's a possibility that the database server might have
// performed the operation. Even if the server sends back an error,
// you shouldn't return ErrBadConn.
var ErrBadConn = errors.New("driver: bad connection")
```

Network error could be a timeout error where the database might
have executed the query.

This commit adds a check so that ErrBadConn is not returned
for timeout errors


andreimatei:
Please reference https://github.com/lib/pq/issues/422 in the commit.
Haven't looked at the PR, but I believe the problem is more general than just timeouts.

vijaykarthik-rubrik:
The original post in #422 is related to timeouts. I'm facing a similar issue where an `INSERT` query times out, and is internally retried by `database/sql`, and I get a duplicate primary key error for a row being inserted the first time (because the first insert succeeded in the database, but timed out).

mjibson:
I agree this should apply for more tests. I've opened #730 to address this.

PR #729, opened 2018-03-19
Travis update

mjibson:
I think it's safe to stop supporting Go versions < 1.8 since they aren't supported by Go itself anymore.


mjibson:
I think it's safe to stop supporting Go versions < 1.8 since they aren't supported by Go itself anymore.

Issue #681, opened 2017-11-24
Don't panic and errcheck


ainar-g:
I've noticed that the code uses `panic` for error handling in several places. In this issue I propose removing those panics and instead use explicit errors everywhere. To ensure that all errors are checked or explicitly ignored, I propose adding the `errcheck` tool to the TravisCI script.

Any reasons not to do this?

mjibson:
Some of us have wanted to do this for a while but lacked motivation. Please go ahead, I'd love to see this change.

chrispassas:
I agree, this would be a welcome change.

ainar-g:
@mjibson What should I do about "impossible" panics? Like `panic("unreachable")` and those that marked as "this shouldn't happen"? In my code I usually don't panic at all, so "impossible" code paths return an error, but I'm not sure if we want that in lib/pq as well.

mjibson:
I think they should return an error.

mjibson:
Let me revise that. I'm looking at `ListenerConn.Ping`. I think it should be:

```
func (l *ListenerConn) Ping() error {
	_, err := l.ExecSimpleQuery("")
	return err
}
```

Instead of all the complication it has now. Are there other examples?

ainar-g:
Sanity checks in `(*ListenerConn).ExecSimpleQuery`, `(*Listener).disconnectCleanup()`, `hQuote`, etc.

mjibson:
hQuote is fine as is. Other go functions in the stdlib panic on bad inputs. Unsure about the other two. I looked at them and it's not immediately clear if you can just return an error. It feels like you also need to make it do the other side effects of the err recovery catcher, which set the conn to bad. I don't have the time to dig deeper now.

Overall guidance about removing panics is: make sure the side effects from any defers that now won't happen are the same.

ainar-g:
@mjibson If we plan on adding `errcheck` to the CI, should I also explicitly ignore errors from `(*conn).Close()` and `net.Conn.Close()`? Most of the time something is closed when there is already an error, so we either ignore the closing error with

    // We don't care about errors from Close.
    _ = cn.Close()

, or do something like 

    return fmt.Errorf("main err: %q, close err: %q", err, errClose)

PR #734, opened 2018-03-24
Don't panic in SSL code

ainar-g:
This change removes panics from SSL code. This is a part of the ongoing process of removing panics from lib/pq code.

Refs #681.

### DIFF
--- a/ssl.go
+++ b/ssl.go
@@ -12,7 +12,7 @@ import (
 
 // ssl generates a function to upgrade a net.Conn based on the "sslmode" and
 // related settings. The function is nil when no upgrade should take place.
-func ssl(o values) func(net.Conn) net.Conn {
+func ssl(o values) (func(net.Conn) (net.Conn, error), error) {
 	verifyCaOnly := false
 	tlsConf := tls.Config{}
 	switch mode := o["sslmode"]; mode {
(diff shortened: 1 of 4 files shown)


ssl.go
```diff
@@ -12,7 +12,7 @@ import (
 
 // ssl generates a function to upgrade a net.Conn based on the "sslmode" and
 // related settings. The function is nil when no upgrade should take place.
-func ssl(o values) func(net.Conn) net.Conn {
+func ssl(o values) (f func(net.Conn) (net.Conn, error), err error) {
```
mjibson: What's the reason to use a named return here? I don't see one. Ditto below.

ssl.go
```diff
@@ -31,7 +31,7 @@ func ssl(o values) func(net.Conn) net.Conn {
 		// behavior is discouraged, and applications that need certificate
 		// validation should always use verify-ca or verify-full.
 		if sslrootcert, ok := o["sslrootcert"]; ok {
-			if _, err := os.Stat(sslrootcert); err == nil {
```
mjibson: This should stay since the `err` here is definitely not the same as the one in the return.

ssl.go
ainar-g: Removed.

ssl.go
ainar-g: Done.

ainar-g:
@mjibson PTAL.

cbandy:
Quite late for a review, but I wanted to mention that this line is partly why I left the panics alone:

https://github.com/lib/pq/commit/f4fe5872e124ea1b0ebc74039d87da5d399532dd

https://github.com/lib/pq/pull/525#issuecomment-256772579

PR #739, opened 2018-04-09
fix NullTime.Scan(value) parsing if value is []uint8

hhh0pE:
Recently I found that code with field pq.NullTime works incorrect - it doesn't write right value from DB(DB field TIMESTAMP WITH TIME ZONE, nullable). So if there is a field in DB with date, NullTime parses it like {"Time":"0001-01-01T00:00:00Z","Valid":false}.
After debug I found that interface type of value in NullTime.Scan has []uint8 type, and after casting to string it can be parsed as postgres data with already exist method PasreTimestamp. And previous code doesn't work because []uint8 cannot be casted to time.Time.
I don't know how it worked before(maybe the case in Go version), but in my enviroment: Go1.10.1, Postgres 9.4 && Postgres 9.6 there was a bug that I fixed for my code and created Pull Request.
Changes are backward-compatible. (if value has []uint8 - use ParseTimetamp, if time.Time - just cast as it was).

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -590,7 +590,17 @@ type NullTime struct {
 
 // Scan implements the Scanner interface.
 func (nt *NullTime) Scan(value interface{}) error {
-	nt.Time, nt.Valid = value.(time.Time)
+	switch typed := value.(type) {
+	case []uint8:
+		var parse_err error
+		nt.Time, parse_err = ParseTimestamp(nil, string(typed))
+		nt.Valid = parse_err == nil
+		return parse_err
+	case time.Time:
+		nt.Time = typed
+		nt.Valid = true
+		return nil
+	}
 	return nil
 }
 


johto:
> After debug I found that interface type of value in NullTime.Scan has []uint8 type

You didn't add a test case or provide a sufficient explanation as to why exactly you're getting []uint8s back if this is supposed to be a timestamptz field.  That shouldn't happen.

johto:
> I have no time to find out why it happens(my suggestion that it is an ORM influence), but my solution doesn't affect already written code and add functionality for []uint8 type.

No, that's not how this works.  As far as I can tell, your code is just dead code and there's no reason to commit it.  If you don't have the time to prove that your code does anything useful, then it's not going in.

rogerhub:
I’ve been using NullTime with nullable timestamp + tz fields and it works perfectly fine. Can you provide example code that shows why []uint8 is needed?

Also, the JSON stuff sounds like it doesn’t belong in this project. Why not wrap NullTime and implement JSON marshaling in your own code?

arp242:
`pq.NullTime` is now a type alias for the identical `sql.NullTime`, so this is no longer relevant.

PR #743, opened 2018-04-19
Add length support for bit and varbit type

pf-qiu:
ColumnType is the interface to obtain length, precision, scale of
a database column. The lib/pq implementation only support char,
varchar, text, bytea. Bit and varbit support is missing.

Signed-off-by: Huiliang Liu <huliu@pivotal.io>


macroyuyang:
@nickuraltsev Could you please help have a look at this? We are blocking on this, thanks!

PR #744, opened 2018-04-23
add support for plan9

mjl-:
Make this compile (and work) by adding a file user_plan9.go. It's nearly identical to the posix version.

On plan9 the test TestCloseBadConn in conn_test.go was failing. The hardcoded localhost:5432 won't work (postgres isn't available on plan9). The README already encourages you to use environment variables to override database connection parameters, so this patch makes the code look at env vars.

### DIFF
--- a/conn_test.go
+++ b/conn_test.go
@@ -724,7 +724,17 @@ func TestBadConn(t *testing.T) {
 // TestCloseBadConn tests that the underlying connection can be closed with
 // Close after an error.
 func TestCloseBadConn(t *testing.T) {
-	nc, err := net.Dial("tcp", "localhost:5432")
+	env := parseEnviron(os.Environ())
+	host := env["host"]
+	port := env["port"]
+	if host == "" {
+		host = "localhost"
+	}
+	if port == "" {
+		port = "5432"
+	}
+	addr := host + ":" + port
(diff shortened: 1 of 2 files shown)


mjl-:
can someone restart the build?
it failed due to golint missing (externally, which has been fixed, see https://github.com/golang/lint/issues/397).

arp242:
In #952, Plan 9 support was added via user_posix.go rather than creating a new user_plan9.go. The only difference seems to be the casing of the environment variable. Does it need to be lower-case?

The test was fixed in #905.

mjl-:
> The only difference seems to be the casing of the environment variable. Does it need to be lower-case?

Yes, in plan9 the standard environment variables are in lower case. See "$user" and "$home" in https://9p.io/magic/man2html/1/intro. 

But, it's probably fine to not make further changes. `user.Current` should
normally just succeed. It would only fail if something was wrongly set up in
the environment, and the user/developer can just fix their environment. The
fixed test (looking at the PG-env vars) also look good.

So, I suggest we just close this PR. Thanks for looking at this, and for
working on lib/pq.

arp242:
Okay, I changed it to lower-case for Plan 9 with #1233. Maybe not necessary, but easy enough to do so might as well.

Issue #746, opened 2018-05-02
data rece when use pq.CopyIn


felipeweb:
I found this data race when use `pq.CopyIn` and `tx.Prepare` together 

```
== RUN   TestPostgres_BatchInsertCopy/batch_copy
==================
WARNING: DATA RACE
Read at 0x00c4200b2860 by goroutine 64:
  internal/race.ReadRange()
      /home/travis/.gimme/versions/go1.8.linux.amd64/src/internal/race/race.go:45 +0x42
  syscall.Write()
      /home/travis/.gimme/versions/go1.8.linux.amd64/src/syscall/syscall_unix.go:183 +0xaa
  net.(*netFD).Write()
      /home/travis/.gimme/versions/go1.8.linux.amd64/src/net/fd_unix.go:332 +0x1c0
  net.(*conn).Write()
      /home/travis/.gimme/versions/go1.8.linux.amd64/src/net/net.go:193 +0x96
  net.(*TCPConn).Write()
      <autogenerated>:73 +0x72
  github.com/lib/pq.(*conn).send()
      /home/travis/gopath/src/github.com/lib/pq/conn.go:915 +0x240
  github.com/lib/pq.(*conn).simpleExec()
      /home/travis/gopath/src/github.com/lib/pq/conn.go:630 +0x24b
  github.com/lib/pq.(*conn).Commit()
      /home/travis/gopath/src/github.com/lib/pq/conn.go:585 +0x1f1
  database/sql.(*Tx).Commit.func1()
      /home/travis/.gimme/versions/go1.8.linux.amd64/src/database/sql/sql.go:1509 +0x6e
  database/sql.withLock()
      /home/travis/.gimme/versions/go1.8.linux.amd64/src/database/sql/sql.go:2545 +0x76
  database/sql.(*Tx).Commit()
      /home/travis/.gimme/versions/go1.8.linux.amd64/src/database/sql/sql.go:1510 +0x16b
  github.com/prest/adapters/postgres.(*Postgres).BatchInsertCopy()
      github.com/prest/adapters/postgres/_test/_obj_test/postgres.go:825 +0x95d
  github.com/prest/adapters/postgres.TestPostgres_BatchInsertCopy.func1()
      /home/travis/gopath/src/github.com/prest/adapters/postgres/postgres_test.go:1231 +0x17d
  testing.tRunner()
      /home/travis/.gimme/versions/go1.8.linux.amd64/src/testing/testing.go:657 +0x107
[...]

PR #747, opened 2018-05-02
lock pooling when copy

felipeweb:
related to https://github.com/lib/pq/issues/746


rogerhub:
Why isn't the solution to add another

```go
if cn.inCopy {
	return nil, errCopyInProgress
}
```

to conn.Prepare? The conn should really on support one COPY at a time.

PR #749, opened 2018-05-19
Add support for database/sql Connector and OpenDB

leighmcculloch:
What
===
Add support for `database/sql/driver` `Connector` and `database/sql`
`OpenDB` that were added in go1.10.

Why
===
To allow users to open a connection without using the registered drivers
interface, and instead explicitly getting a `Connector` from the this
driver and passing that connector to `database/sql` `OpenDB`.

Notes
===
- The new types and functions are only built for go1.10 or newer because
the `database/sql/driver` `Connector` interface and `database/sql`
`OpenDB` function are only available in go1.10 or later.
- The `NewConnector` has support for context, but the context isn't used
at this point. Use of it could be added if desired, but that could occur
separately to this change.

### DIFF
--- a/connector.go
+++ b/connector.go
@@ -29,6 +29,8 @@ func (c *connector) Driver() driver.Driver {
 	return &Driver{}
 }
 
+var _ driver.Connector = &connector{}
+
 // NewConnector returns a connector for the pq driver in a fixed configuration
 // with the given name. The returned connect can be used to create any number
 // of equivalent Conn's. The returned connector is intended to be used with
--- a/connector.go
+++ b/connector.go
@@ -34,7 +34,7 @@ var _ driver.Connector = &connector{}
 // NewConnector returns a connector for the pq driver in a fixed configuration
 // with the given name. The returned connector can be used to create any number
 // of equivalent Conn's. The returned connector is intended to be used with
-// database/sql OpenDB.
+// database/sql.OpenDB.
 //
 // See https://golang.org/pkg/database/sql/driver/#Connector.
 // See https://golang.org/pkg/database/sql/#OpenDB.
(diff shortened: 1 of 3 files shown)


leighmcculloch:
The go1.10 builds are failing but I don't see any of the tests in those builds being stated as failed.

connector.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+// Connector represents a fixed configuration for the pq driver with a given
+// name. Connector satisfies the database/sql/driver Connector interface and
+// can be used to create any number of DB Conn's via the database/sql OpenDB
+// function.
+//
+// See https://golang.org/pkg/database/sql/driver/#Connector.
+// See https://golang.org/pkg/database/sql/#OpenDB.
+type connector struct {
```
mjibson: assert the interface: `var _ driver.Connector = &connector{}`

connector.go
```diff
@@ -0,0 +1,41 @@
...
+	return (&Driver{}).Open(c.name)
+}
+
+// Driver returnst the underlying driver of this Connector.
+func (c *connector) Driver() driver.Driver {
+	return &Driver{}
+}
+
+// NewConnector returns a connector for the pq driver in a fixed configuration
+// with the given name. The returned connect can be used to create any number
```
mjibson: connect -> connector

connector.go
```diff
@@ -0,0 +1,41 @@
...
+// of equivalent Conn's. The returned connector is intended to be used with
+// database/sql OpenDB.
```
mjibson: `database/sql.OpenDB`

connector.go
leighmcculloch: 691d77e

connector.go
leighmcculloch: b2ae663

connector.go
leighmcculloch: 3c42961

leighmcculloch:
@mjibson I've addressed your comments and fixed the references to the deprecated function. I also added an additional test. Build is now 💚.

PR #750, opened 2018-05-22
Add more instructions for running tests

leighmcculloch:
What
===
Add more instructions for running tests.

Why
===
For those starting out and not familiar with how to get a PostgresSQL
up and running for tests an example would be helpful. This isn't
supposed to define the only way to run the tests, but just provide an
example for those who have less experience and need a helping hand.

### DIFF
--- a/TESTS.md
+++ b/TESTS.md
@@ -1,0 +1,32 @@
+# Tests
+
+## Running Tests
+
+`go test` is used for testing.  A running PostgreSQL server is required, with
+the ability to log in.  The database to
+connect to test with is "pqgotest," on "localhost" but these can be overridden using
+[environment variables](https://www.postgresql.org/docs/9.3/static/libpq-envars.html).
+
+Example:
+
+	PGHOST=/run/postgresql go test
+
+## Benchmarks
+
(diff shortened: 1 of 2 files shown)


TESTS.md
```diff
@@ -0,0 +1,32 @@
+# Tests
+
+## Running Tests
+
+`go test` is used for testing.  A running PostgreSQL server is required, with
+the ability to log in.  The database to
```
mjibson: can you do something to fix this line wrapping? either fmt or one sentence per line. i don't care.

leighmcculloch: I'm happy to fix it however is needed, but it's not clear how it should be formatted based off the existing code. Right now I formatted it as wrap at 80 chars and I retained the two character spacing between sentences. How should I fix it?

mjibson:
```
`go test` is used for testing. A running PostgreSQL
server is required, with the ability to log in. The
database to connect to test with is "pqgotest," on
"localhost" but these can be overridden using [environment
variables](https://www.postgresql.org/docs/9.3/static/libpq-envars.html).
```

PR #752, opened 2018-05-25
Fix testing

hiroakis:
TestCloseBadConn did not use `PGHOST` and `PGPORT`. Running tests will failed, if these environment variables exist. Actually, I run tests on Docker machine. It failed.


QuLogic:
Looks like this was fixed by #905?

hiroakis:
@QuLogic Ah.. That's the same as this.

hiroakis:
closed

PR #755, opened 2018-06-01
Allow to use pq.Array(sliceOfSliceOfBytes)

esdrasbeleza:
I had some issues today to use `pq.Array(sliceOfSliceOfBytes)` to perform a multiple insertion with a single query to my database. I had to use a loop instead.
 
This code allows me to use `pq.Array` in my query. I'm not 100% sure if it's a bug, a missing feature or if there's a more clever way to do it.

### DIFF
--- a/array.go
+++ b/array.go
@@ -39,6 +39,8 @@ func Array(a interface{}) interface {
 		return (*Int64Array)(&a)
 	case []string:
 		return (*StringArray)(&a)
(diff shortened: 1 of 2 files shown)


arp242:
Looks like this was fixed some time ago via #876.

esdrasbeleza:
Oh, nice, that PR! 😂 
Happy New Year!

PR #759, opened 2018-06-11
Do not invoke sync command while preparing unnamed prepared statement

ikitiki:
Current implementation of the execution of the queries with arguments (i.e. non-simpleQueries) implicitly creates unnamed prepared statement followed by execute command; both done in the separate transactions (by invoking Sync after prepare).

This might be an issue if one uses a connection pooler in a transaction pool mode like [pgbouncer](https://github.com/pgbouncer/pgbouncer) or [odyssey](https://github.com/yandex/odyssey)

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -780,6 +780,7 @@ func decideColumnFormats(colTyps []fieldDesc, forceText bool) (colFmts []format,
 }
 
 func (cn *conn) prepareTo(q, stmtName string) *stmt {
+	var err error
 	st := &stmt{cn: cn, name: stmtName}
 
 	b := cn.writeBuf('P')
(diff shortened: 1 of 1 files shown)


mescam:
Hi,

is there any particular reason why this patch was not merged? We are experiencing issues described in #889 and after applying this patch it looks like the errors are gone. We are still investigating if this breaks anything else, but would be nice to hear from the maintainers.

cobolbaby:
@all Is there any progress on this request?

arp242:
How can this be tested? I added pgbouncer to the CI a few weeks ago, and all tests seem to work? I added `pool_mode=transaction` (locally), and this seems to work as well?

mkabilov:
I think at this point someone had already fixed the issue with connection poolers.

and in case it is not fixed, here is the workaround: use "binary_parameters=yes" as stated here: https://github.com/lib/pq/issues/889#issuecomment-644738440 or use https://github.com/jackc/pgx

PR #762, opened 2018-06-13
Fix server_version parsing

mkabilov:
Server version string can consist of 2 parts, not always 3

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1549,14 +1549,19 @@ func (cn *conn) processParameterStatus(r *readBuf) {
 	param := r.string()
 	switch param {
 	case "server_version":
-		var major1 int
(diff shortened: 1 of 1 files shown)


cheald:
This has implications for CopyIn usage - when the detected server version is less than 9.0, it falls back to a much slower Sprintf encoding. Merge would be appreciated!

My local startup parses:

server_version
10.5 (Ubuntu 10.5-0ubuntu0.18.04) => input does not match format

So the server version ends up being 0, which then falls back to the slower version.

Lekensteyn:
This solves the same issue as #979 but is more complex. I would suggest merging #979 and closing this one. What do you think?

PR #768, opened 2018-07-02
implement additional context specific sql interfaces

kylejbrock:


basvanbeek:
Currently context methods on Conn are provided in the conn_go18.go file. I think it's best if you move these stmt context methods there too. 

A next PR could potentially merge the conn*.go files as it seems minimum required version for lib/pq is Go 1.9.

kylejbrock:
@basvanbeek Are there any additional changes you'd like to see?

twpayne:
Any update on this? I tried to use a prepared COPY IN statement inside a context-aware transaction. When the context is canceled, `github.com/lib/pq` deadlocks (there's a race condition found by the Go race detector). The underlying reason seems to be that `github.com/lib/pq` doesn't have this PR merged :)

Code to reproduce the race condition is [here](https://gist.github.com/twpayne/f8e448f69944dbec1f05e2d3d7b3cd39).

prksu:
any update? i hope this PR to be merged

jalan:
Is there anything blocking this change?

It is surprising behavior that contexts are supported properly when using `DB.ExecContext` but not when using `Stmt.ExecContext`.

kylebrock:
🤷

KillerX:
I mean, I understand that the maintainers do not owe us anything, but it would sure be nice to hear back if we can do anything to get this merged!

It appears that there are several "forks" popping up mostly for the purpose of merging this patch, so it's somewhat splitting up the users, which seems like a bad idea...

mjibson:
I didn't really read it over in detail. Someone tell me if it breaks anything.

kjbass:
This breaks things


```
# github.com/lib/pq
../../lib/pq/conn_go18.go:213:21: not enough arguments in call to st.cn.cancel
	have ()
	want (context.Context)
```

kjbass:
Above snippet is the output of executing `go get github.com/lib/pq`

mjibson:
Thanks. Reverted.

kylejbrock:
This PR was from 1.5 yrs ago.  Give me a second and I'll updated it.

mjibson:
I didn't realize I hadn't re-run the CI tests which have changed a lot since then. Thanks for getting another one together.

kylejbrock:
@mjibson It's been a really long time since I've thought about any of this.

But, here is a PR that compiles: https://github.com/lib/pq/pull/921

Issue #688, opened 2017-12-13
copy.go: Missing field to add options


charlespare:
copy.go: Missing field to add options

According to the doc, when you use CopyIn you can add options at the end of the statement:
https://www.postgresql.org/docs/9.2/static/sql-copy.html

In my case I wanted to add options to handle null values. To fix it quickly I ended up writing a function that modify the statement before sending it to the database:

txn.Prepare(AddOptionToStmt(pq.CopyIn("column1", "column2", "column3")))

func AddOptionToStmt(inputStatement string) string {
    outputStatement := inputStatement + ` WITH NULL 'NULL'`
    return outputStatement
}

It would be nice to handle options in the code directly in the pq.CopyIn function by adding a parameter.

arp242:
CopyIn() and CopyInSchema() are just very simple string-concat query builders:

```
func CopyIn(table string, columns ...string) string {
    b := bytes.NewBufferString("COPY ")
    BufferQuoteIdentifier(table, b)
    makeStmt(b, columns...)
    return b.String()
}
```

You don't really need to use them: you can just use `copy` directly; these are identical:

    tx.Prepare(CopyIn("temp", "num", "text", "blob", "nothing"))

    tx.Prepare(`copy temp (num, text, blob, nothing) from stdin`)

It's not even that much shorter.

---

We can't really modify the existing functions because that would not be a compatible change. So we'd have to add new ones.

I think I will mark CopyIn() and CopyInSchema() and deprecated and update the documentation to not mention it.

PR #769, opened 2018-07-06
copy support null string field.

as0x:
英文水平不好，请见谅。


jsmestad:
@as0x looks like you have some minor formatting things you need to change in order for CI to pass.

arp242:
I would rather have a generic way to add options, rather than creating a new function. There's a bunch of options one would want to add.

That's #688

PR #775, opened 2018-08-01
Add Equal method for comparing NullTimes

jbarciauskas:


arp242:
`pq.NullTime` is now a type alias for the identical `sql.NullTime`, so this is no longer relevant.

PR #780, opened 2018-08-29
Add support for net.IP for inet type

Tsuki:


ip.go
```diff
@@ -0,0 +1,35 @@
...
+		return nil
+	case nil:
+		*a = nil
+		return nil
+	}
+
+	return fmt.Errorf("pq: cannot convert %T to net.IP", src)
+}
+
+func B2S(bs []uint8) string {
```
rogerhub: this can be simplified to just `string(bs)`

Tsuki:
I thank using type alias is much better than embedding

PR #807, opened 2018-11-20
added fix for Redshift COPY command affected rows

g8y3e:
Redshift COPY didn't return inserted rows in normal way.


arp242:
This needs a better description of what problem this fixes, as well as a test.

Since this is a few years old I'll just go ahead and close it, but feel free to come back with more details if you're still having problems and we can re-open.

PR #815, opened 2018-12-05
Lib master

bstein-clever:


bstein-clever:
Sorry, not the merge I wanted to do here. :(

PR #818, opened 2018-12-16
Support inline SSL certificates

eirslett:
> Presently, pq only supports SSL connections by
loading PEM certificates from files on disk.
There are some situations (for example integration
with HashiCorp Vault) where it's not so feasible
to load certificates from a file system, but better
to store them in-memory.
>
>This patch lets you set ?sslinline=true in the
connection string, which changes the behavior of
the paramters sslrootcert, sslcert and sslkey, so
they contain the contents of the certificates
directly, instead of file names pointing to the
certificates on disk.

What do you think about a change like this?
My specific use case is adding SSL support to [Vault](https://www.vaultproject.io/)'s integration with PostgreSQL - today it can only communicate over a plaintext socket. However, Vault doesn't store secrets/certificates directly on disk, but rather on an encrypted, distributed, pluggable backend system, which is the reason for this feature request: Vault can read the certificates from wherever they are stored, and pass them into pq via the connection string.

### DIFF
--- a/url.go
+++ b/url.go
@@ -40,10 +40,10 @@ func ParseURL(url string) (string, error) {
 	}
 
 	var kvs []string
-	escaper := strings.NewReplacer(` `, `\ `, `'`, `\'`, `\`, `\\`)
+	escaper := strings.NewReplacer(`'`, `\'`, `\`, `\\`)
 	accrue := func(k, v string) {
 		if v != "" {
-			kvs = append(kvs, k+"="+escaper.Replace(v))
+			kvs = append(kvs, k+"='"+escaper.Replace(v)+"'")
 		}
 	}
 
--- a/ssl.go
+++ b/ssl.go
@@ -59,6 +59,9 @@ func ssl(o values) (func(net.Conn) (net.Conn, error), error) {
 		return nil, err
 	}
 
+	// This pseudo-parameter is not recognized by the PostgreSQL server, so let's delete it after use.
+	delete(o, "sslinline")
+
 	// Accept renegotiation requests initiated by the backend.
 	//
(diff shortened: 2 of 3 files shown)


url.go
```diff
@@ -40,10 +40,10 @@ func ParseURL(url string) (string, error) {
 	}
 
 	var kvs []string
-	escaper := strings.NewReplacer(` `, `\ `, `'`, `\'`, `\`, `\\`)
+	escaper := strings.NewReplacer(`'`, `\'`, `\`, `\\`)
 	accrue := func(k, v string) {
 		if v != "" {
-			kvs = append(kvs, k+"="+escaper.Replace(v))
+			kvs = append(kvs, k+"='"+escaper.Replace(v)+"'")
```
eirslett: This change was made because PEM certificates include newlines, equal signs and various stuff that is tricky to escape - but by adding single quotes, we can work around the problem.

eirslett:
I'm not sure why the Travis build fails... looks like a flaky test, maybe? Could somebody please trigger a new build?

umran:
I have the exact same use case, which is what led me to this PR and the PR you submitted to Vault: https://github.com/hashicorp/vault/pull/5963. I really hope this gets reviewed at some point.

binlab:
@mjibson could you please look at this PR? Would be nice to have the possibility to use inline certificates. Hope for your help, thanks!

mjibson:
Can you or someone rebase this on master so the tests re-run? Open a new PR and mention me if needed.

eirslett:
2 years already... time flies! I rebased it on top of master now. 🤞

binlab:
@mjibson @eirslett thanks for making it possible!

PR #837, opened 2019-03-14
pq: support returning multiple result sets from a single query

kardianos:
Tested with Cockroach v2.1.6.


kardianos:
Tested with: https://github.com/kardianos/crdb-test using a replace directive in the go.mod.

mjibson:
The code LGTM. Do you think you could add a simple test to this PR? A single query with two result sets is fine.

kardianos:
I've updated the PR to include a test. Thanks.

PR #838, opened 2019-03-15
fix(conn): properly inspect interfaces for nil values

### LINKED ISSUE #773: inserting `[]byte(nil)` to `bytea` is not treated as NULL
this is not a bug report nor a feature request, just a question.

when I try to insert `[]byte(nil)` as `NULL` like this, it was treated as 0 length `bytea`.
(you can see full source code at https://play.golang.org/p/6WBCgbtJKYS)

```go
for _, b := range [][]byte{
	[]byte("hello"),
	[]byte("world"),
	nil, // this won't be stored as NULL.
} {
	_, err := db.Exec(`INSERT INTO foo1 (data) VALUES($1)`, b)
	if err != nil {
		return err
	}
}
```

Is this handling of `[]byte(nil)` intentional?

epelc:
Fixes #773 by further inspecting arguments to check if they are nil interfaces or slices.

Sending a nil byte slice use to not work because it was a nil value of type `[]byte` which `!= nil` hence it would really send an empty byte slice. This prevented you from setting columns to null unless you had a different path to send an explicit nil which also had no type(confusing and a pain to implement). Checking interface values equal nil is a common bug in go which usually requires using the reflect package to further inspect the interface's actual value is nil since the typing usually isn't.

This does slightly slow down writes theoretically but it should be worth it for such a common use case and standard feature of sending nil.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -14,6 +14,7 @@ import (
 	"os/user"
 	"path"
 	"path/filepath"
+	"reflect"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


joeandaverde:
I have a similar issue. What's the best way to go since this PR has been open over a year?

PR #839, opened 2019-03-21
pq: pipe the context through to the dialer, make use of the Connector

### LINKED ISSUE #783: Connector context not passed into dial
The connector interface is satisfied, but the context is not used when dialing. See related issue:

https://github.com/golang/go/issues/27476

kardianos:
Make minimal changes to enable using the Connector for real.
This sets the stage to allow setting custom properties on the
Connector (like DialerContext) or other userful properties
that don't fit into a dsn.

Fixes #783

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -2,6 +2,7 @@ package pq
 
 import (
 	"bufio"
+	"context"
 	"crypto/md5"
 	"database/sql"
 	"database/sql/driver"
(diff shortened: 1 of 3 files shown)


kardianos:
cc @mjibson

go.mod
```diff
@@ -1 +1,3 @@
 module github.com/lib/pq
+
+go 1.12
```
mjibson: This probably shouldn't go in this change.

kardianos: Removed.

PR #842, opened 2019-03-29
Add Detail to Error()

### LINKED ISSUE #679: Add Detail to Error()
As is, Error() only returns the error Message, which means most applications swallow the error details which are sometimes very important.

dbresson:
Resolves #679


arp242:
PostgreSQL assumes it's printed on a separate line:

    (0)=# create schema pg_xxx;
    ERROR:  unacceptable schema name "pg_xxx"
    DETAIL:  The prefix "pg_" is reserved for system schemas.

I think the "detail" looks quite ugly like this:

    pq: unacceptable schema name "pg_xx": The prefix "pg_" is reserved for system schemas.

This is a relatively simple message, but some are quite long.

What I'll do is rewrite #521 to provide more detailed messages via `ErrorDetailed()` something, leaving the regular Error() method as it is now.

dbresson:
@arp242 The whole point was to include the necessary detail in Error() because some of the errors are completely useless without that detail.

It seems pretty silly to prioritize cosmetics over utility in an error message.

arp242:
I had a look at the PostgreSQL source, and most errors did not seem appropriate to add to the Error() string. But we can make some selected exceptions for some errors, maybe.

dbresson:
Doing so selectively would be a reasonable solution.

Too bad I didn't include the specific error that prompted me to open the ticket as an example of why.

PR #843, opened 2019-04-02
Fix connection leak on SSL failure

### LINKED ISSUE #840: Connection overflow pq: SSL is not enabled on the server
Hello, thanks for the great lib! 
We are having some trouble on windows when there is no ssl enabled on postgresql..
If I add the  ?sslmode=disable the library works ok, 
If I left the lib returning the error message,  **pq: SSL is not enabled on the server**, db.Closes does not work and the connections keep rising. Take a look.... 

db.SetMaxOpenConns(10) //DOES NOT WORK
db.SetMaxIdleConns(5) //DOES NOT WORK

Untitled

```

db, err := sql.Open("postgres", m.HostData().URI)
	if err != nil {
		logger.Error(err)
		reporter.Error(err)
		return
	}

	// Set the maximum number of concurrently open connections to 5. Setting this
	// to less than or equal to 0 will mean there is no maximum limit (which
	// is also the default setting).
	db.SetMaxOpenConns(10) //DOES NOT WORK
	db.SetMaxIdleConns(5) //DOES NOT WORK
	defer db.Close()

results, err := postgresql.QueryStats(db, "SELECT * FROM pg_stat_bgwriter") //DO A DB.QUERY
	if err != nil {
		err = errors.Wrap(err, "QueryStats")
		logger.Error(err)
		reporter.Error(err)
		return //WILL EXIT HERE AND CALL db.Close()
	}
```

adriansr:
When connecting to a server using SSL fails (i.e. with SSL disabled) the connection was never freed.

Fixes #840

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -299,6 +299,7 @@ func (c *Connector) open(ctx context.Context) (cn *conn, err error) {
 
 	err = cn.ssl(o)
 	if err != nil {
+		cn.c.Close()
 		return nil, err
 	}
 
--- a/conn.go
+++ b/conn.go
@@ -299,7 +299,9 @@ func (c *Connector) open(ctx context.Context) (cn *conn, err error) {
 
 	err = cn.ssl(o)
 	if err != nil {
-		cn.c.Close()
+		if cn.c != nil {
+			cn.c.Close()
+		}
 		return nil, err
 	}
 


mjibson:
This appears to fail some tests. Can you fix those?

PR #845, opened 2019-04-08
Fix: prevent unexpected connection info string

Keijun-KUMAGAI:
I manage the database connection info as a toml file, and my connection code is something like

```
[psql]
  dbname = "my_database"
  host   = "localhost"
  port   = 5432
  user   = "keijun"
  pass   = ""
  blacklist = ["schema_migrations"]
  sslmode= "disable"
```

```
type Config struct {
	Psql psqlConfig
}

type psqlConfig struct {
	Dbname string `toml:"dbname"`
	Host   string `toml:"host"`
	Port   int    `toml:"port"`
	User   string `toml:"user"`
	Pass   string `toml:"pass"`
}

var config Config
_, err = toml.DecodeFile("./sqlboiler.toml", &config)

psqlInfo := fmt.Sprintf("host=%s port=%d user=%s password=%s dbname=%s sslmode=disable",
	config.Psql.Host, config.Psql.Port, config.Psql.User, config.Psql.Pass, config.Psql.Dbname,
)

DB, err = sql.Open("postgres", psqlInfo)
rows, err := DB.Query(`SELECT * FROM users`)

```

However, if the password info was not provided or "", the driver give me a error.

```
panic: pq: database "keijun" does not exist
```

I know this is because password is blank and following part (in this case "dbname") is handled as a password value, and insted of that, the username was used as a database name.

this is a expected behaive as I could see from the test code, like
```
{"user= password=foo", values{"user": "password=foo"}, true},
```
but I thought checking if the value contains "=" and if so, add new key is better and make this PR.

This is my first time to make PR so maybe I am saying stupid things, so if you find anything strange please let me know, thanks in advance

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -455,6 +455,9 @@ func parseOpts(name string, o values) error {
 						return fmt.Errorf(`missing character after backslash`)
 					}
 				}
+				if r == '=' {
+					return fmt.Errorf(`missing character after "=" in %q connection info string`, string(keyRunes))
+				}
 				valRunes = append(valRunes, r)
 
 				if r, ok = s.Next(); !ok {
--- a/conn.go
+++ b/conn.go
@@ -455,9 +455,6 @@ func parseOpts(name string, o values) error {
 						return fmt.Errorf(`missing character after backslash`)
 					}
 				}
-				if r == '=' {
-					return fmt.Errorf(`missing character after "=" in %q connection info string`, string(keyRunes))
-				}
 				valRunes = append(valRunes, r)
 
 				if r, ok = s.Next(); !ok {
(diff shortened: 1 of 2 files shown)


arp242:
Seems to work in a quick test?

```
func TestXXX(t *testing.T) {
    db, err := sql.Open("postgres", "host=localhost user=pqgo password= dbname=pqgo sslmode=disable")
    if err != nil {
        t.Fatal(err)
    }

    rows, err := db.Query(`select 'zxc'`)
    if err != nil {
        t.Fatal(err)
    }
    rows.Next()
    var s string
    err = rows.Scan(&s)
    fmt.Println(err, s)
}
```

Outputs `<nil> zxc`.

arp242:
As near as I can tell this behaves correctly, so either this has already been fixed since this PR was opened or one of us is mistaken. Closing for now.

Issue #534, opened 2016-11-07
Implement Named Parameters


itsjamie:
The addition of named parameters support in Go 1.8 is documented in #531, tracking this individual change here.

Notes from the document:
```
Drivers should be clear in their documentation what form of names they accept. For example, does the driver support "@ID", "?ID", ":ID", "$ID", "ID". If no placeholder is specified and the query language requires one, will the driver prepend the default symbol to the name if none are found? This behavior should be documented in the driver.

If mixing named parameters with ordinal position, the ordinal position is always assigned to, values don't skip named parameter positions. For example: "... where ID = :ID and Name = ?", the second placeholder ("?") will be ordinal position 2.
```

I would like to implement this in lib/pq, but figured some discussion would be good. For example, should `lib/pq` require a specific prefix? I would imagine no because they will be reduced down to `$1` and `$2` ordinal operators.

I would still want to implement named parameters though so that any shared names in a query can be automatically reduced to the same ordinal parameter.

Thoughts?

nathany:
Looks like PostgreSQL doesn't have any special support for named parameters in prepared statements, so is the idea that this would be implemented entirely in the driver?

`sql.Named("Name", value)` substitutes `:Name` for `$1` in SQL, which receives the value.

Not 100% sure if it's worthwhile at the driver level, since it can also be done at a higher level: http://jmoiron.github.io/sqlx/#namedParams

PostgreSQL has a named notation when calling functions:
https://www.postgresql.org/docs/current/static/sql-syntax-calling-funcs.html
`a => 'Hello'` this syntax appears to be 9.5+
`a := 'Hello'` this syntax is still supported

Not sure if it provides any benefit over passing $1 over to PostgreSQL if the SQL looks the same either way.

cbandy:
> should `lib/pq` require a specific prefix? I would imagine no because they will be reduced down to `$1` and `$2` ordinal operators.

I don't understand what you're asking here.

> I would still want to implement named parameters though so that any shared names in a query can be automatically reduced to the same ordinal parameter.

The parameter symbols/markers `$n` can be used more than once in a query to refer to the same parameter value/argument. Is that what you mean by "shared?" Perhaps I'm misunderstanding what you desire here.

For example, the following query multiplies a number by three:

```go
err := db.QueryRow(`SELECT $1 + $1 + $1`, x).Scan(&y)
```

itsjamie:
When speaking to the prefix, I was referencing the google document for `database/sql` 1.8 changes.

[...]

PR #846, opened 2019-04-11
pq: simple support for named parameters

jeromedoucet:
what:
Add named parameters support for "conn". It works basically like ordinals
parameters, but with an alphabetical label prefixed with ":"

Example:

Select $1, $2, $1; may become Select :firstName, :lastName, :firstName

Why:
SQL queries may be difficult to maintain, especially when the complexity
is growing. Ordinal parameters are great, but may not be meaningfull for
the one who is trying to understand the query.

The goal here is not to have something full featured and complex : pq is
"just" a driver, but having a thin layer, simple to maintain, that will
fit 90% of the needs.

Limitations:
Its simplicity means that there is many limitations :

First of all, using mixed types of parameters is not
allowed. Man can not use ordinal parameters and named
parameters is the same query.

Refs #534

### DIFF
--- a/conn_go18.go
+++ b/conn_go18.go
@@ -12,12 +12,23 @@ import (
 
 // Implement the "QueryerContext" interface
 func (cn *conn) QueryContext(ctx context.Context, query string, args []driver.NamedValue) (driver.Rows, error) {
-	list := make([]driver.Value, len(args))
-	for i, nv := range args {
-		list[i] = nv.Value
+	var err error
+	var wrapper *requestWrapper
+	var list []driver.Value
+	var r *rows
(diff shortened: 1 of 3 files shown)


Congee:
any blocker?

arp242:
I'm quite hesitant to use a regexp-based solution for this, as per my comment in https://github.com/lib/pq/issues/534#issuecomment-3702010998

While this at least avoids the `col::text` problem that sqlx has, it will also replace things in comments and string literals, which is not really correct and may result in weird/confusing logs and pg_stat_statements at best, and malformed SQL at worst.

PR #852, opened 2019-04-23
Add `QuoteLiteral` function.

jkatz:
`QuoteLiteral` performs a similar function to `QuoteIdentifier`: dynamically, but **safely**, pass literals into SQL statements and avoid SQL injections.

There are statements, e.g. DDL statements like `CREATE ROLE`, that accept literals in them, and there may be situations where a user may want to dynamically pass in parameters. You cannot pass parameters directly to DDL statements, so`QuoteLiteral` allows a user to do so and by following a similar methodology to `QuoteIdentifier`.

This patch includes both `QuoteLiteral` and a series of tests similar to what is found with `QuoteIdentifier`.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1500,6 +1500,39 @@ func QuoteIdentifier(name string) string {
 	return `"` + strings.Replace(name, `"`, `""`, -1) + `"`
 }
 
+// QuoteLiteral quotes a 'literal' (e.g. a parameter, often used to pass literal
+// to DDL and other statements that do not accept parameters) to be used as part
+// of an SQL statement.  For example:
+//
+//    exp_date := pq.QuoteLiteral("2023-01-05 15:00:00Z")
+//    err := db.Exec(fmt.Sprintf("CREATE ROLE my_user VALID UNTIL %s", exp_date))
+//
+// Any single quotes in name will be escaped. Any backslashes (i.e. "\") will be
+// replaced by two backslashes (i.e. "\\") and the C-style escape identifier
+// that PostgreSQL provides ('E') will be prepended to the string.
+func QuoteLiteral(literal string) string {
+	// This follows the PostgreSQL internal algorithm for handling quoted literals
+	// from libpq, which can be found in the "PQEscapeStringInternal" function,
+	// which is found in the libpq/fe-exec.c source file:
+	// https://git.postgresql.org/gitweb/?p=postgresql.git;a=blob;f=src/interfaces/libpq/fe-exec.c
+	//
+	// substitute any single-quotes (') with two single-quotes ('')
+	literal = strings.Replace(literal, `'`, `''`, -1)
+	// determine if the string has any backslashes (\) in it.
+	// if it does, replace any backslashes (\) with two backslashes (\\)
+	// then, we need to wrap the entire string with a PostgreSQL
+	// C-style escape. Per how "PQEscapeStringInternal" handles this case, we
+	// also add a space before the "E"
+	if strings.Contains(literal, `\`) {
+		literal = strings.Replace(literal, `\`, `\\`, -1)
+		literal = ` E'` + literal + `'`
+	} else {
+		// otherwise, we can just wrap the literal with a pair of single quotes
+		literal = `'` + literal + `'`
+	}
+	return literal
+}
+
 func md5s(s string) string {
 	h := md5.New()
 	h.Write([]byte(s))
--- a/conn_test.go
+++ b/conn_test.go
@@ -1554,6 +1554,38 @@ func TestQuoteIdentifier(t *testing.T) {
 	}
 }
 
+func TestQuoteLiteral(t *testing.T) {
+	var cases = []struct {
+		input string
+		want  string
+	}{
+		{`foo`, `'foo'`},
+		{`foo bar baz`, `'foo bar baz'`},
+		{`foo'bar`, `'foo''bar'`},
+		{`foo\bar`, ` E'foo\\bar'`},
+		{`foo\ba'r`, ` E'foo\\ba''r'`},
+		{`foo"bar`, `'foo"bar'`},
+		{`foo\x00bar`, ` E'foo\\x00bar'`},
+		{`\x00foo`, ` E'\\x00foo'`},
+		{`'`, `''''`},
+		{`''`, `''''''`},
+		{`\`, ` E'\\'`},
+		{`'abc'; DROP TABLE users;`, `'''abc''; DROP TABLE users;'`},
+		{`\'`, ` E'\\'''`},
+		{`E'\''`, ` E'E''\\'''''`},
+		{`e'\''`, ` E'e''\\'''''`},
+		{`E'\'abc\'; DROP TABLE users;'`, ` E'E''\\''abc\\''; DROP TABLE users;'''`},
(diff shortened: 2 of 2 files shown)


jkatz:
The test that is failing appears to be independent of this commit:

https://travis-ci.org/lib/pq/jobs/523709631

cbandy:
Related to #718.

jkatz:
@cbandy Looking at #718 and the behavior of the PostgreSQL, there are a few scenarios to look at. Following examples in the [PostgreSQL docs](https://www.postgresql.org/docs/current/functions-string.html)

```sql
SELECT quote_literal(E'O\'Reilly');
```
yields
```bash
 quote_literal 
---------------
 'O''Reilly'
```

but:

```sql
SELECT quote_literal($$O\'Reilly$$);
```

yields:

```bash
 quote_literal  
----------------
 E'O\\''Reilly'
```

So the check in #718 to see if a `\` exists in the string, and if so prepend the escape is not necessarily accurate. The first thing to check would be to see if the string was prepended with `E` (or `e`). If prepended with either, then do not check for `\`, o/w check for `\` and prepend.

Happy to incorporate work from said patch.

jkatz:
@cbandy ...so I decided to go to the source of truth, being the internal `quote_literal` in the [PostgreSQL source](https://git.postgresql.org/gitweb/?p=postgresql.git;a=blob;f=src/backend/utils/adt/quote.c;h=5bfa327d50a693ba7cfcab0a8f4c2203e3bfdef4;hb=HEAD#l46):

It's actually a bit more complicated than what appears on the surface (and perhaps one of the times you can do things more concisely in C 😄 ). Basically, to match the behavior of `quote_literal` I perform the following:

#  Check to see if this is already a C-escaped string, i.e. it starts with `E` or `e`

## If it is a C-style escaped String

1. Determine if there are any escaped backslashes `\\` -- if there are, leave it as a C-escaped string, otherwise, remove the first character (i.e. `E` or `e`).

2. Substitute any `'` (which will be represented as `\'` for `''`

3. The literal is either wrapped with `E''` or `''` at this point.

## If it is not a C-style escaped string

1. First, turn any `'` into `''`

2. Then, determine if there are any `\`. If there are, substitute them to be `\\` and indicate this is now a C-style escape string (i.e. add a `E` to the beginning and wrap with `''`). Otherwise, just wrap with `''`

And that should do it.

I added some more tests, specifically adding some of the typical "SQL injections" one might find when trying to use an unquoted literal.

conn_test.go
```diff
@@ -1550,6 +1550,36 @@ func TestQuoteIdentifier(t *testing.T) {
...
+		{`foo'bar`, `'foo''bar'`},
+		{`foo\bar`, `E'foo\\bar'`},
+		{`foo\ba'r`, `E'foo\\ba''r'`},
+		{`foo"bar`, `'foo"bar'`},
+		{`foo\x00bar`, `E'foo\\x00bar'`},
+		{`\x00foo`, `E'\\x00foo'`},
+		{`''`, `''''''`},
+		{`'abc'; DROP TABLE users;`, `'''abc''; DROP TABLE users;'`},
+		{`\'`, `E'\\'''`},
+		{`E'\''`, `''''`},
```
cbandy: These cases surprise me. I would prefer that `QuoteLiteral("E''") == "'E'''''"`. That is, a Go string consisting of three ASCII characters is interpreted by PostgreSQL as a literal containing three ASCII characters.

conn.go
```diff
@@ -1500,6 +1500,50 @@ func QuoteIdentifier(name string) string {
...
+//
+//    exp_date := pq.QuoteLiteral("2023-01-05 15:00:00Z")
+//    err := db.Exec(fmt.Sprintf("CREATE ROLE my_user VALID UNTIL %s", exp_date))
+//
+// Any single quotes in name will be escaped. Any string that contains a backslash
+// (i.e. "\")
+func QuoteLiteral(literal string) string {
+	// This follows the PostgreSQL internal algorithm for handling quoted literals
+	// which is found here:
+	// https://git.postgresql.org/gitweb/?p=postgresql.git;a=blob;f=src/backend/utils/adt/quote.c;hb=HEAD#l46
```
cbandy: We should be looking at the `libpq` implementations for inspiration, I think?

[PQescapeLiteral](https://github.com/postgres/postgres/blob/1f39a1c0641531e0462a4822f2dba904c5d4d699/src/interfaces/libpq/fe-exec.c#L3564-L3568):
```c
char *
PQescapeLiteral(PGconn *conn, const char *str, size_t len)
{
	return PQescapeInternal(conn, str, len, false);
}
```

[PQescapeInternal](https://github.com/postgres/postgres/blob/1f39a1c0641531e0462a4822f2dba904c5d4d699/src/interfaces/libpq/fe-exec.c#L3437-L3444):
```c
/*
 * Escape arbitrary strings.  If as_ident is true, we escape the result
 * as an identifier; if false, as a literal.  The result is returned in
 * a newly allocated buffer.  If we fail due to an encoding violation or out
 * of memory condition, we return NULL, storing an error message into conn.
 */
static char *
PQescapeInternal(PGconn *conn, const char *str, size_t len, bool as_ident
```

There, I see what looks like the bottom half of the function implemented in this PR:

1. Repeat any single quotes.
2. Repeat any backslashes.
3. Indicate an escape string when there are any backslashes.

What's missing here is the whitespace before the escape string's `E`. (I understand the English of the comment that justifies it, but ... 🤷‍♂️) We should do that too because, as I recall, the goal of this module is least-surprise compared to `libpq`.

Introduced in 9.0. The related thread: (maybe) https://www.postgresql.org/message-id/flat/162867790912290857m340ab1f1s76ef48c6e7241398%40mail.gmail.com

jkatz: Yeah, I agree with the above :-) I'll take a look at that + the libpq code, revise, and revise the tests as well.

jkatz:
@cbandy Pushed a revision. I've incorporated all of your feedback. This follows `PQescapeStringInternal` though down the path of escaping literals. It does include the behavior for putting a space before any C-style escaped string, so `\` => ` E'\\'`.

Tests are updated to reflect this.

jkatz:
Hi - are there any other things that need to be done in order for this to be committed? Open to other feedback or suggestions. Thanks!

cbandy:
@mjibson any thoughts?

conn.go
```diff
@@ -1500,6 +1500,38 @@ func QuoteIdentifier(name string) string {
...
 
+// QuoteLiteral quotes a 'literal' (e.g. a parameter, often used to pass literal
+// to DDL and other statements that do not accept parameters) to be used as part
+// of an SQL statement.  For example:
+//
+//    exp_date := pq.QuoteLiteral("2023-01-05 15:00:00Z")
+//    err := db.Exec(fmt.Sprintf("CREATE ROLE my_user VALID UNTIL %s", exp_date))
+//
+// Any single quotes in name will be escaped. Any string that contains a backslash
+// (i.e. "\")
```
mjibson: Can you fix this sentence? It doesn't have a verb or period.

jkatz: @mjibson Thanks for the feedback! This has been addressed:

https://github.com/lib/pq/pull/852/files#diff-a2ddca1600f34abb91244424f8431c2cR1511

I also rebased against master.

PR #857, opened 2019-05-03
encode: use 64-bit parsing for float32s

jordanlewis:
Previously, float4 values were incorrectly parsed by this library. For
example, a value of `1.2::float4` would be returned to the user as a
float64 of `1.2000000476837158`. Now, float4s are parsed correctly,
using the 64-bit-float string parsing facility, and return 1.2 as
expected. Note that this problem only showed up if you used the returned
float64 value from lib/pq directly - casting to a float32 would make the
problem stay hidden - but since lib/pq returns float64, we should make
an effort for that value to be correct anyway.

This was found during investigation here: https://github.com/cockroachdb/cockroach/issues/36811

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -117,11 +117,10 @@ func textDecode(parameterStatus *parameterStatus, s []byte, typ oid.Oid) interfa
 		}
 		return i
 	case oid.T_float4, oid.T_float8:
-		bits := 64
-		if typ == oid.T_float4 {
-			bits = 32
-		}
-		f, err := strconv.ParseFloat(string(s), bits)
+		// We always use 64 bit parsing, regardless of whether the input text is for
+		// a float4 or float8, because clients expect float64s for all float datatypes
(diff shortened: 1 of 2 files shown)


jordanlewis:
@mjibson could you PTAL?

knz:
LGTM

encode.go
```diff
@@ -117,11 +117,10 @@ func textDecode(parameterStatus *parameterStatus, s []byte, typ oid.Oid) interfa
 		}
 		return i
 	case oid.T_float4, oid.T_float8:
-		bits := 64
-		if typ == oid.T_float4 {
-			bits = 32
-		}
-		f, err := strconv.ParseFloat(string(s), bits)
+		// We always use 64 bit parsing, retardless of whether the input text is for
```
mjibson: "regardless"

jordanlewis: Done

PR #859, opened 2019-05-06
support google app engine

tamuhey:
Thanks for this great repo!

Google App Engine(GAE) refuse `import "syscall"`, so you cannot build with the [user_windows.go](https://github.com/lib/pq/blob/ceb88a0649027d76542fd8084fcb4e07b4d52295/user_windows.go#L6).
If we are on GAE, exclude "user_windows.go" by build constraint, and use the newly added "user_notwindows.go" instead.


cbandy:
Which App Engine environment are you using (standard or flexible) and which version of Go?

I'm surprised that `user_windows.go` is part of your build. As I understand it, this file [implicitly contains `// +build windows`](https://golang.org/pkg/go/build/#hdr-Build_Constraints).

Perhaps this PR should add `appengine` to `user_posix.go`, instead?
https://github.com/lib/pq/blob/ceb88a0649027d76542fd8084fcb4e07b4d52295/user_posix.go#L3

tamuhey:
@cbandy standard.

How to reproduce:

1. clone the example : https://github.com/GoogleCloudPlatform/golang-samples/tree/master/appengine/helloworld
2. add `import _ "github.com/lib/pq"`
3. `$dev_appserver.py app.yaml

cbandy:
Which version of Go?

tamuhey:
@cbandy maybe 1.9.

The reason I wrote "maybe" is that `dev_appserver.py` runs the following command to build the project.

``` C:\Users\ooo\AppData\Local\Google\Cloud SDK\google-cloud-sdk\platform\google_appengine\goroot-1.9\bin\go-app-builder.exe```

tamuhey:
I simply added `appengine` to `user_posix.go`, but the problem is not solved.
But I added `// +build !appengine` to `user_windows.go`, solved.

Issue #863, opened 2019-05-13
Support for GSSAPI auth


jake-scott:
I would like to add Kebreros (GSSAPI) authentication support to library.  I have a patch that I will submit as a PR, but interested in feedback before this is considered ready for merge, as this is my first look at the driver code.

PR #864, opened 2019-05-13
Add Kerberos (GSSAPI) authentication support

jake-scott:
The patch supports the following new connection
  options:
    - service:  The Kerberos service name to use when constructing
                the SPN;  default: "postgres"
    - spn:      Specify the full SPN;  avoids construction on the
                fly from the host and service names

   In the future, this code should use SSPI on Windows, as it is
   not common to use MIT style credential caches on that platform.

   Fixes: #863

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -152,6 +152,9 @@ type conn struct {
 
 	// If not nil, notices will be synchronously sent here
 	noticeHandler func(*Error)
+
+	// GSSAPI context
+	gss Gss
 }
 
 // Handle driver-side settings in parsed connection string.
@@ -1066,7 +1069,10 @@ func isDriverSetting(key string) bool {
 		return true
 	case "binary_parameters":
 		return true
-
+	case "service":
+		return true
+	case "spn":
+		return true
 	default:
 		return false
 	}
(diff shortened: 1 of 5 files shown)


jheiss:
This appears to be missing a call to GetServiceTicket(). If I already have a service ticket for "postgres/<server name>" then my client successfully connects. But if I don't have a service ticket for the database then the connection fails.

jheiss:
Oh, but looking further InitSecContext is supposed to call GetServiceTicket. Huh, I wonder why that doesn't seem to be happening for me.

jake-scott:
Let me try to reproduce.  How did you configure this?.. did you let the module construct the credentials cache (/tmp/krb5cc_$(id -u))  -- or did you specify it with the KRB5CCNAME env var?  Did you specify 'spn' or 'service' as part of the connection options?

What about running:   kvno postgres/<hostname> at the command line?.. does that get you a ticket?

jheiss:
So far it looks like a bug in jcmturner/gokrb5 but I'm still working on tracking it down. It doesn't appear to be anything your code is doing wrong. I'm specifying my cred cache via KRB5CCNAME and am not specifying spn or service, the Kerberos library is constructing the SPN (correctly) on its own.

If you're more familiar with the gokrb5 code maybe you can confirm what I'm seeing, but it looks to me like this is passing `tgsRep.Ticket` when it should be passing `tgt`:

https://github.com/jcmturner/gokrb5/blob/master/client/TGSExchange.go#L17

By passing what appears to be the empty `tgsRep.Ticket` then way down in the call stack it blows up on me with a "runtime error: index out of range" when it tries to read into the empty ticket. Anyway, I'll keep poking at it. Thanks for submitting this and checking in, hopefully we can get this merged as it would be super helpful to me.

jake-scott:
Sorry its taken me so long to get around to testing this again.

It seems to be working for me with the current jcmturner/gokrb5 library.. I created a new test wrapper and posted it here :  `https://github.com/jake-scott/libpq-gss-test`

I override the import path of github.com/lib/pq in go.mod, to a new version that includes of the GSSAPI patch but re-based as of yesterday.

Could you give this test a go? It expects a database on the local machine called 'test' with a single  column 'a' (`CREATE TABLE test (a VARCHAR)`).

jheiss:
@jake-scott Your test code works for me.

jake-scott:
Good news!  And does it also work in your own test harness now  (I'm wondering if the later krb lib fixed it or if there is something different to the way you're using it to my test.. )

jake-scott:
Hi - FYI I re-worked this patch a little and also included Windows support.

sudeeptoroy:
Hi @jheiss @jake-scott,  I am interested in this PR. I tested it on my setup and found it working. It would be great if this PR is merged. Let me know if i can be of any help.

jake-scott:
OK I'm glad it worked for you.  It looks like I have a conflict to fix so I'll rebase and push a new version.  I'm not sure what to do about getting this merged TBH.  But I notice a notice on the readme that I hadn't before 'effective in maintenance mode and not actively developed' so I'm wondering if we should look at the referenced pgx driver instead..

sudeeptoroy:
Hi @jake-scott , my suggestion would be to ready this PR for merge.

sudeeptoroy:
Hi @jake-scott, Do you know by when you may be able to get this PR ready for merge. let me know if you need any help. This PR would be very helpful to many.

mjibson:
This works for me so I'm going to resolve and merge it.

mjibson:
Closed by #970

PR #866, opened 2019-05-15
remove unreachable code

andriikushch:
Error 'err' already checked for nil


kandros:
I checked that `sql.OpenDB` does not return an error to make sure it wasn't a mistake in the first place.
https://golang.org/pkg/database/sql/#OpenDB

Looks good to me

andriikushch:
Outdated

PR #868, opened 2019-05-21
Support connection service file and service names

### LINKED ISSUE #538: Implement Connection Service File
Currently the driver does not support specifying a connection using a service description.
Details: https://www.postgresql.org/docs/current/static/libpq-pgservice.html

I'm currently not a go programmer myself, however looking at the code I might be able to create a proof-of-concept implementation.

Basically, what needs to happen that iff a service is specified, we need to read the libpq parameters from the Connection Service File. For a start, we could use the same logic (and limitations) as the pgpass reading: https://github.com/lib/pq/blob/master/conn.go#L159-L168

Questions:
- Does anyone - other than me - think this is useful?
- The Connection Service File is in an ini format, should we use an external ini (or toml) library to implement this?

feikesteenbergen:
The whole point of supporting this can best be said by directly quoting
the PostgreSQL manual:

> The connection service file allows libpq connection parameters to be
> associated with a single service name. That service name can then be
> specified by a libpq connection, and the associated settings will be
> used. This allows connection parameters to be modified without
> requiring a recompile of the libpq application. The service name can
> also be specified using the PGSERVICE environment variable.

source: https://www.postgresql.org/docs/current/libpq-pgservice.html

Fixes #538


ribasushi:
@rafiss ping on this PR? Seems pretty decent.

The alternative today is to use something like this, which I have to admit is pretty gross... https://github.com/behrang/go-pgservice

arp242:
Merged via #1285, as re-implementing this from scratch was easier than solving the conflicts on this 😅

Issue #870, opened 2019-05-27
db operation fails with "broken pipe" instead of reconnecting transparently after server restart


fho:
I'm using lib/pq v1.1.1, go 1.12.5, Linux 4.19.45-1-lts.

I have a db handler on that I run >=1 operations,  restart the db-server and wait until the startup finished, the next query fails **if** all the previous TCP connections of the Operation System to the database were closed in the meantime (`ss -na | grep 5432'` shows nothing).
The operation fails with: `write tcp [::1]:45676->[::1]:5432: write: broken pipe`.
If another query is done after the failed one, it succeeds.

I expect that the `db.Exec()` query succeeds after the postgresql restart finished and the sql package or pq driver retries and reconnects transparently  if needed in the background.

If the postgresql-server restart happens quickly while they are still TCP connections in `FIN-WAIT-2` or another state, the db operations after the postgres restart succeeds.

-----

How to reproduce:
- run `docker run -p 5432:5432 postgres:latest` 
- run in another terminal https://gist.github.com/fho/777ccc77971612f3659cbdf5cef27ede 
  pass as command-line argument `postgresql://postgres@localhost?sslmode=disable`
- Press `q` to do a sql query
- Shutdown the postgres server by pressing `ctrl + c` in the terminal
- check `watch -n 0.5  "sh -c 'ss -na |grep 5432"`, wait until the TCP connections vanished
- start postgres again: run `docker run -p 5432:5432 postgres:latest`
- press `q` in the terminal that runs the go program to trigger another query => it fails

------

My first idea for a fix was to return `ErrBadConn` on broken pipe errors but like discussed in https://github.com/lib/pq/issues/422 this has the issue that operations might be redone.

The mysql driver seems to solve it by having a custom error type to indicate retryable connection errors.
If a Write() on the tcp socket failed before  a whole SQL statements was send, it's safe to retry the operation.
The caller of `conn.send()` could decide it and set error to `ErrBadConn`.

dharmjit:
Hi, I am also getting random broken pipe errors. Any update on this issue.

fho:
@dharmjit no, nothing new regarding the issue.
You can use my fix from my PR: https://github.com/lib/pq/pull/871

nullbio:
[...]

PR #871, opened 2019-05-28
allow retry on netErrors in safe situations

fho:
In some situations the sql package does not retry a pq operation when it
should.
One of the situations is https://github.com/lib/pq/issues/870.
When a postgresql-server is restarted and after the restart is finished an
operation is triggered on the same db handle, it failed with an broken
pipe error in some circumstances.

The sql package does not retry the operation and instead fail because
the pq driver does not return driver.ErrBadConn for network errors.
The driver must not return ErrBadConn when the server might have already
executed the operation. This would cause that sql package is retrying it
and the operation would be run multiple times by the postgresql server.
In some situations it's safe to return ErrBadConn on network errors.
This is the case when it's ensured that the server did not receive the
message that triggers the operation.

This commit introduces a netErrorNoWrite error. This error should be
used when network operations panic when it's safe to retry the
operation.
When errRecover() receives this error it returns ErrBadConn() and marks
the connection as bad.
A mustSendRetryable() function is introduced that wraps a netOpError in
an netErrorNoWrite when panicing.
mustSendRetryable() is called in situations when the send that triggers
the operation failed.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -882,23 +882,24 @@ func (cn *conn) Exec(query string, args []driver.Value) (res driver.Result, err 
 	return r, err
 }
 
-func (cn *conn) send(m *writeBuf) error {
-	_, err := cn.c.Write(m.wrap())
-	return err
+func (cn *conn) send(m *writeBuf) (int, error) {
+	return cn.c.Write(m.wrap())
 }
 
 func (cn *conn) mustSend(m *writeBuf) {
-	err := cn.send(m)
+	_, err := cn.send(m)
 	if err != nil {
 		panic(err)
 	}
 }
 
 func (cn *conn) mustSendRetryable(m *writeBuf) {
-	err := cn.send(m)
+	sentBytes, err := cn.send(m)
 	if err != nil {
 		if _, ok := err.(*net.OpError); ok {
-			err = &netErrorNoWrite{err}
+			if sentBytes == 0 {
+				err = &netErrorNoWrite{err}
+			}
 		}
 		panic(err)
 	}
--- a/conn.go
+++ b/conn.go
@@ -600,7 +600,7 @@ func (cn *conn) gname() string {
 func (cn *conn) simpleExec(q string) (res driver.Result, commandTag string, err error) {
 	b := cn.writeBuf('Q')
 	b.string(q)
-	cn.send(b)
+	cn.mustSendRetryable(b)
 
 	for {
 		t, r := cn.recv1()
@@ -632,7 +632,7 @@ func (cn *conn) simpleQuery(q string) (res *rows, err error) {
 
 	b := cn.writeBuf('Q')
 	b.string(q)
-	cn.send(b)
+	cn.mustSendRetryable(b)
 
 	for {
 		t, r := cn.recv1()
@@ -765,7 +765,7 @@ func (cn *conn) prepareTo(q, stmtName string) *stmt {
 	b.string(st.name)
 
 	b.next('S')
-	cn.send(b)
+	cn.mustSendRetryable(b)
 
 	cn.readParseResponse()
 	st.paramTyps, st.colNames, st.colTyps = cn.readStatementDescribeResponse()
@@ -882,9 +882,24 @@ func (cn *conn) Exec(query string, args []driver.Value) (res driver.Result, err 
 	return r, err
 }
 
-func (cn *conn) send(m *writeBuf) {
+func (cn *conn) send(m *writeBuf) error {
 	_, err := cn.c.Write(m.wrap())
+	return err
+}
+
+func (cn *conn) mustSend(m *writeBuf) {
+	err := cn.send(m)
 	if err != nil {
+		panic(err)
+	}
+}
+
+func (cn *conn) mustSendRetryable(m *writeBuf) {
+	err := cn.send(m)
+	if err != nil {
+		if _, ok := err.(*net.OpError); ok {
+			err = &netErrorNoWrite{err}
+		}
 		panic(err)
 	}
 }
@@ -1109,7 +1124,7 @@ func (cn *conn) auth(r *readBuf, o values) {
 	case 3:
 		w := cn.writeBuf('p')
 		w.string(o["password"])
-		cn.send(w)
+		cn.mustSend(w)
 
 		t, r := cn.recv()
 		if t != 'R' {
@@ -1123,7 +1138,7 @@ func (cn *conn) auth(r *readBuf, o values) {
 		s := string(r.next(4))
 		w := cn.writeBuf('p')
 		w.string("md5" + md5s(md5s(o["password"]+o["user"])+s))
-		cn.send(w)
+		cn.mustSend(w)
 
 		t, r := cn.recv()
 		if t != 'R' {
@@ -1145,7 +1160,7 @@ func (cn *conn) auth(r *readBuf, o values) {
 		w.string("SCRAM-SHA-256")
 		w.int32(len(scOut))
 		w.bytes(scOut)
-		cn.send(w)
+		cn.mustSend(w)
 
 		t, r := cn.recv()
 		if t != 'R' {
@@ -1165,7 +1180,7 @@ func (cn *conn) auth(r *readBuf, o values) {
 		scOut = sc.Out()
 		w = cn.writeBuf('p')
 		w.bytes(scOut)
-		cn.send(w)
+		cn.mustSend(w)
 
 		t, r = cn.recv()
 		if t != 'R' {
@@ -1219,9 +1234,9 @@ func (st *stmt) Close() (err error) {
 	w := st.cn.writeBuf('C')
 	w.byte('S')
 	w.string(st.name)
-	st.cn.send(w)
+	st.cn.mustSend(w)
 
-	st.cn.send(st.cn.writeBuf('S'))
+	st.cn.mustSend(st.cn.writeBuf('S'))
 
 	t, _ := st.cn.recv1()
 	if t != '3' {
@@ -1299,7 +1314,7 @@ func (st *stmt) exec(v []driver.Value) {
 	w.int32(0)
 
 	w.next('S')
-	cn.send(w)
+	cn.mustSend(w)
 
 	cn.readBindResponse()
 	cn.postExecuteWorkaround()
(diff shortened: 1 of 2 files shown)


fho:
CI jobs fails during tool installation:

```
E: Package 'postgresql-9.0' has no installation candidate
E: Unable to locate package postgresql-server-dev-9.0
E: Couldn't find any package by glob 'postgresql-server-dev-9.0'
E: Couldn't find any package by regex 'postgresql-server-dev-9.0'
E: Unable to locate package postgresql-contrib-9.0
E: Couldn't find any package by glob 'postgresql-contrib-9.0'
E: Couldn't find any package by regex 'postgresql-contrib-9.0'
```

```
fatal: cannot change to '/home/travis/gopath/src/honnef.co/go/tools/': No such file or directory
The command "./.travis.sh megacheck_install" failed and exited with 128 during .
```

fho:
@mjibson could you please provide some feedback regarding the change when you have time? thanks!

conn.go
```diff
@@ -882,13 +882,28 @@ func (cn *conn) Exec(query string, args []driver.Value) (res driver.Result, err
...
+	err := cn.send(m)
 	if err != nil {
 		panic(err)
 	}
 }
 
+func (cn *conn) mustSendRetryable(m *writeBuf) {
+	err := cn.send(m)
+	if err != nil {
+		if _, ok := err.(*net.OpError); ok {
```
mjibson: I'm not convinced this makes something safe to retry. If you send something to a server and get a network error, it is still possible for the server to have received and processed the result, but the error happens during return to the client. In that case it's not safe to retry. Do I understand the circumstances where this can happen correctly? If so, this is closer to a "maybe it worked?" error, which means the client has to check if it is safe to retry or not.

cbandy: Agreed.

fho: Thanks for your review, @mjibson and @cbandy!

>  If you send something to a server and get a network error, it is still possible for the server to 
have received and processed the result, but the error happens during return to the client. 

My networking knowledge is a bit rusty but from my understanding this can not happen:

When `send()` is called the data is only written to the kernel's send buffer and then the call returns. The `send()` call does not block until the peer acknowledged receiving the data.
Therefore it can not happen that `cn.send(m)` returns an error because something went wrong during returning an ACK to the sender.

Also errors that are returned by `send()` only indicate local errors.

Quote from Linux manpage of the `send()` syscall:
> No indication of failure to deliver is implicit in a send().  Locally detected errors are indicated by a return value of -1.

http://man7.org/linux/man-pages/man2/send.2.html

(Golang uses internally the `write()` syscall to write the data to the fd of the socket. It's the same  then using the `send()` syscall without additional flags.)

fho: Multiple SQL queries can be passed to at least the `db.Exec()` method like:
`DELETE 1 FROM TABLE; DELETE 2 FROM TABLE;`

It might possible that `cn.send(m)` returns an error after a whole statement was successfully transmitted.
The PostgreSQL server might be implemented in a way that executes the statement already as soon as a termination symbol (`;`) is received.
In this case part of the queries might be rerun if `netErrorNoWrite` is returned for those cases.

To prevent this case we can additionally check if the number of bytes that `cn.c.Write(m.wrap())` returned when an error happens are less then the length up to the first `;` in the statement (or the shortest valid postgreSQL statement). Only in this case we return `netErrorNoWrite`.

To keep it more simple, I would suggest we only return `netErrorNoWrite` if the number of bytes that were successfully written to the socket was `0` and error type is `net.OpError`.

mjibson: That's a lot of assumptions that care about how the linux kernel and go function and I'm not at all comfortable changing this until there are tests that clearly demonstrate these assumptions. Also, what about if the client and or server are on windows, mac, or some version of linux or go that doesn't happen to function this way? I'm still not convinced that it's always safe to retry if there's an error here. Distributed systems indeed have a "maybe this failed?" error because of this exact problem.

fho: @mjibson ok, I can make the code more safe by narrowing the situation when a `send()` error is identified as retryable by:

1. Also check if `Conn.Write()` returned that 0 bytes were transferred when it returned a net.OpError error.
2. Narrow down the error types to:
- `syscall.EPIPE`: "The local end has been shut down on a connection oriented socket.".
- `syscall.ENOTSOCK`: "The file descriptor sockfd does not refer to a socket.".
- `syscall.EINVAL`: "Invalid argument passed."
- `syscall.ECONNRESET`: "Connection reset by peer."
- `syscall.EBADF`: "sockfd is not a valid open file descriptor."
3. Narrow down the error type to `syscall.EPIPE` only. The PR would then address specifically the problem described in https://github.com/lib/pq/issues/870.

What do you think?
I would favor doing 1 and 2.

mjibson: I don't really know what exactly the EPIPE error means actually happened and what didn't happen. Can it change between OS or version of linux? Is the bytes written count always non-zero if any bytes were sent even if there's an error return for all paths? These questions make me very wary to do even 1 and 3.

fho: > I don't really know what exactly the EPIPE error means actually happened and what didn't happen.

You can look it up in the documentation for the related syscalls:

Posix: https://pubs.opengroup.org/onlinepubs/009695399/functions/write.html
The documentation in the e.g. linux or freebsd manpages are more clear and they implement the Posix standard, see e.g.:
https://www.freebsd.org/cgi/man.cgi?write(2) or http://man7.org/linux/man-pages/man2/write.2.html

See for Windows: https://docs.microsoft.com/en-us/windows/desktop/api/fileapi/nf-fileapi-writefile

Google can also help with understanding what EPIPE means:
https://www.google.com/search?q=what+does+%22broken+pipe%22+mean&oq=what+does+%22broken+pipe%22+mean

> Can it change between OS or version of linux?

The syscall API of operating systems is stable. It does not change in downwards incompatible manner. This would break tons of applications for the effected operation system.
Unix-like operating systems implement the POSIX API which specify this behavior. This includes OSX.

Windows Winsock API is not POSIX compliant.
But in this case it behaves the same, the `WriteFile()` returns `ERROR_BROKEN_PIPE` when the filedescriptor is closed. `ERROR_BROKEN_PIPE` is mapped to the same error then `EPIPE` by golang

See: 
https://docs.microsoft.com/en-us/windows/desktop/api/fileapi/nf-fileapi-writefile
https://golang.org/src/syscall/zerrors_windows.go#L120

> Is the bytes written count always non-zero if any bytes were sent even if there's an error return for all paths?

Yes
See the implementation of Golangs `conn.Write()` and the parent methods that it calls.
https://golang.org/src/net/net.go?s=6460:6503#L175
Unix: https://golang.org/src/internal/poll/fd_unix.go#L254
Windows: https://golang.org/src/internal/poll/fd_windows.go#657

`fd.Write()` calls `syscall.Write()` in a loop until all bytes were written out.
If it fails in the first loop iteration, it only returns an error nothing was sent out in this case.
If it fails on subsequent loop iterations it returns with the error how many bytes were send out.

fho: I pushed a commit to fix it.

aeneasr:
It looks like someone in the MySQL driver proposed having an extra go routine that checks for closed connections:

https://github.com/go-sql-driver/mysql/issues/529#issuecomment-289753856

Not sure if that's something worth considering but it would remove the issues of retrying when the client can't be sure if the server accepted the request.

stevendanna:
In my own service, I'll likely be using sql.DB's `SetMaxConnLifetime` and higher-level retries to get around the main place I hit this, but I was wondering if there is still maintainer interest in moving this forward?

The current approach in this PR (only marking the error as retriable if 0-bytes were written) is the same approach used in pgx:

https://github.com/jackc/pgconn/blob/master/pgconn.go#L860
https://github.com/jackc/pgconn/blob/master/errors.go#L160-L162
https://github.com/jackc/pgx/blob/master/stdlib/sql.go#L282-L284

I understand the concern that there may be cases where Write to a network socket will erroneously report 0 bytes written on error.  Would adding a tests for our expected behavior (for example, a test that creates a tcp connection, a writer that attempts a large write, a writer that does a small, incomplete read and then closes the connection) on particular platforms and go versions and then using conditional compilation to only enable the retry behavior on those platforms be an acceptable solution?  Changes to Go or the OS could still introduce issues, but this would perhaps give us some early warning.

No pressure either way, just wanted to check in on this since I hit it recently.

gnuletik:
Seeing the Github article today about the same issue in the mysql driver : https://github.blog/2020-05-20-three-bugs-in-the-go-mysql-driver/

Would it make sense to implement the same check for `lib/pq` ?
https://github.com/go-sql-driver/mysql/pull/934

cbandy:
> Seeing the Github article today about the same issue in the mysql driver : https://github.blog/2020-05-20-three-bugs-in-the-go-mysql-driver/
> 
> Would it make sense to implement the same check for `lib/pq` ?
> [go-sql-driver/mysql#934](https://github.com/go-sql-driver/mysql/pull/934)

1. That solution (and perhaps any solution) is still dependent on the OS. In their master branch the list of supported systems is longer than the PR but far from complete. The fallback is no-op: https://github.com/go-sql-driver/mysql/blob/master/conncheck_dummy.go

2. The linked article states that this approach (non-blocking read before write) is safe for the MySQL protocol because _the server only sends data in response to requests from the client_. This is not the case for the PostgreSQL protocol in which the server can send unprompted messages to the client. It may be possible to manage that but the implementation won't be as simple as the MySQL pull request.

    https://www.postgresql.org/docs/current/protocol-flow.html#PROTOCOL-ASYNC

hibrid:
Any update on this? Just wondering if this is planned to be merged or how we should be dealing with this because it's a real problem

mjibson:
I'm not convinced of the safety of this PR and do not intend to merge it. As described in my comments above, I believe that a network error could occur after a message has been sent to the server, in which case it is not guaranteed that it is safe for the client to auto retry. In that case, the client must check itself if a retry is safe. I could absolutely be wrong, and the error messages coming back here could be used to verify that 0 bytes have been sent, at which point it's clearly safe to retry. However I'm not certain about the guarantees of the error messages and that those guarantees are the same across all OSs running Go. Currently you'll have to solve this yourself in the application layer or merge this PR locally and hope it works.

fho:
The issue is solved.
The same solution from the PR https://github.com/lib/pq/pull/1013 was merged.

Issue #848, opened 2019-04-15
Deadlock when attempting to commit transaction after query error


polarina:
When given a cancellable context (context only has to be cancellable, no need to cancel it), attempting to call `tx.Commit()` after a failed query, will result in a deadlock.

## How to reproduce

Execute the following program:

```go
package main

import (
	"context"
	"database/sql"
	"log"
	_ "github.com/lib/pq"
)

func main() {
	ctx, _ := context.WithCancel(context.Background())
	//ctx := context.Background()

	if db, err := sql.Open("postgres", ""); err == nil {
		if tx, err := db.BeginTx(ctx, nil); err == nil {
			if _, err := tx.ExecContext(ctx, "SELECT x"); err != nil {
				log.Printf("got error as expected: %v", err)
			} else {
				log.Fatal("query succeded, shouldn't have")
			}

			if err := tx.Commit(); err != nil {
				log.Printf("got error as expected: %v", err)
			} else {
				log.Fatal("commit was successful, shouldn't have")
			}
		} else {
			log.Fatal(err)
		}
	} else {
		log.Fatal(err)
	}
}
```

## Expected output
```
2019/04/15 12:34:09 got error as expected: pq: column "x" does not exist
2019/04/15 12:34:09 got error as expected: pq: Could not complete operation in a failed transaction
```

## Actual output
```
2019/04/15 12:33:54 got error as expected: pq: column "x" does not exist
```

The program never terminates.

itizir:
Duplicate of #731. And suggesting a fix there. :)

PR #872, opened 2019-06-01
Refactor rollback in commit

### LINKED ISSUE #731: Commit on broken transaction hangs
If you attempt to Commit a broken transaction, the Commit call hangs, blocking indefinitely. If the transaction was opened with a context, the context cancellation is ignored.

This is easy to accidentally reproduce by forgetting to check the error returned when executing an SQL statement.

```go
package main

import (
	"context"
	"database/sql"
	"errors"
	"fmt"
	"time"

	_ "github.com/lib/pq"
)

func hang() error {
	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
	defer cancel()

	conn, err := sql.Open("postgres", "")
	if err != nil {
		return err
	}

	tx, err := conn.BeginTx(ctx, nil)
	if err != nil {
		return err
	}

	const q = `BROKEN SQL TO FAIL`
	_, err = tx.ExecContext(ctx, q)
	if err == nil {
		return errors.New("failed to break transaction!")
	}

	fmt.Printf("Attempting to commit broken transaction\n")

	commitCh := make(chan error, 1)
	go func() {
		commitCh <- tx.Commit()
	}()

	select {
	case err := <-commitCh:
		return err
	case <-time.After(10 * time.Second):
		fmt.Printf("Commit has hung. ctx.Err()==%v\n", ctx.Err())
	}

	return <-commitCh
}

func main() {
	err := hang()
	fmt.Printf("hang returned %v\n", err)
}
```

jaym:
Rollback cannot be called in Commit as the function that watches the context for cancelation assumes it is only ever called once. Calling it multiple times hangs the goroutine.

This is an alternative to #829.

Fixes #731 and #848 
Replaces #829

Signed-off-by: Jay Mundrawala <jmundrawala@chef.io>

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -549,7 +549,7 @@ func (cn *conn) Commit() (err error) {
 	// would get the same behaviour if you issued a COMMIT in a failed
 	// transaction, so it's also the least surprising thing to do here.
 	if cn.txnStatus == txnStatusInFailedTransaction {
-		if err := cn.Rollback(); err != nil {
+		if err := cn.rollback(); err != nil {
 			return err
 		}
 		return ErrInFailedTransaction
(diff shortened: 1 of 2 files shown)


timbunce:
Thank you!

PR #877, opened 2019-07-01
Reduce writeBuf.string run-time by 87.68%

tsenart:
```
name                 old time/op    new time/op    delta
_writeBuf_string-16    14.8ns ± 0%     1.8ns ± 0%  -87.68%  (p=0.000 n=12+15)

name                 old alloc/op   new alloc/op   delta
_writeBuf_string-16     0.00B          0.00B          ~     (all equal)

name                 old allocs/op  new allocs/op  delta
_writeBuf_string-16      0.00           0.00          ~     (all equal)
```

### DIFF
--- a/buf.go
+++ b/buf.go
@@ -66,7 +66,7 @@ func (b *writeBuf) int16(n int) {
 }
 
 func (b *writeBuf) string(s string) {
-	b.buf = append(b.buf, (s + "\000")...)
+	b.buf = append(append(b.buf, s...), "\000"...)
 }
 
 func (b *writeBuf) byte(c byte) {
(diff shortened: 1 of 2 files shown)


johto:
`"\000"...` is probably better written as `'\000'`, yes?

buf.go
```diff
@@ -66,7 +66,7 @@ func (b *writeBuf) int16(n int) {
 }
 
 func (b *writeBuf) string(s string) {
-	b.buf = append(b.buf, (s + "\000")...)
+	b.buf = append(append(b.buf, s...), "\000"...)
```
tsenart:
```suggestion
	b.buf = append(append(b.buf, s...), '\000')
```

tsenart:
@johto: Yep, thanks.

PR #883, opened 2019-07-16
bump supported versions; fix tests

mjibson:
Our tested support for Go versions is changing to the two latest instead
of three latest. This matches what the Go team supports. If lib/pq
happens to not work in older versions of Go, you must use an earlier
version of lib/pq.

Remove tested support for Go 1.9 and 1.10, add Go 1.12. Remove tested
support for unsupported Postgres versions (< 9.4).

Megacheck was renamed to staticcheck.

Fix two issues found by golint.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,8 @@
 language: go
 
 go:
-  - 1.9.x
-  - 1.10.x
   - 1.11.x
+  - 1.12.x
   - master
 
 sudo: true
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -14,26 +13,21 @@ env:
     - PQGOSSLTESTS=1
     - PQSSLCERTTEST_PATH=$PWD/certs
     - PGHOST=127.0.0.1
-    - MEGACHECK_VERSION=2017.2.2
   matrix:
     - PGVERSION=10
     - PGVERSION=9.6
     - PGVERSION=9.5
     - PGVERSION=9.4
-    - PGVERSION=9.3
```
cbandy: Should we have a job with `PGVERSION=11`?

mjibson:
I tried that but there's some apt get problem that will require more investigation, so someone can do that in another PR.

PR #890, opened 2019-08-13
Fix Database Column DATE type parsed as Timestamp when scan into string

sunfmin:


msiomkin:
An extremely destructive change. I can't scan date into time.Time any more!

Error: sql: Scan error on column index 12, name "since": unsupported Scan, storing driver.Value type string into type *time.Time

mjibson:
:( I'll revert

msiomkin:
@mjibson Thank you. It works again now.

PR #891, opened 2019-08-21
ssl: look for ~/.postgresql/root.crt

pschultz:
If the sslrootcert option hasn't been specified, use ~/.postgresql/root.crt if it exists.

[This is what libpq does](https://www.postgresql.org/docs/11/libpq-connect.html#LIBPQ-CONNECT-SSLROOTCERT).

### DIFF
--- a/ssl.go
+++ b/ssl.go
@@ -13,6 +13,11 @@ import (
 // ssl generates a function to upgrade a net.Conn based on the "sslmode" and
 // related settings. The function is nil when no upgrade should take place.
 func ssl(o values) (func(net.Conn) (net.Conn, error), error) {
+	// user.Current() might fail when cross-compiling. We have to ignore the
+	// error and continue without home directory defaults, since we wouldn't
+	// know from where to load them.
+	user, _ := user.Current()
+
 	verifyCaOnly := false
 	tlsConf := tls.Config{}
 	switch mode := o["sslmode"]; mode {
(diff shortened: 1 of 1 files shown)


pschultz:
The failing test seems unrelated to this patch.

```
--- FAIL: TestCopyRespLoopConnectionError (0.03s)
    copy_test.go:402: expected error
[...]
FAIL
FAIL	github.com/lib/pq	2.734s
```

xoxys:
Any chance to get this PR done?

rafiss:
This project is in maintenance mode (https://github.com/lib/pq#status), I can merge this patch if it gets rebased to resolve conflicts, and tests are added.

xoxys:
Sorry, missed that information. In that case, I guess it's not worth it.

pschultz:
I rewrote the patch and added a test. The patch is now simpler than before.

pschultz:
I'm currently changing the existing tests and only just now realized that this PR will change the behavior when sslrootcert is absent or empty from using system CAs to failing (unless ~/.postgresql/root.crt exists).

I'm not convinced anymore that that is a good idea, even if it matches libpq's behavior (which never considers system CAs as far as I can tell).

WDYT?

arp242:
> I'm currently changing the existing tests and only just now realized that this PR will change the behavior when sslrootcert is absent or empty from using system CAs to failing (unless ~/.postgresql/root.crt exists).
>
> I'm not convinced anymore that that is a good idea, even if it matches libpq's behavior (which never considers system CAs as far as I can tell).

As a rule I'd say that matching libpq is usually the best course of action, as that's the least surprising. It looks like pgx also looks at ~/.postgresql/root.crt

So seems okay?

I pushed an updated version here by the way: https://github.com/lib/pq/tree/root.crt – I made some changes to how it gets the user home path that conflict.

Issue #602, opened 2017-04-17
Export the function `parseArray`


pengux:
The function `parseArray` is  handy when dealing with custom array types. It would be helpful not to copy-paste that code and instead use an exported function from this package. Happy to send a PR if approved!

arp242:
Closing per the discussion at #898.

PR #898, opened 2019-09-24
export ParseArray

bleakley:
Fixes https://github.com/lib/pq/issues/602

This function is needed for writing custom Scan functions to parse a `JSONB[]` column.

### DIFF
--- a/array.go
+++ b/array.go
@@ -341,7 +341,7 @@ func (a GenericArray) Scan(src interface{}) error {
 
 func (a GenericArray) scanBytes(src []byte, dv reflect.Value) error {
 	dtype, assign, del := a.evaluateDestination(dv.Type().Elem())
-	dims, elems, err := parseArray(src, []byte(del))
+	dims, elems, err := ParseArray(src, []byte(del))
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


cbandy:
Can you give an example of what you're trying to scan into? `pq.Array()` can be used if you are scanning into an array or slice.

bleakley:
What I'm trying to do is scan an array of arbitrary types. I'm representing it in the DB as `JSONB[]` column because it might be strings or floats or another type, so each element is a JSON scalar in an array.

cbandy:
When array support was added in #466, the `parseArray` implementation was the least desired aspect. I don't expect we should commit to it further by exporting it as-is.

The unexported `parseArray` function returns a slice of byte slices. You can achieve a similar result with the exported `Array` function and a small Scanner implementation. For example,

```go
type Bytes []byte
func (b *Bytes) Scan(src interface{}) error {
  *b = append(*b, src.([]byte)...)
  return nil
}

var result []Bytes
err := db.QueryRow(`SELECT ARRAY['{}', '"a"', '9']::jsonb[]`).Scan(pq.Array(&result))
```

In your case, something like `json.RawMessage` might be more appropriate?

```go
type JSONB json.RawMessage
func (j *JSONB) Scan(src interface{}) error {
  *j = append(*j, src.([]byte)...)
  return nil
}

var result []JSONB
err := db.QueryRow(`SELECT ARRAY['{}', '"a"', '9']::jsonb[]`).Scan(pq.Array(&result))
```

Keep in mind that all PostgreSQL arrays can contain nulls... You might be happiest with the built-in types:

```go
var result []sql.NullString
err := db.QueryRow(`SELECT ARRAY['{}', '"a"', '9', NULL]::jsonb[]`).Scan(pq.Array(&result))
```

arp242:
I would rather not export this as it means it can never be changed. PostgreSQL arrays have some rather odd bits where you can do things like `[4:5]={1,2}` to explicitly set the start and end indexes, which isn't really supported right now and supporting that might require changing how this works.

And using type parameters to make this sort of thing a bit easier is probably a better way forward regardless.

PR #900, opened 2019-09-27
Implement database/sql/driver.DriverContext

dcormier:
In order to fully instrument SQL operations (including new conns) by wrapping this driver with something like [`github.com/luna-duclos/instrumentedsql`](https://godoc.org/github.com/luna-duclos/instrumentedsql), this driver needs to implement [`database/sql/driver.DriverContext`](https://godoc.org/database/sql/driver#DriverContext).

Now it does.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -45,6 +45,7 @@ var (
 // Compile time validation that our types implement the expected interfaces
 var (
 	_ driver.Driver             = Driver{}
+	_ driver.DriverContext      = Driver{}
 	_ driver.ConnBeginTx        = (*conn)(nil)
 	_ driver.ConnPrepareContext = (*conn)(nil)
 	_ driver.Execer             = (*conn)(nil) //lint:ignore SA1019 x
@@ -78,6 +79,15 @@ type Driver struct{}
 // library.
 func (d Driver) Open(name string) (driver.Conn, error) {
 	return Open(name)
+}
+
+// OpenConnector sets up a new Connector, ready to create connections. name is a
+// connection string. Most users should only use it through database/sql package
+// from the standard library.
+//
+// Implements [database/sql/driver.DriverContext].
+func (d Driver) OpenConnector(name string) (driver.Connector, error) {
+	return NewConnector(name)
 }
 
 type parameterStatus struct {
--- a/connector_test.go
+++ b/connector_test.go
@@ -82,7 +82,6 @@ func TestNewConnector(t *testing.T) {
 			t.Fatalf("wrong option for pgassfile: %q", have)
 		}
 	})
-
 	t.Run("WithConfig", func(t *testing.T) {
 		cfg, err := NewConfig("")
 		if err != nil {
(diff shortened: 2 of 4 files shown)


LasTshaMAN:
LGTM!

sam-github:
> The command "PQTEST_BINARY_PARAMETERS=yes go test -race -v ./..." exited with 1.

dcormier:
> > The command "PQTEST_BINARY_PARAMETERS=yes go test -race -v ./..." exited with 1.

So, [the docs](https://pkg.go.dev/github.com/lib/pq?tab=doc) state that that's expected:

> Note that the connection parameter client_encoding (which sets the text encoding for the connection) may be set but must be "UTF8", matching with the same rules as Postgres. It is an error to provide any other value. 

_[Source](/lib/pq/blob/9eb73efc1fcc404148b56765b0d3f61d9a5ef8ee/doc.go#L82-L85)_

[That test](/lib/pq/blob/f91d3411e481ed313eeab65ebfe9076466c39d01/conn_test.go#L1468) uses a value for `client_encoding` _other_ than `UTF8`. So, this appears to be an expected failure.

This is a test that existed before. The question is: why didn't it fail before?

Perhaps someone with more familiarity can shed some light on what's expected here.

jaredallard:
@dcormier Looks like this was introduced in https://github.com/lib/pq/commit/67337afa3f68114351e1121d67edfbe22695af13. However, that was 7 years ago.. oof.

Maybe @johto has some ideas. To me it looks like it's expected to fail, and that test verifies that it fails so I'm unsure the issue going on here. Actually looking more into this, it _is_ supposed to fail, but it errors when it does fail. This looks like a bad test to me and needs some help.

13rac1:
_Adding additional background, so others can find this PR more easily in a search._
 
This change is needed to avoid the default stdlib `dsnConnector` which throws out the Context.
```go
func (t dsnConnector) Connect(_ context.Context) (driver.Conn, error) {
	return t.driver.Open(t.dsn)
}
```
https://github.com/golang/go/blob/8f4151ea67e1d498e0880f28d3fd803dc2c5448f/src/database/sql/sql.go#L701-L708

The TCP connection `Dial()` call blocks over two minutes on my machine (I assume this is OS specific) when given an inaccessible IP with or without `Context.WithTimeout()`. [The workaround](https://stackoverflow.com/questions/40527808/setting-tcp-timeout-for-sql-connection-in-go) is to set `connect_timeout=5` in the connection string. This works, but makes a _new_ Context: https://github.com/lib/pq/blob/07d3f726dff716dcf25068ff1b74dd2229f07840/conn.go#L350-L361

13rac1:
@dcormier 

>  The question is: why didn't it fail before?

The test setup is using `sql.Open` which uses the stdlib `dsnConnector` instead of the `pq.Connector` when `DriverContext` is not implemented:
https://github.com/lib/pq/blob/f91d3411e481ed313eeab65ebfe9076466c39d01/conn_test.go#L50-L51
database/sql/sql.go:Open() https://github.com/golang/go/blob/8f4151ea67e1d498e0880f28d3fd803dc2c5448f/src/database/sql/sql.go#L770-L778
```go
	if driverCtx, ok := driveri.(driver.DriverContext); ok {
		connector, err := driverCtx.OpenConnector(dataSourceName)
		if err != nil {
			return nil, err
		}
		return OpenDB(connector), nil
	}

	return OpenDB(dsnConnector{dsn: dataSourceName, driver: driveri}), nil
```

When `DriverContext` is implemented then `driverCtx.OpenConnector` is called which runs  `pq.NewConnector` during initialization generating the error `client_encoding must be absent or 'UTF8'`.
https://github.com/lib/pq/blob/f91d3411e481ed313eeab65ebfe9076466c39d01/connector.go#L85-L88

Possible solutions:
1. Move this test to the `NewConnector` specific tests in `connector_test.go`
2. Refactor the test failure handling to check the `NewConnector` error returns earlier at https://github.com/lib/pq/blob/f91d3411e481ed313eeab65ebfe9076466c39d01/conn_test.go#L1485-L1488

Issue #902, opened 2019-10-04
Deprecate pq.NullTime 


lanrat:
database/sql will have a `NullTime` type from Go1.13 onward, which should be used instead: golang/go#30305

For backwards compatibility `pq.NullTime` could be make to alias `sql.NullTime` when compiling on Go 1.13+.

https://go-review.googlesource.com/c/go/+/170699/

PR #903, opened 2019-10-05
encode: use sql.NullTime instead of custom type

callmehollow:
Addressed issue #902.
Pull request breaks backwards compatibility, so may be custom type shouldn't be removed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,6 @@
 language: go
 
 go:
-  - 1.11.x
-  - 1.12.x
   - 1.13.x
   - master
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ language: go
 go:
   - 1.11.x
   - 1.12.x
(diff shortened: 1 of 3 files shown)


lanrat:
Does this work with older go versions and backwards compatible with libraries that still use `pq.NullTime`?

.travis.yml
```diff
@@ -1,8 +1,7 @@
 language: go
 
 go:
-  - 1.11.x
```
mjibson: Nope, we support the last two versions of Go. You'll need to make some version-tagged files for this PR.

gunturaf: the last two versions of Go is now 1.14.x and 1.13.x as of 23 Nov 2020. Should we go with this update now?

mjibson: Sure. Open up a new PR that is mergeable and I'll merge it. Be sure to @ me or I won't see it.

PR #906, opened 2019-10-11
failing test with parametrised view

steeeveb:
hi,
I'm expecting that something like the following stmt would work:
`txn.Exec("CREATE VIEW foo3 AS SELECT $1", 100)`

but instead it seems that the $1 is ignored by the driver and it then fails.

in the test I've put the same stmt with the TABLE and it works, is it a bug or I'm missing something?
thanks,
stefano

### DIFF
--- a/create_view_with_parameters_test.go
+++ b/create_view_with_parameters_test.go
@@ -1,0 +1,32 @@
+package pq
+
+import (
+	"testing"
+)
+
+func TestFoo(t *testing.T) {
+	db := openTestConn(t)
+	defer db.Close()
+
+	txn, err := db.Begin()
+	if err != nil {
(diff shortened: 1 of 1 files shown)


mjibson:
This is a misuse of the API. You have to close txns before reusing them on the same connection.

steeeveb:
I just removed the usage of txns and the problem it's still there...

mjibson:
Ah. I think you just can't use parameters in postgres views. Try it in a different driver or with psql to verify.

steeeveb:
hi again,
I tried the same thing from both psql and python/psycopg2 and it works perfectly.  IMHO the statement is correct and it never arrive to postgres, because I cannot see the error there.

PR #921, opened 2019-12-02
implement additional context specific sql interfaces

kylejbrock:


mjibson:
This failed with:

```
=== RUN   TestContextCancelBegin
    TestContextCancelBegin: go18_test.go:212: unexpected error: pq: current transaction is aborted, commands ignored until end of transaction block
--- FAIL: TestContextCancelBegin (0.02s)
```

Looks like this has bugs.

conn_go18.go
```diff
@@ -147,3 +155,67 @@ func (cn *conn) cancel(ctx context.Context) error {
...
+
+	if finish := st.watchCancel(ctx); finish != nil {
+		defer finish()
+	}
+
+	return st.Exec(list)
+}
+
+func (st *stmt) watchCancel(ctx context.Context) func() {
+	if done := ctx.Done(); done != nil {
```
mjibson: Spidey sense says this is problematic. Way too many go routines and chan listens to make me feel safe here.

kylejbrock:
Yea.  I’ve not had time to look into why.  I tried testing it locally and it passed.  So, it’ll require some deeper investigation.

mjibson:
You may need to stress it locally. Could be a race condition.

autarch:
Oops, ignore that PR I linked against this one. I meant to link to a different one.

jalan:
```
=== RUN   TestContextCancelBegin
    TestContextCancelBegin: go18_test.go:212: unexpected error: pq: current transaction is aborted, commands ignored until end of transaction block
--- FAIL: TestContextCancelBegin (0.02s)
```

The failure here seems unrelated to this PR. I spent some time trying to reproduce the failure and was unable to do so with either this PR or master. I did find one other failure for this test on Travis, https://travis-ci.org/lib/pq/jobs/355557649, so maybe the test itself has a race.

> Spidey sense says this is problematic. Way too many go routines and chan listens to make me feel safe here.

I agree that this code looks scary! But it's the same code as used here https://github.com/lib/pq/blob/master/conn_go18.go#L90

I have been test-driving this PR on a real project in a test environment, and I also wrote a couple of toy servers to check that the context cancellation works as I expect. Everything so far looks good :+1:

Lorena-babygirl (approved):
approving these changes.

Lorena-babygirl (commented):
submit tour feedback without any explicit approval.

PR #924, opened 2019-12-06
Corrected typo in connector.go OS-username fetching

grammeaway:
I tried compiling to WebAssembly using this library, and was given an error due to this little issue. Looks like it might have been a typo, it works with this fix :)

### DIFF
--- a/connector.go
+++ b/connector.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"os/user"
 	"strings"
 )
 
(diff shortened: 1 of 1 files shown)


connector.go
```diff
@@ -99,11 +100,11 @@ func NewConnector(dsn string) (*Connector, error) {
 	// resort is to use the current operating system provided user
 	// name.
 	if _, ok := o["user"]; !ok {
-		u, err := userCurrent()
```
cbandy: `userCurrent` is a local implementation to match the behavior of `libpq` on Windows:
https://github.com/lib/pq/blob/90697d60dd844d5ef6ff15135d0203f65d2f53b8/user_windows.go#L9-L17

If this PR works as expected in WebAssembly, then the better change is to add `js` here:
https://github.com/lib/pq/blob/d7918e36bee0ef49ef3dd9a17e575de83b26aac6/user_posix.go#L3

grammeaway: Roger that, I'll close this PR then 👍

PR #926, opened 2019-12-10
allow ping during a failed txn

mjibson:


apantel:
Confirmed that this prevents broken pings for me locally.

PR #794, opened 2018-10-02
Support Notices

cretz:
I needed notices and didn't see them supported (ref #580), so I implemented them as a synchronous callback. I decided to leverage the connector infrastructure as a pluggable entry point. Much thought was put into the way I handle it here and much effort was made to ensure this has as little impact as possible and is as simple as possible.

Please review.

Issue #580, opened 2017-03-14
how to view the raised notices from postgresql


whiskas:
hi! thank you for gread pg driver.

I was not found where postgresql raised messages stored in *sql.Conn

ex. in pg:

create function show_mes() returns int as $$
begin
     raise notice 'test message from pg';
     return 1;
end
$$ language plpgsql;

from go app I call this function and can get the result.

But how can i access to that messages from go app?

thank you!

whiskas:
find related questions in stackoverflow:
[http://stackoverflow.com/questions/42070023/how-do-i-get-postgresql-procedure-warning-messages-in-golang](url)

uhoh-itsmaciek:
Unfortunately these [are ignored](https://github.com/lib/pq/blob/master/conn.go#L973-L974) by pq. Given the design of database/sql, I don't think there's a good way to expose them.

whiskas:
I's very sad (
I think lot of people using pg need this feauture..

uhoh-itsmaciek:
I agree, it's sad. If you have suggestions on what the API might look like given the constraints of database/sql, I think the implementation would be straightforward, but I don't see a reasonable way to support it at all.

whiskas:
maybe we can import lib/pq not as named import

like 

import pq "github.com/lib/pq"
....
pq.logNotice(true)

and here
```

switch t {
		case 'A', 'N':
             if notice == true {
                 log.Print(........
             }
          // ignore
```

whiskas:
or set some handler like

```
pq.logNotice(handler)
...
```
```
and^
.....
case 'A', 'N':

                 cn.noticeHandler(buf)
   
```

whiskas:
[http://i.imgur.com/1YN2CUP.png](image)

cbandy:
As I recall, package-level features like this apply to all connections and have no way to indicate to you which `sql` call/DB/Txn raised the notice.

whiskas:
I think we can specify notice whene DialOpen called and conn created.

DialOpen
![http://i.imgur.com/INRX9HR.png](http://i.imgur.com/INRX9HR.png)


conn struct

![http://i.imgur.com/gSocGyN.png](http://i.imgur.com/gSocGyN.png)

PR #932, opened 2020-01-16
Add support for a handler func that receives notice messages from Pg

autarch:
Fixes lib/pq#580

Closes #794


autarch:
There's a test failure but I think this may be due to flakiness in the test code itself. The same test fails in a different way at https://travis-ci.org/lib/pq/jobs/527939031. I note that the other failure was _also_ with Pg 9.6, though that could be a red herring.

PR #935, opened 2020-01-28
url: reduce allocations in ParseURL

kevinburkemeter:
If you use e.g. "postgres://" as the connection string, ParseURL will
be called and the result will be computed each time a connection is
established to the database. This can be expensive because a new
strings.Replacer is created each time ParseURL is called. It's also
unnecessary because the output can be computed by the input and for
most processes only a single input will ever exist for the lifetime of
the process.

Reuse the same *Replacer across calls, and also cache the first 100
unique results from ParseURL, to speed up the second and third calls.

### DIFF
--- a/url.go
+++ b/url.go
@@ -6,7 +6,14 @@ import (
 	nurl "net/url"
 	"sort"
 	"strings"
+	"sync"
 )
+
+var escaperOnce sync.Once
+var escaper *strings.Replacer
+
+var parsedURLCache map[string]string
+var parsedURLMu sync.Mutex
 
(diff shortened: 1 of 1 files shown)


kevinburkemeter:
Tested this locally, it's working fine.

arp242:
Same comment from [over here](https://github.com/lib/pq/pull/1033#issuecomment-3706894516) applies: no concrete benchmarks, and not convinced it's worth optimising this.

Since this is a few years old and the author has not been active on GitHub for several years, I'll just go ahead and close it. But feel free to come back and we can re-open.

Issue #936, opened 2020-01-28
possible test flake in TestCopyLoopConnectionError


kevinburke1:
Running the tests for #935, I got a failure in code that seems unrelated: https://travis-ci.org/lib/pq/jobs/643130375

```
=== RUN   TestCopyRespLoopConnectionError
--- FAIL: TestCopyRespLoopConnectionError (0.02s)
    copy_test.go:402: expected error
```

I can't reproduce the failure locally, and the test passed on other versions of the same build matrix: https://travis-ci.org/lib/pq/builds/643130370?utm_source=github_status&utm_medium=notification

QuLogic:
This test seems racy #563. I've run a build 3 times over, but still not been able to pass this test.

PR #937, opened 2020-01-28
.travis.yml: use Go 1.13

kevinburkemeter:
It's the most recent version so we should be using it in tests.

Updates \#936.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: go
 
 go:
-  - 1.11.x
   - 1.12.x
+  - 1.13.x
   - master
 
 sudo: true


akshayjshah:
Could we update to use `stable` and `oldstable` instead of the explicitly-numbered releases? They automatically update to the two supported Go releases so we won't need PRs like this with every Go release.

kevinburkemeter:
This PR is out of date, a new Go version was released between the date it was submitted (January 28) and the current date (April 21).

mjibson:
Already taken care of in https://github.com/lib/pq/commit/ecb7a527368511ea5cfe62a8e1ce0ea5fdeef693. Just wanted you to get credit for that one.

PR #943, opened 2020-02-17
Skip TestCloseBadConn if PGHOST is a Unix-domain socket.

QuLogic:
The test fails to dial, since $PGHOST:$PGPOST would not be a valid address to connect to.


QuLogic:
The error seems unrelated, since it only occurred on one build, but there's not much output.

PR #944, opened 2020-02-25
encode: fix 2400 time encoding for time/timetz

otan:
Note that Postgres supports 24:00 for both time and timetz operations.

When evaluating "24:00" for both Time and TimeTZ datatypes, the
time.Time library does not recognise 24 as a legitimate hour. This
requires special parsing for it to work. As such, work around the
problem by subtracting a day, and adding it back later when we recognize
it as 24:00 time.

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"math"
+	"regexp"
 	"strconv"
 	"strings"
 	"sync"
@@ -15,6 +16,8 @@ import (
 
 	"github.com/lib/pq/oid"
 )
+
+var time2400Regex = regexp.MustCompile(`^(24:00(?::00(?:\.0+)?)?)(?:[Z+-].*)?$`)
 
 func binaryEncode(parameterStatus *parameterStatus, x interface{}) []byte {
 	switch v := x.(type) {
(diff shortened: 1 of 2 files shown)


otan:
looks like it passes on CI, but the "PASS" is interpreted incorrectly.

encode.go
```diff
@@ -202,10 +205,27 @@ func mustParse(f string, typ oid.Oid, s []byte) time.Time {
...
 	}
+	// Special case for 24:00 time.
+	// Unfortunately, golang does not parse 24:00 as a proper time.
+	// In this case, we want to try "round to the next day", to differentiate.
+	// As such, we find if the 24:00 time matches at the beginning; if so,
+	// we default it back to 00:00 but add a day later.
+	var is2400Time bool
+	switch typ {
+	case oid.T_timetz, oid.T_time:
+		if matches := time2400Regex.FindStringSubmatch(str); matches != nil {
```
otan: (alternative parsing suggestions welcome)

encode.go
```diff
@@ -16,6 +17,8 @@ import (
 	"github.com/lib/pq/oid"
 )
 
+var time2400Regex = regexp.MustCompile("^(24:00(:00(\\.0+)?)?)([Z+-].*)?$")
```
mjibson, line 20: since you only use the first group, use the non-capturing group syntax for the others. i think it'd be:

`^(24:00(?::00(?:\\.0+)?)?)(?:[Z+-].*)?$`

go.mod
```diff
@@ -1 +1,3 @@
 module github.com/lib/pq
+
+go 1.13
```
mjibson, line 3: this should be reverted

otan: running go test keeps bringing it back, urgh

Issue #227, opened 2014-01-31
Truncate time.Time to microsecond resolution before inserting


joeshaw:
Postgres (at least 9.3) can only store timestamps in microsecond resolution.  Go supports nanosecond resolution and `pq` formats them in `RFC3339Nano` format, which includes nanoseconds.

Postgres takes the nanoseconds and rounds them to the nearest microsecond.  Go's `time.Time` has a `Truncate` method which will truncate, rather than round, to the provided precision.

When writing unit tests to ensure that the data written to the database is as expected, I (half of the time) get failures due to the different behavior.  It would seem like truncating the `time.Time` to microsecond resolution before inserting in Postgres is the right thing to do, to get more predictable behavior between Postgres and Go.

cbandy:
I think we should leave this to the user. It is true that Postgres and Go have differing precisions for temporal values. Users (and their tests) that are concerned about nanosecond precision *should* notice (and their tests *should* fail.)

I think it is the responsibility of the driver to ensure values it sends to the backend are semantically equivalent to the values it receives from the user. Conversely, it should ensure values received from the backend are semantically equivalent to the values it sends to the user.

PSUdaemon:
I agree with @cbandy here. Don't make assumptions on what the user might want. I think the real fix is a patch to postgres to support nanosecond precision.

johto:
> I think the real fix is a patch to postgres to support nanosecond precision.

That's never gonna happen, so it's either time.Truncate() or status quo.

ST-DDT:
@johto Is there a special reason for that? I cannot find any discussion regarding that.

johto:
> Is there a special reason for that?

You'd have to either make the on-disk footprint of timestamps bigger than it already is (8 bytes), or greatly reduce the range of possible timestamps for (at least with current hardware) no gain.  And in both cases breaking backwards compatibility for an absolutely massive number of applications.

leighmcculloch:
@johto Does `rint()` always round to nearest? I thought the method that `rint` uses for rounding is configurable via `fromfp`.

anicoll:
PR to resolve issue with Postgres accuracy limit
https://github.com/lib/pq/pull/954

https://www.postgresql.org/docs/current/datatype-datetime.html

chabad360:
> > Is there a special reason for that?
> 
[...]

PR #954, opened 2020-03-27
Encode time with Postgres accuracy limit

anicoll:
Postgres limit for accuracy means we need to truncate the time to be as accurate as possible
https://www.postgresql.org/docs/current/datatype-datetime.html

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -484,7 +484,7 @@ func FormatTimestamp(t time.Time) []byte {
 		t = t.AddDate((-t.Year())*2+1, 0, 0)
 		bc = true
 	}
-	b := []byte(t.Format("2006-01-02 15:04:05.999999999Z07:00"))
+	b := []byte(t.Truncate(time.Microsecond).Format("2006-01-02 15:04:05.999999999Z07:00"))
 
 	_, offset := t.Zone()
 	offset %= 60
--- a/encode_test.go
+++ b/encode_test.go
@@ -151,18 +151,18 @@ var formatTimeTests = []struct {
 	expected string
 }{
 	{time.Time{}, "0001-01-01 00:00:00Z"},
-	{time.Date(2001, time.February, 3, 4, 5, 6, 123456789, time.FixedZone("", 0)), "2001-02-03 04:05:06.123456789Z"},
-	{time.Date(2001, time.February, 3, 4, 5, 6, 123456789, time.FixedZone("", 2*60*60)), "2001-02-03 04:05:06.123456789+02:00"},
-	{time.Date(2001, time.February, 3, 4, 5, 6, 123456789, time.FixedZone("", -6*60*60)), "2001-02-03 04:05:06.123456789-06:00"},
+	{time.Date(2001, time.February, 3, 4, 5, 6, 123456789, time.FixedZone("", 0)), "2001-02-03 04:05:06.123456Z"},
+	{time.Date(2001, time.February, 3, 4, 5, 6, 123456789, time.FixedZone("", 2*60*60)), "2001-02-03 04:05:06.123456+02:00"},
+	{time.Date(2001, time.February, 3, 4, 5, 6, 123456789, time.FixedZone("", -6*60*60)), "2001-02-03 04:05:06.123456-06:00"},
 	{time.Date(2001, time.February, 3, 4, 5, 6, 0, time.FixedZone("", -(7*60*60+30*60+9))), "2001-02-03 04:05:06-07:30:09"},
 
-	{time.Date(1, time.February, 3, 4, 5, 6, 123456789, time.FixedZone("", 0)), "0001-02-03 04:05:06.123456789Z"},
-	{time.Date(1, time.February, 3, 4, 5, 6, 123456789, time.FixedZone("", 2*60*60)), "0001-02-03 04:05:06.123456789+02:00"},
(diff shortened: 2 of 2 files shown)


encode.go
```diff
@@ -484,7 +484,7 @@ func FormatTimestamp(t time.Time) []byte {
 		t = t.AddDate((-t.Year())*2+1, 0, 0)
 		bc = true
 	}
-	b := []byte(t.Format("2006-01-02 15:04:05.999999999Z07:00"))
+	b := []byte(t.Truncate(time.Microsecond).Format("2006-01-02 15:04:05.999999999Z07:00"))
```
leighmcculloch: Some applications may prefer to round time, or round up, round down, or to truncate. Should we make this configurable? It's arguably likely to be unimportant so I'd personally be fine with truncation only, but just wondering.

heedson: The example that we ran into was with https://github.com/jinzhu/now and using the EndOfDay() of a time. It would give the exact moment before the end of a day.

Inserting that into postgres, then reading it back out (with the rounding on insert), then the output will claim it was 00:00:00... of the next day - not the same day.

anicoll: **The current issue with over accurate data in Postgres**
*`postgres-database=# select '2021-03-21T23:59:59.9999999Z'::timestamptz;  
timestamptz*
2021-03-22 00:00:00+00`
**The proposed solution**
*`postgres-database=# select '2021-03-21T23:59:59.999999Z'::timestamptz;  
timestamptz*
2021-03-21 23:59:59.999999+00`

heedson: I agree you're thought is correct too. If I had to guess, though, most Go developers will not truncate their times before giving it off to the drivers. In the situation you're dealing with contracts/time sensitive situations, it may make the difference of a customer/service hearing the current day versus the next day.

cbandy:
My opinion is the same as last time, https://github.com/lib/pq/issues/227#issuecomment-88530781. This driver should not silently throw away the nanoseconds given to it by the user.

heedson:
@cbandy Postgres doesn't seem to mention how it deals with more than microsecond precision. What are your thoughts on https://github.com/lib/pq/pull/954#discussion_r399305269 and that time isn't necessarily always best compared to a number to be rounded?

Since Postgres will end up throwing the data away, I don't think a driver doing that in a more usable way is a bad thing.

What would your opinion be on @leighmcculloch suggestion of a configurable value to decide how it's handled? We were almost bit *hard* by not truncating our times (and I believe this to not be common) that were sensitive to being `EndOfDay()` but also staying on the day. Postgres was rounding to the next day... I think that's less accurate but maybe that's only my opinion on an edge case.

leighmcculloch:
I think the bigger issue here is that the Go stdlib does not provide a function to round that parallels that used by C and by Postgres. If it did have a function that allowed us to round a number to match Postgres' rounded number in every case, would these problems exist?

heedson:
Just to clarify - if you execute the commands @anicoll provided directly on a postgres instance, it will round the same way. I think the main concern I have with rounding a `time` is that you can potentially change other facts about the "wallclock" values of said time by accident (difference of billing/expiring a customer account 1 day late). 

Rounding 1 microsecond up means 1 more microsecond. 1 more microsecond could mean a new year, new month or new day. I think it's a bit more complex than saying that we shouldn't throw away user data however I do understand that the best way this could all be solved is if Postgres itself handled it.

cbandy:
> @cbandy Postgres doesn't seem to mention how it deals with more than microsecond precision, therefore it's undefined in the spec.

Agreed. I don't see any mention of how nanoseconds are interpreted in the docs.

> What are your thoughts on [#954 (comment)](https://github.com/lib/pq/pull/954#discussion_r399305269) and that time isn't necessarily always best compared to a number to be rounded?
>
>> The example that we ran into was with https://github.com/jinzhu/now and using the EndOfDay() of a time. It would give the exact moment before the end of a day.

`EndOfDay()` seems ... ill-founded. It is the largest timestamp that `time.Time` can represent for the same `.Date()` but ... there is no "exact moment before the end of a day."

Because there is no "last" moment of any day that can be compared, most software uses "before the next day."

- SQL [`overlaps` operator](https://www.postgresql.org/docs/12/functions-datetime.html#id-1.5.8.14.6.1) takes this approach.
- PostgreSQL range types have [exclusive bounds](https://www.postgresql.org/docs/12/rangetypes.html#RANGETYPES-INCLUSIVITY). By default, the `tsrange` and `tstzrange` constructors have exclusive upper bounds.

> Since Postgres will end up throwing the data away, I don't think a driver doing that in a more usable way is a bad thing.

PostgreSQL doesn't store the nanoseconds but it does interpret them when choosing what to store. It looks like round-to-even is taking place on my machines.

> What would your opinion be on @leighmcculloch suggestion of a configurable value to decide how it's handled?

As far as I know, this driver tries to stay close to the behavior provided by `libpq`. To that end, new settings probably belong on the `Connector` type rather than in the DSN.

> We were almost bit _hard_ by not truncating our times (and I believe truncating before insertion to not be common) that were sensitive to being `EndOfDay()` but also staying on the day. Postgres was rounding to the next day... I think that's less accurate but maybe that's only my opinion on an edge case.

My experience is that (regardless of PostgreSQL) you probably want to be handling your "today" and "end of day" concepts differently. The way you're using `time.Time`, nanoseconds are really important. I suspect other parts of your stack will encounter this edge case as well.

anicoll:
Closing this PR as clearly there is no consensus on the matter.

PR #963, opened 2020-04-30
add numerice type on decode

MrSmallLiu:


arp242:
The numeric type is not a float, but an exact precision format. Converting it to a float loses data and is not correct. I'm not sure if there's any type in the standard library that has the correct semantics that match PostgreSQL's numeric type. I've used https://github.com/shopspring/decimal in the past.

Either way, I'm closing this as it's just wrong.

PR #964, opened 2020-05-02
Use evergreen version IDs in Travis config

akshayjshah:
Rather than pinning to specific Go minor releases, use the "stable" and
"oldstable" aliases. These should always point to the most recent patch
releases of the two officially-supported Go minor releases.

This should remove the need for PRs to update the Travis config every
six months.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: go
 
 go:
-  - 1.13.x
-  - 1.14.x
+  - stable
+  - oldstable
   - master
 
 sudo: true


akshayjshah:
CI failures seem to be the same flaky tests plaguing other PRs. @mjibson, mind taking a look and restarting tests if the change otherwise looks okay?

mjibson:
After thinking about this, I want to stay with the hard coded versions. There's some linter and check stuff that specifies the oldest version, and I think it's ok to have to bump occasionally.

PR #965, opened 2020-05-03
Add handler for LISTEN/NOTIFY notifications

jordanlewis:
lib/pq currently has a dedicated set of functionality for subscribing to
LISTEN/NOTIFY messages, but you have to use it on a separate connection.

This commit adds a simple way to register a handler for notifications
that is on an ordinary database connection. This is useful if a user
wants to LISTEN to a channel on the same connection as their ordinary
database connection.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -152,6 +152,9 @@ type conn struct {
 
 	// If not nil, notices will be synchronously sent here
 	noticeHandler func(*Error)
+
+	// If not nil, notifications will be synchronously sent here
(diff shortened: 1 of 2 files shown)


jordanlewis:
@mjibson PTAL. This is for adding messages to `cockroach sql` when a connection receives a LISTEN/NOTIFY message. See https://github.com/cockroachdb/cockroach/pull/48308.

jordanlewis:
Thanks!

PR #966, opened 2020-05-03
Add NotificationConnector and methods

jordanlewis:
The previous commit was kind of incomplete - it let you add a
notification handler on a single connection, but didn't have the
standard factory method (aka Connector) that let you customize the
handlers for all new connections.

This commit remedies the issue and adds a test.


jordanlewis:
cc @mjibson - sorry for the double post here, this should be good enough now.

Issue #971, opened 2020-05-27
Unnecessary dependencies for non-Kerberos users


AGWA:
An unfortunate consequence of the recently-merged Kerberos support (#970) is that users who don't use Kerberos are now forced to pull in 8 dependencies (previously pq had no dependencies) which increases both technical and legal risk.  Of particular concern is that the jcmturner libraries are licensed under a more restrictive license (Apache 2) than pq itself.  This license isn't compatible with GPLv2, which means it's no longer legal for GPLv2 projects to use pq.

My proposal for fixing this (which I am volunteering to implement if you are OK with it) is to move krb_unix.go and krb_windows.go to their own package and module. They would have an `init` function that would register the `NewGSS` function with pq.  When pq needs to create a `Gss` object it would call the registered function (or return an error if no Kerberos implementation has been registered).  The users who need Kerberos support would import the Kerberos package with a blank package name - much like how pq is often imported.  The users who don't need Kerberos wouldn't import this package and won't pay the cost of the additional dependencies.

mjibson:
Yes, I'd merge that.

amenzhinsky:
Totally agree.
The reason I really like this module over pgx or anything else is the fact it has zero dependencies, but after the latest upgrade our binary sizes grew ~3mb in size.

PR #972, opened 2020-05-29
Move Kerberos support to separate module

AGWA:
As discussed in #971. 

Before releasing this, you should update `auth/kerberos/go.mod` as follows:
 * Remove the `replace` directive.
 * Increase the version of the `github.com/lib/pq` dependency. (Moving forward this will need to be increased any time `github.com/lib/pq` has a new release.)

Thanks for working on pq!


AGWA:
The CI failures appear to be unrelated, and caused by a change in SSL certificate validation in Go master.

mjibson:
Maybe looks good except for the forced version bumping. I think it's circular and can't work. Say I want to make a new release. I update go.mod to contain the future (but non-existent) tag. I now need to update go.sum, but I haven't pushed the tag, so the go proxy can't download it yet. It seems liked we'd have to refer to the previous lib/pq version, or else make 2 releases every time? I don't like any of these options.

I think we should instead have the krb package not auto-register itself with lib/pq. This would remove the dependency on lib/pq from krb. It would add one other line for GSS-users, but that's fine. It'd be like:

```
import "github.com/lib/pq/auth/krb"

func init() { pq.RegisterNewGSSFunc(krb.NewGSS) }
```

Am I correct? I haven't tried any of this so I could be wrong about some details. Or maybe there are better ways to avoid the forced go.sum bump thing?

doc.go
```diff
@@ -243,5 +243,17 @@ bytes by the PostgreSQL server.
...
 https://godoc.org/github.com/lib/pq/example/listen.
 
+
+Kerberos Support
+
+
+If you need support for Kerberos authentication, add the following import
+statement to your program:
+
+import _ "github.com/lib/pq/auth/kerberos"
```
mjibson, line 253: indent this or whatever so godoc knows to treat it like code

mjibson: or don't and i'll do it in a follow up PR

AGWA:
I can't think of a way to avoid a two-step release process, I'm afraid.

I can rework this to require explicit registration, but it will have to look like this:

```go
import "github.com/lib/pq/auth/krb"

func init() { pq.RegisterNewGSSFunc(func() (pq.Gss, error) { return krb.NewGSS() }) }
```

That's because `RegisterNewGSSFunc` expects a function returning an interface, but to avoid the dependency on `pq`, `NewGSS` will need to return a pointer to a concrete type instead.

mjibson:
Yes that's fine. I can't see a better way either.

ggilley:
I came here to write a ticket to suggest doing this. Thanks for being on top of it. I pulled the latest pq and was surprised at all the extra dependencies (especially the gorilla sessions and securecookie)

AGWA:
I've reworked the patch to remove the dependency and require explicit registration.

PR #979, opened 2020-06-13
removed minor from version

andboson:
fixed server version detection for builds like "11.2 (Debian 11.2-1.pgdg90+1)"


andboson:
it need to be done for correct bytea encoding

Lekensteyn:
Additionally, the Travis CI tests are failing due to an unrelated change, I would recommend merging as-is @mjibson

PR #1000, opened 2020-10-02
Fix query cancellation collateralizing future queries using the same connection

kahuang:
This change solves the following issue we were seeing in production:
1. A context would get cancelled due to timeout
2. watchCancel would get triggered, and start to send the cancellation
3. Because the cancellation happens concurrently with the normal query path, and doesn't block the finish() path until its done cancelling the query on the postgres side, this connection could be returned to the pool before getting cancelled
4. This connection then gets reused, and gets collateralized by the previous query's cancel request

The changes I've made to ensure we don't run into this collateral damage:
1. When the context is cancelled, we immediately try to send the finished signal to the finished channel to prevent a race. Either we successfully send this, or the query we tried to cancel finished first. In the latter case, we just return, and let a future query (if one happens) trigger the cancellation.
2. In the former case we set the connection state to bad so it doesn't get reused, and cancel the query.
3. When the finish() function gets called, the querier goroutine closes the connection.

Note: We _probably_ don't need to change `bad` to an atomic variable, but I did just to be safe.

### DIFF
--- a/go18_test.go
+++ b/go18_test.go
@@ -3,6 +3,7 @@ package pq
 import (
 	"context"
 	"database/sql"
+	"database/sql/driver"
 	"runtime"
 	"strings"
 	"testing"
@@ -142,7 +143,7 @@ func TestContextCancelQuery(t *testing.T) {
 			cancel()
 			if err != nil {
 				t.Fatal(err)
-			} else if err := rows.Close(); err != nil {
+			} else if err := rows.Close(); err != nil && err != driver.ErrBadConn {
 				t.Fatal(err)
 			}
 		}()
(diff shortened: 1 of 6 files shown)


go18_test.go
```diff
@@ -176,6 +177,10 @@ func TestIssue617(t *testing.T) {
 			}
 		}()
 	}
+
+	// Give time for goroutines to terminate
+	time.Sleep(time.Millisecond * 50)
```
mjibson, line 182: This is a test flake waiting to happen. Can you put this into a retry loop with like a 50ms delay between attempts that fails after 1s or something?

mjibson (commented):
Please squash your commits and fix the test flake and I think I can merge this.

go18_test.go
kahuang: Will do!

kahuang:
I've updated the tests and squashed the commits

mjibson:
Need to remove go 1.13 which doesn't support some of the sync stuff here. Doing that in #1014.

jwatte:
This change causes a public API change that breaks our app.
The error returned when transacting on a canceled context used to be:
`"sql: transaction has already been committed or rolled back"`
It is now instead `"driver: bad connection"` which is not just a breaking change, but arguably the wrong error.

mjibson:
If you submit a PR that reverts this change and adds a test to prevent this regression in the future I'll merge it.

NOMORECOFFEE:
Hello.
Is it possible that connections remain in the pool in the `ErrBadConn` state? 
Should we implement the `SessionResetter` in this case?

knz:
> This change causes a public API change that breaks our app.

We've just run into the same issue inside CockroachDB. A query cancellation should not drop the connection.

rafiss:
I think https://github.com/lib/pq/pull/1079 will address the issues with this PR

evanj:
I have updated #1079 with the code review comments in case it helps.

Issue #1003, opened 2020-10-13
Empty result sets are handled incorrectly


roelvandergoot:
I expected the following test to pass:

```
func TestMultipleEmptyQuery(t *testing.T) {
       db := openTestConn(t)
       defer db.Close()

       rows, err := db.Query("select 1 where false; select 2")
       if err != nil {
               t.Fatal(err)
       }
       defer rows.Close()

       for rows.Next() {
               t.Fatal("unexpected row")
       }
       if !rows.NextResultSet() {
               t.Fatal("expected more result sets", rows.Err())
       }
       for rows.Next() {
               var i int
               if err := rows.Scan(&i); err != nil {
                       t.Fatal(err)
               }
               if i != 2 {
                       t.Fatalf("expected 2, got %d", i)
               }
       }
       if rows.NextResultSet() {
               t.Fatal("unexpected result set")
       }
}
```

roelvandergoot:
Pull request #1004

hmh:
Looks like this issue can be closed as fixed by the PR #1004 merge ?

PR #1004, opened 2020-10-14
Roel/empty-query

roelvandergoot:
* Added a test and a fix for handling empty result sets.
 * Had to set res.tag to prevent another test from failing.
 * Not sure what the purpose of this test is. Would prefer to remove both the clearing of res.tag and the offending test (commit 2).
 * Preferably, I would merge both commits, but if the offending test is somehow important we can merge only the first commit.
 * Issue: #1003

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -666,7 +666,6 @@ func (cn *conn) simpleQuery(q string) (res *rows, err error) {
 			if t == 'C' {
 				res.result, res.tag = cn.parseComplete(r.string())
 				if res.colNames != nil {
-					res.tag = ""
 					return
 				}
 			}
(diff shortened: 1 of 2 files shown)


jba:
Can we get this to land? I'm having the same issue.

roelvandergoot:
An easy work around is to add a `select 1` as the first statement in your query, and prepend your query processing with
```
for rows.Next() {}
rows.NextResultSet()
```
The better thing to do is likely to convert to jackc/pgx. :\

roelvandergoot:
@jba: Can you replace your `github.com/lib/pq` imports with `github.com/roelvandergoot/pq` and see if this solves your problem? Thank you!

roelvandergoot:
While more cases pass correctly now, I am still seeing problems...

roelvandergoot:
False alarm. Bug in my program. This pull request seems to be good! :)

Issue #897, opened 2019-09-23
broken pipe errors are not treated as driver.ErrBadConnection


vijay1811:
'write tcp 172.22.6.246:53576-\u003e 172.30.236.26:5432: write: broken pipe' this error is not treated as special error and query to database fails.

vijay1811:
I think fix could be 
`
case *net.OpError:
		cn.bad = true
		*err = v // err  = -> driver.ErrBadConnection
`
https://github.com/lib/pq/blob/master/error.go#L497

PR #1013, opened 2020-11-19
Mark `net.Conn` failed writes as recoverable when 0 bytes were written.

### LINKED ISSUE #723: goroutines and tcp errors
Full disclosure I may have something misconfigured with postgres.

max_connections on db = 1000

My understanding is that once you get the db object from the db.Open() that it is a connection pool, and it can be safely used inside goroutines for concurrent access.

However the below minimal example does one of 3 things on my machine.

1. Very rarely it runs to completion as expected.
2. Most often it fails with : "read tcp [::1]:port->[::1]:5432: read: connection reset by peer"
3. Rarely it fails with "write tcp [::1]:port->[::1]:5432: write: broken pipe"

Lowering the goroutines int variable increases the reliability, raising it makes failure a guarantee.
The number of records that prints to the terminal before an error changes each time I run a given binary.

```go
package main

import (
	"database/sql"
	"fmt"
	_ "github.com/lib/pq"
	"os"
)

func main() {
	// Connect to db
	connStr := "postgres://@localhost/test?sslmode=disable"
	db, err := sql.Open("postgres", connStr)
	if err != nil {
		panic(err.Error())
	}
	defer db.Close()

	err = db.Ping()
	if err != nil {
		panic(err.Error())
	}

	goroutines := 300

	// Channel to syncronize goroutines
	channel := make(chan int, goroutines)

	// Generate goroutines to select records and interate them
	for i := 0; i < goroutines; i++ {
		go func(j int) {
			rows, err := db.Query("select name, age from contacts where age = $1", j)
			if err != nil {
				fmt.Println(err)
				os.Exit(-1)
			}
			var name string
			var age int
			for rows.Next() {
				err := rows.Scan(&name, &age)
				if err != nil {
					fmt.Println(err)
					os.Exit(-1)
				}
			}
			fmt.Println(name, age)
			err = rows.Err()
			if err != nil {
				fmt.Println(err)
				os.Exit(-1)
			}
			rows.Close()
			channel <- 0
		}(i)
	}

	// Wait for goroutines to finish
	for i := 0; i < goroutines; i++ {
		<-channel
	}
}
```

dhermes:
Fixes #723, #870, #897.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -891,9 +891,20 @@ func (cn *conn) Exec(query string, args []driver.Value) (res driver.Result, err 
 	return r, err
 }
 
+type safeRetryError struct {
+	Err error
+}
+
+func (se *safeRetryError) Error() string {
+	return se.Err.Error()
+}
+
 func (cn *conn) send(m *writeBuf) {
-	_, err := cn.c.Write(m.wrap())
+	n, err := cn.c.Write(m.wrap())
 	if err != nil {
+		if n == 0 {
+			err = &safeRetryError{Err: err}
+		}
 		panic(err)
 	}
 }
--- a/error.go
+++ b/error.go
@@ -495,6 +495,9 @@ func (cn *conn) errRecover(err *error) {
 	case *net.OpError:
 		cn.bad = true
 		*err = v
+	case *safeRetryError:
+		cn.bad = true
+		*err = driver.ErrBadConn
 	case error:
 		if v == io.EOF || v.(error).Error() == "remote error: handshake failure" {
 			*err = driver.ErrBadConn


conn.go
```diff
@@ -891,9 +891,20 @@ func (cn *conn) Exec(query string, args []driver.Value) (res driver.Result, err
...
+	return se.Err.Error()
+}
+
 func (cn *conn) send(m *writeBuf) {
-	_, err := cn.c.Write(m.wrap())
+	n, err := cn.c.Write(m.wrap())
 	if err != nil {
+		if n == 0 {
+			err = &safeRetryError{Err: err}
+		}
```
dhermes, line 907: The motivation here is `pgx`:

The underlying connection just invokes `Write()` on a `net.Conn` and a "broken pipe" comes backs (same as here). But then a `safeToRetry` flag is set if exactly `n == 0` bytes went through the write (https://github.com/jackc/pgconn/blob/cba610c245265ff50ea3c56a9961da218ed7d730/pgconn.go#L923).

Then in the high-level code used by the driver, if an error is detected but has `safeToRetry` set, a `driver.ErrBadConn` will be surface so the `database/sql` package can handle evicting the connection and replacing it in the pool (https://github.com/jackc/pgx/blob/1df45d758d4f57b45d56fbce1e5bb8cf9aabaaf0/stdlib/sql.go#L336).

mjibson:
Merging this because it looks pretty good? Unsure if it'll break anything.

dhermes:
Thanks @mjibson. Turns out the `Fixes #N1, #N2, #N3` syntax I used only closed `#N1`. (I guess I needed `Fixes #N1. Fixes #N2. Fixes #N3.`)

Mind manually closing out #870 and #897?

mjibson:
Done.

PR #1015, opened 2020-11-24
Update error.go

splicefracture:
Missed one.

cannot use true (type bool) as type *"sync/atomic".Value in assignment


splicefracture:
@mjibson

PR #1017, opened 2020-11-28
Show position in syntax errors

### LINKED ISSUE #194: Report position of syntax errors in queries

CorentinDeBoisset:
We now format them like psql, e.g.:

    pq: syntax error at or near "FROM"
    LINE 1: SELECT * FROM FROM items;
                          ^

Closes #194

### DIFF
--- a/error_test.go
+++ b/error_test.go
@@ -1,0 +1,68 @@
+package pq
+
+import (
+	"strings"
+	"testing"
+)
+
+func TestSyntaxErrorFormatting(t *testing.T) {
+	for _, tt := range []struct {
+		err      Error
+		expected string
+	}{
+		// Single line
+		{Error{Message: "test", Position: "8", originalQuery: "SELECT *;"},
+			"pq: test\nLINE 1: SELECT *;\n               ^"},
+
+		// Syntax error in first line
+		{Error{Message: "test", Position: "1", originalQuery: "SELECT\n *;"},
+			"pq: test\nLINE 1: SELECT\n        ^"},
+
+		// Syntax error in last line
+		{Error{Message: "test", Position: "9", originalQuery: "SELECT\n *;"},
+			"pq: test\nLINE 2:  *;\n         ^"},
+
+		// Bad input: position non-positive
+		{Error{Message: "test", Position: "0", originalQuery: "SELECT\n *;"},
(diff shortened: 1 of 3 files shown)


CorentinDeBoisset:
This merge request is an updated version of !521 which has been inactive for several years.

vegarsti:
This would be super nice to have! Nice work implementing it.

error_test.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+func TestSyntaxErrorFormatting(t *testing.T) {
+	for _, tt := range []struct {
+		err      Error
+		expected string
+	}{
+		// Single line
+		{Error{Message: "test", Position: "8", originalQuery: "SELECT *;"},
+			"pq: test\nLINE 1: SELECT *;\n               ^"},
```
otan, line 15: can you make this easier to read by using the the backtick syntax? e.g.
```
expected: `pq: test
LINE 1: SELECT *;
                ^`
```

error_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+		// Syntax error in first line
+		{Error{Message: "test", Position: "1", originalQuery: "SELECT\n *;"},
```
otan, line 18: can you add a multi-line syntax error, e.g. `SELECT \n 1\n 2` (or something broken like so)

error_test.go
```diff
@@ -0,0 +1,68 @@
...
+			Error{Message: "test", Position: "not a number", originalQuery: "SELECT\n *;"}.normalError()},
+	} {
+		actual := tt.err.syntaxError()
+		if tt.expected != actual {
+			t.Errorf("bad message, expected %#v, got %#v", tt.expected, actual)
+		}
+	}
+}
+
+func TestSyntaxErrorHandlingWithQuery(t *testing.T) {
```
otan, line 42: i would delete these two tests, and make the above `originalQuery` tests go through postgres and as an integration test ensure the errors match. make one use PREPARE and one use the standard `Query`/`Exec` interface.

Issue #1031, opened 2021-03-17
Concurrent map writes


xzyfer:
After upgrading from 1.9.0 to 1.10.0 we started seeing the following panic.

```
fatal error: concurrent map writes
goroutine 52636306 [running]:
runtime.throw(0xb54c60, 0x15
    /usr/local/go/src/runtime/panic.go:1117 +0x72 fp=0xc0015aa8c8 sp=0xc0015aa898 pc=0x4382f2
runtime.mapdelete_faststr(0xaa6d60, 0xc0004014a0, 0xb4c4a1, 0x9
    /usr/local/go/src/runtime/map_faststr.go:306 +0x3de fp=0xc0015aa930 sp=0xc0015aa8c8 pc=0x414d1e
github.com/lib/pq.ssl(0xc0004014a0, 0x1000000000011, 0x1, 0x3
    /go/pkg/mod/github.com/lib/pq@v1.10.0/ssl.go:63 +0x178 fp=0xc0015aa9a0 sp=0xc0015aa930 pc=0x9c3c18
github.com/lib/pq.(*conn).ssl(0xc001f922c0, 0xc0004014a0, 0xc15678, 0xc0003346c0
    /go/pkg/mod/github.com/lib/pq@v1.10.0/conn.go:1064 +0x45 fp=0xc0015aaa40 sp=0xc0015aa9a0 pc=0x9b06c5
github.com/lib/pq.(*Connector).open(0xc00009c108, 0xc1a6a8, 0xc0024ebb60, 0xc001f922c0, 0x0, 0x0
    /go/pkg/mod/github.com/lib/pq@v1.10.0/conn.go:321 +0x245 fp=0xc0015aab48 sp=0xc0015aaa40 pc=0x9aad45
time.Sleep(0x12a05f200
    /usr/local/go/src/runtime/time.go:193 +0xd2
```

There have been issues about this which have been related to SSL options. Since 1.10 touched the SSL space it's possible there has been a regression.

jakedt:
This is also impacting us. Downgrading to 1.9.0 fixes it. We don't use inline SSL certificates (the feature in 1.10).

bjornouderoelink:
We also ran into this bug.
And indeed v1.9.0 seems to not surface the same issue.

The bug seems to be caused in the `ssl` method when upgrading a connection. 
For us this happened in the `cancel` method in `watchCancel`, but the same apparently happens in the `open` method on a Connector.
[...]

PR #1033, opened 2021-03-25
ssl: reuse *tls.Config if connection settings are identical

### LINKED ISSUE #1032: Frequently allocating in x509.parseCertificate (to repeatedly parse the same certificate)
My program provides a custom sslrootcert. Every time sql opens a new connection (which happens fairly frequently), the same certificate is re-parsed.

It would be good to check to see if we have parsed the certificate before and then reuse it instead of repeatedly doing the same work over and over again.

kevinburkemeter:
Previously, we would reload and re-parse a certificate from disk every
single time we initialized a connection and the sslrootcert setting
was enabled. This results in a lot of allocations that can be avoided.

Instead, save the *tls.Config for a given configuration hash, and
reuse it when we see it again.

Fixes #1032.


ssl.go
```diff
@@ -8,12 +8,34 @@ import (
...
 	verifyCaOnly := false
+	configMapOnce.Do(func() {
+		configMap = make(map[string]*ssldata)
+	})
+	// this function modifies o, so take the hash before any modifications are
+	// made
+	hash := string(o.Hash())
+	// This pseudo-parameter is not recognized by the PostgreSQL server, so let's delete it after use.
+	configMapMu.Lock()
+	conf, ok := configMap[hash]
```
Lekensteyn, line 34: The hash can change due to `delete(o, "sslrootcert")` and `delete(o, "sslinline")` below.

Lekensteyn (commented):
Looks reasonable, however there is a potential issue with certain parameter values that would result in two instances being stored (one with the original value, one with the deleted values). See also #1031

jmunson:
Unless I'm missing something, this looks like it would result in a stale certificate staying in memory until your program is restarted.  I don't know if it would be appropriate to register a SIGHUP handler in a library to drop the cache, but that would be preferable to needing to restart the service.

Or maybe we could still stat() the file on connect time and re-parse if mtime is newer than what we have cached?

arp242:
> this looks like it would result in a stale certificate staying in memory until your program is restarted

Yes, it changes the behaviour.

In general I'm not convinced it's worth the effort to optimise this. It saves "a lot of allocations that can be avoided", but how many? Are we talking about 100ns per connection or 50ms? This needs some concrete benchmarks.

Since this is a few years old and the author has not been active on GitHub for several years, I'll just go ahead and close it. But feel free to come back and we can re-open.

PR #1034, opened 2021-04-14
Fix string quoting in an example in Array

vegarsti:
The current quote is invalid Go code and e.g. lights up in VSCode when you hover to get the docstring


vegarsti:
Thanks @mjibson!

PR #1039, opened 2021-05-12
encode: fix TimeTZ with second offsets

otan:
lib/pq previously did not work with TimeTZ offsets with second offsets.
Rectify this by modifying the parse format string as appropriate.

Refs: #65066


knz:
FYI this is failing CI so there's a bit of polish with this.

@mjibson do you reckon you could help us with getting this merged at some point? Also, what's the process to have someone of the new team here to get permission to integrate changes in lib/pq?

mjibson:
CI is failing due to a known goimports change on master - I don't think it's a real bug.

PR #1048, opened 2021-06-16
Add support for sslcrl option

Tasssadar:
Adds support for the certificate revocation mechanism via CRL. I have tried to mimic what libpq does when verifying CRL.

I'm willing to write tests for this, but I need the private key of pq CA to generate the testing CRLs :/


mjibson:
https://github.com/lib/pq/tree/master/certs might have that key

Tasssadar:
> https://github.com/lib/pq/tree/master/certs might have that key

There are only the leaf cert private keys there, the CA private key (`C=US, ST=Nevada, L=Las Vegas, O=github.com/lib/pq, CN=pq CA`) is needed to generate CRL.

otan:
You can recreate the keys if you want from scratch, just make sure you change them in CI as well.
See: https://github.com/lib/pq/pull/1054/files

arp242:
> I'm willing to write tests for this, but I need the private key of pq CA to generate the testing CRLs :/

It should be okay to just regenerate all the files with a new private key.

I looked a bit at this, but wasn't really able to generate CRLs for the tests – the openssl CLI can be such a pain.

I don't mind solving the conflicts etc. myself, but any help with that would be appreciated – if you're still interested in this after five years 😅

PR #1054, opened 2021-09-01
implement gh actions workflow

michaelshobbs:
ref: https://github.com/lib/pq/pull/1047#issuecomment-906062580

the `TestSSLClientCertificates` fails on postgres 13 & 12 with `ssl_test.go:269: remote error: tls: bad certificate` but the other postgres versions seem to work.

EDIT: also `TestSSLVerifyFull` fails on golang 1.17 with `ssl_test.go:104: x509: certificate relies on legacy Common Name field, use SANs instead`. From what I can tell, they removed the `GODEBUG=x509ignoreCN=0` flag.


michaelshobbs:
@otan @mjibson I'm not an SSL expert by any stretch so any pointers here would be great

otan:
i'm no i'm no export either, but my guess is those certificates have to be regenerated.

Makefile
```diff
@@ -0,0 +1,36 @@
+.PHONY: all root-ssl server-ssl client-ssl
```
mjibson, line 1: Can you move this to the certs dir and add a comment explaining what it does?

michaelshobbs: done 7da16d9

otan:
this is awesome, cheers buddy

Issue #1062, opened 2021-10-19
drivers.ErrBadConn returned for DB.QueryRowContext.Scan when context is cancelled


mjl-:
I have been seeing unexpected "driver: bad connection" (driver.ErrBadConn) errors in my logging. In my case, these are returned by calls to database/sql's DB.QueryRowContext.Scan (and likely also on Tx's) when their context is cancelled. Cancelling contexts happens for me in practice because I pass contexts from http.Request's on to lib/pq. I believe DB.QueryRowContext.Scan should not be made to return driver.ErrBadConn when contexts are cancelled.

I'll add a small reproducer after I've created the issue.

I looks the following ordered events take place in lib/pq:
- In `conn.QueryContext`, `conn.query()` returns successfully.
- The goroutine from `conn.watchCancel()` (scheduled in `conn.QueryContext`) handles the cancellation, calling `conn.setBad()` to mark the connection.
- `rows.Next` (for `Scan` after `DB.QueryRowContext`) checks the connection with `conn.getBad()` and returns a driver.ErrBadConn.

I would expect the `Scan` on the sql `Row` to return context.Canceled in this case.

Perhaps lib/pq's connections could start to keep track not just whether connections are bad, but also the error to return to future calls on the connection. For expired contexts, the errors would become eg context.Canceled or context.DeadlineExceeded.

lunemec:
We are facing similar issue, cancelling the context by http server when user cancels the request triggers `driver: bad connection` to be raised. It seems that in the end it causes sql package to re-connect over and over again, until for some reason the connection pool is exhausted (that is most likely caused by different part of the code, maybe even stdlib or our own).

mjl-:
FWIW, I haven't seen unexpected connection pool exhaustion or problematic reconnects.

The workaround I put in my code is to check for the database/sql/driver.ErrBadConn explicitly and treating it similarly to how I treat context.Canceled errors: marking it a user error, not a server error. That saves me getting alerted.

fho:
This issue was solved in release 1.10.4 via via https://github.com/lib/pq/pull/1064 ?

otan:
Should be!

qburst-sreeraj:
@otan @mjl- 
We still have this issue.

PR #1064, opened 2021-10-19
issue 1062: Keep track of (context cancelled) error on connection, and make rows.Next return it

mjl-:
For https://github.com/lib/pq/issues/1062

Instead of just whether the connection is ErrBadConn. Often times, the error
will still be ErrBadConn. But for expired/cancelled contexts, it will be the
error for the context. Most functions still return ErrBadConn per the
database/sql/driver contract ("ErrBadConn should only be returned from [...] a
query method"). For rows.Next() we return the context-related error.

The database/sql/driver contract doesn't look very precise. Is Next a "query
method" and should database/sql handle ErrBadConns when Next returns them?

Do we have more functions that should return the canceled-context error
message?

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -18,7 +18,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
-	"sync/atomic"
+	"sync"
 	"time"
 	"unicode"
 
@@ -140,9 +140,10 @@ type conn struct {
 	saveMessageType   byte
 	saveMessageBuffer []byte
 
-	// If true, this connection is bad and all public-facing functions should
-	// return ErrBadConn.
-	bad *atomic.Value
+	// If an error is set, this connection is bad and all public-facing
+	// functions should return the appropriate error by calling get()
+	// (ErrBadConn) or getForNext().
+	err syncErr
 
 	// If set, this connection should never use the binary format when
 	// receiving query results from prepared statements.  Only provided for
@@ -164,6 +165,40 @@ type conn struct {
 
 	// GSSAPI context
 	gss GSS
+}
+
+type syncErr struct {
+	err error
+	sync.Mutex
+}
+
+// Return ErrBadConn if connection is bad.
+func (e *syncErr) get() error {
+	e.Lock()
+	defer e.Unlock()
+	if e.err != nil {
+		return driver.ErrBadConn
+	}
+	return nil
+}
+
+// Return the error set on the connection. Currently only used by rows.Next.
+func (e *syncErr) getForNext() error {
+	e.Lock()
+	defer e.Unlock()
+	return e.err
+}
+
+// Set error, only if it isn't set yet.
+func (e *syncErr) set(err error) {
+	if err == nil {
+		panic("attempt to set nil err")
+	}
+	e.Lock()
+	defer e.Unlock()
+	if e.err == nil {
+		e.err = err
+	}
 }
 
 // Handle driver-side settings in parsed connection string.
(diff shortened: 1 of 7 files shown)


conn.go
```diff
@@ -166,6 +167,40 @@ type conn struct {
 	gss GSS
 }
 
+type syncErr struct {
+	err error
```
otan, line 171: any reason not to use `type syncErr atomic.Value` here, avoiding the mutex in the impls below?

conn_test.go
```diff
@@ -696,23 +695,19 @@ func TestErrorDuringStartupClosesConn(t *testing.T) {
...
+	cn := conn{}
 	func() {
 		defer cn.errRecover(&err)
 		panic(io.EOF)
 	}()
 	if err != driver.ErrBadConn {
 		t.Fatalf("expected driver.ErrBadConn, got: %#v", err)
 	}
-	if !cn.getBad() {
+	if err := cn.err.get(); err == nil {
```
otan, line 706: we should assert what kind of error gets returned here

conn_test.go
```diff
@@ -721,7 +716,7 @@ func TestBadConn(t *testing.T) {
...
```
otan, line 719: here too

copy.go
```diff
@@ -49,12 +49,12 @@ type copyin struct {
 	buffer  []byte
 	rowData chan []byte
 	done    chan bool
-	driver.Result
 
 	closed bool
 
-	sync.Mutex // guards err
+	sync.Mutex // guards err and Result
```
otan, line 55: whilst you're here, can you make this easier to not make a mistake by going:
```
type copyin struct {
  // ....
  type mu struct {
     sync.Mutex
     driver.Result
     err error
  }
}

```

then use

```
var c copyin
c.mu.Lock() ....
c.mu.Result() ...
```

that way it's clear how the mutex is used.

issues_test.go
```diff
@@ -58,3 +59,22 @@ func TestIssue1046(t *testing.T) {
...
+	// Ensure that cancelling a QueryRowContext does not result in an ErrBadConn.
+
+	for i := 0; i < 100; i++ {
+		ctx, cancel := context.WithCancel(context.Background())
+		go cancel()
+		row := db.QueryRowContext(ctx, "select 1")
+
+		var v int
+		err := row.Scan(&v)
+		if err == driver.ErrBadConn {
```
otan, line 76: can we assert instead what error we expect?

otan (commented):
i don't check this repo often, sorry for the super slow reply (you'll need to ping me if its >24 business hours)

have a couple of cleanup suggestions for you

otan:
looks like this is also failing tests

conn.go
mjl-: IIRC, I tried and ran into not being able to store a nil value. I currently don't see the code storing nil though, so I could have run into it while testing only. I'll give it a try.

mjl-:
The failed error was due to existing checks around "context canceled" not handling the error.
Those places still check for driver.ErrBadConn. I'll look into if I can remove those errors there.

I'll also look into your other cleanup suggestions. Hopefully within the next few days.

conn.go
mjl-: ah, the actual reason is that syncErr.set only wants to set an error if it is currently nil. atomic.Value's CompareAndSwap was introduced only in go1.17.

copy.go
mjl-: i added the mu struct. by making the err & Result available under a different name, and fixing the uses, i noticed the usage wasn't correct.

conn_test.go
mjl-: fixed

conn_test.go
mjl-: also fixed

PR #1075, opened 2022-04-12
add raw CopyData command

rafiss:


rafiss:
replaced by https://github.com/lib/pq/pull/1076

PR #1077, opened 2022-04-12
expose raw CopyData command

otan:


copy.go
```diff
@@ -273,6 +274,38 @@ func (ci *copyin) Exec(v []driver.Value) (r driver.Result, err error) {
 	return driver.RowsAffected(0), nil
 }
 
+// CopyData executes a raw CopyData command using the PostgreSQL Frontend/Backend
+// protocol. Use Exec(nil) to finish the command.
```
rafiss, line 278: could fix this comment

rafiss (commented):
ah just saw this. i had done the same in https://github.com/lib/pq/pull/1076/

PR #1079, opened 2022-04-15
conn: Implement driver.Validator, SessionResetter for cancelation

evanj:
Commit 8446d16b89 released in 1.10.4 changed how some cancelled query
errors were returned. This has caused a lib/pq application I work on
to start returning "driver: bad connection". This is because we were
cancelling a query, after looking at some of the rows. This causes a
"bad" connection to be returned to the connection pool.

To prevent this, implement the driver.Validator and
driver.SessionResetter interfaces. The database/sql/driver package
recommends implementing them:

"All Conn implementations should implement the following interfaces:
Pinger, SessionResetter, and Validator"

Add two tests for this behaviour. One of these tests passed with
1.10.3 but fails with newer versions. The other never passed, but
does after this change.


jimenez:
👍

conn.go
```diff
@@ -2058,3 +2058,21 @@ func alnumLowerASCII(ch rune) rune {
...
+
+func (cn *conn) ResetSession(ctx context.Context) error {
+	// Ensure bad connections are reported: From database/sql/driver:
+	// If a connection is never returned to the connection pool but immediately reused, then
+	// ResetSession is called prior to reuse but IsValid is not called.
+	return cn.err.get()
+}
+
+func (cn *conn) IsValid() bool {
+	// panic("TODO IsValid")
```
rafiss, line 2076: nit: remove comment

rafiss (changes requested):
thanks! i think this will fix the concern raised in this comment https://github.com/lib/pq/pull/1000#issuecomment-738920726

just had a small comment

conn.go
evanj: Embarrassing! Thanks for noticing. Removed.

evanj:
Sorry for the massive delay. I must have lost the github notification on the PR. I have merged the latest master, and have fixed the code review comment above. Thanks!

timbunce:
@rafiss can your request for changes be removed now that @evanj addressed your concerns?

rafiss (approved):
thanks for the contribution!

fho:
@evanj thanks for the fix, we ran into the same issue recently

Issue #825, opened 2019-01-15
No way to override - Error: pq: Private key file has group or world access. Permissions should be u=rw (0600) or less


protosam:
I found while using Windows Subsystem for Linux, that when I tried using the Linux binary of cockroachdb, libpq didn't like my certificate files because the permissions are 777.

There should be a way to override this check for people that just want to use libpq in a test environment where permissions don't matter. I have to keep the files on the Windows partition where there are no unix based permissions so I have direct access to my files with IDEs and such.

This could also be problematic for users trying to utilize storage systems from Linux that don't use unix-like permissions, where everything presents as 777 or some arbitrary perm value.

The cockroachdb commands had an elegant solution of reading the environment variable `COCKROACH_SKIP_KEY_PERMISSION_CHECK` to bypass it's check:
https://github.com/cockroachdb/cockroach/blob/master/pkg/security/certificate_loader.go#L33-L41

Maybe something like this could be a viable option for libpq?

cbandy:
Could we remove this build flag and instead check `runtime.GOOS == "windows"`?
https://golang.org/pkg/runtime/#pkg-constants

https://github.com/lib/pq/blob/9eb73efc1fcc404148b56765b0d3f61d9a5ef8ee/ssl_permissions.go#L1

protosam:
I think that would be logical to do and slap in a check for an environment variable instead that a user can modify.

singron:
This also blocks use of keys that are intentionally readable by a group.

theonewolf:
@singron we hit this exact issue when trying to use a PostgreSQL backend with certificate-based authentication within Kubernetes.

Reference issues that this causes:

- https://github.com/banzaicloud/bank-vaults/issues/777
- https://github.com/kubernetes/kubernetes/issues/2630

Ideally, we'd have an easy method within Kubernetes of setting the user and mode bits precisely on mounted files, but that currently doesn't exist.

We can get the group set canonically, but there is then no way to override this check within `lib/pq`.

mikroskeem:
Any updates on this?

whyman:
Any update on this?

SteveVaknin:
Updates?

tisc0:
> The cockroachdb commands had an elegant solution of reading the environment variable `COCKROACH_SKIP_KEY_PERMISSION_CHECK` to bypass it's check:
> https://github.com/cockroachdb/cockroach/blob/master/pkg/security/certificate_loader.go#L33-L41
> 
[...]

PR #1081, opened 2022-05-06
adds support for kubernetes mounted private keys

catj-cockroach:
This PR adds support for using private keys mounted in Kubernetes. It should partially fix issue #825 at least in Kubernetes, with the use of `fsGroup` in the `securityContext` and `defaultMode` on the mounted secret.

### DIFF
--- a/ssl_permissions.go
+++ b/ssl_permissions.go
@@ -3,7 +3,28 @@
 
 package pq
 
-import "os"
+import (
+	"errors"
+	"os"
+	"syscall"
+)
+
+const (
+	rootUserID = uint32(0)
+
+	// The maximum permissions that a private key file owned by a regular user
+	// is allowed to have. This translates to u=rw.
+	maxUserOwnedKeyPermissions os.FileMode = 0600
+
+	// The maximum permissions that a private key file owned by root is allowed
+	// to have. This translates to u=rw,g=r.
+	maxRootOwnedKeyPermissions os.FileMode = 0640
+)
+
+var (
+	errSSLKeyHasUnacceptableUserPermissions = errors.New("permissions for files not owned by root should be u=rw (0600) or less")
+	errSSLKeyHasUnacceptableRootPermissions = errors.New("permissions for root owned files should be u=rw,g=r (0640) or less")
+)
 
 // sslKeyPermissions checks the permissions on user-supplied ssl key files.
 // The key file should have very little access.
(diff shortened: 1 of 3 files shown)


cbandy:
PostgreSQL recently changed the permissions it allows. This driver should use logic similar to `fe-secure-openssl.c` here: https://git.postgresql.org/gitweb/?p=postgresql.git;a=commitdiff;h=a59c79564bdc209a5bc7b02d706f0d7352eb82fa

catj-cockroach:
Hi @cbandy! I've updated the PR to match the permission check in the PostgreSQL!

catj-cockroach:
Sorry about the linter errors, I've fixed them now and just to be safe I ran the test suite again.

catj-cockroach:
Hey @rafiss! This is the PR we spoke about 1:1 :)

ssl_permissions.go
```diff
@@ -14,8 +29,58 @@ func sslKeyPermissions(sslkey string) error {
 	if err != nil {
 		return err
 	}
-	if info.Mode().Perm()&0077 != 0 {
-		return ErrSSLKeyHasWorldPermissions
+
+	err = hasCorrectPermissions(info)
+
+	// simplify the error for output purposes
+	if err == ErrSSLKeyHasUnacceptableUserPermissions || err == ErrSSLKeyHasUnacceptableRootPermissions {
```
rafiss, line 36: it seems  like it would be simpler if we got rid of `ErrSSLKeyHasUnacceptableUserPermissions` and `ErrSSLKeyHasUnacceptableRootPermissions`. it looks like they are only used here, but always immediately re-written, so no caller would ever see those errors.

alternatively, if there is a reason to keep them, could you leave  a comment explaining why? and also, i think they should be lower-cased so they are not exported.

---

yet another alternative is error-wrapping (see suggestion below)

ssl_permissions.go
```diff
@@ -14,8 +29,58 @@ func sslKeyPermissions(sslkey string) error {
...
+	// Postgres does.
+	if unixStat.Uid == rootUserID {
+		rootPermissionMask := (os.FileMode(0777) ^ maxRootOwnedKeyPermissions)
+		if info.Mode().Perm()&rootPermissionMask != 0 {
+			return ErrSSLKeyHasUnacceptableRootPermissions
+		}
+		return nil
+	}
+
+	return ErrSSLKeyHasUnacceptableUserPermissions
```
rafiss, line 85: and this could be
```
return fmt.Errorf("%w Permissions of a key file owned by a non-root user should be u=rw (0600) or less", ErrSSLKeyHasWorldPermissions)
```

ssl_permissions.go
```diff
@@ -14,8 +29,58 @@ func sslKeyPermissions(sslkey string) error {
...
+	if !ok {
+		return ErrSSLKeyUnknownOwnership
+	}
+
+	// if the file is owned by root, we allow 0640 (u=rw,g=r) to match what
+	// Postgres does.
+	if unixStat.Uid == rootUserID {
+		rootPermissionMask := (os.FileMode(0777) ^ maxRootOwnedKeyPermissions)
+		if info.Mode().Perm()&rootPermissionMask != 0 {
+			return ErrSSLKeyHasUnacceptableRootPermissions
```
rafiss, line 80: if we went with the error wrapping approach, then this could be

```
return fmt.Errorf("%w Permissions of a key file owned by root should be u=rw (0640) or less", ErrSSLKeyHasWorldPermissions)
```

and change the ErrSSLKeyHasWorldPermissions definition to
```
ErrSSLKeyHasWorldPermissions = errors.New("pq: Private key file has incorrect acccess.")
```

---

this way the error message returned to the user only has information that is relevant to what they are trying to do. also, since `ErrSSLKeyHasWorldPermissions` is part of the public API of this package, the `%w` will cause that error to be wrapped so that any code that was using an `errors.Is(err, ErrSSLKeyHasWorldPermissions)` check will keep working

rafiss (changes requested):

ssl_permissions.go
catj-cockroach: They're used here and in the tests, and ideally would be returned by the function. However, I didn't want to make this change into one that breaks backwards compatibility with existing consumers of the library, which I'd assume would be necessary if I wrapped the error or updating it to a custom error (which would be my preferred alternative to wrapping the errors).

I see your point about not exporting them though! I'll do that now

rafiss: > which I'd assume would be necessary if I wrapped the error

actually wrapping the error should be fine IMO. as i mentioned in https://github.com/lib/pq/pull/1081#discussion_r869882105 calls to `errors.Is(err, ErrSSLKeyHasWorldPermissions)` would still work. 

but it would break any comparisons that use `err == ErrSSLKeyHasWorldPermissions` (which people shouldn't be doing...)

hmmm, so if you feel like it's best as-is, i'm fine to merge

catj-cockroach: Honestly most of the Go code I've worked with still uses `err ==` instead of `errors.Is` so I'd prefer to leave it as is if that's OK with you! I think you might need to run the workflow though, since I'm a first time contributor.

rafiss: ok, then lgtm!

rafiss (approved):

arjantop-cai:
@catj-cockroach This does not appear to resolve the k8s fsGroup issue, as soon as the container is running as non-root and you need to set fsGroup it does not work, as this PR only allows 0640 permission if the mounted secret is owned by root.

catj-cockroach:
@arjantop-cai you'll also need to set the `defaultMode` for the mounted secret to `288`, which is 0640 in decimal

arjantop-cai:
@catj-cockroach my bad, skaffold just did not rebuild the image with the new pq version, all works as expected 👍

PR #1087, opened 2022-07-28
Fix build tags for user_other file.

Toasterson:
During the compilation of gitea on illumos I ran into the issue that both `user_posix.go` and `user_other.go` where considered for inclusion by the golang compiler with golang 1.18. 

This PR forces the golang compiler to exclude user_other on all platforms mentioned in user_posix.

Testing on Windows should be done for this PR as I do not own any Windows machine and it is likely that `user_other.go` also conflicts with`user_windows.go`.

### DIFF
--- a/user_other.go
+++ b/user_other.go
@@ -1,7 +1,11 @@
 // Package pq is a pure Go Postgres driver for the database/sql package.
 
-//go:build js || android || hurd || zos || !solaris || !illumos
-// +build js android hurd zos !solaris !illumos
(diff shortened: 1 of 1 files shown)


cbandy:
What is your `GOOS` ?

Toasterson:
GOOS="illumos"

arp242:
`GOOS=illumos go test -c` seems to work.

I don't know why you're having problems; with the current `//go:build js || android || hurd || zos || wasip1` it should not build on illumos. Maybe something strange in your setup? Or something else?

Since this is a few years old I'll just go ahead and close it, but feel free to come back if you're still having problems.

Issue #488, opened 2016-08-17
Documentation: TLS SNI


Superdominik:
It would be great if the readme would reflect whether pq supports TLS SNI. 

I couldn't find any info about it. It's a very important piece of information for everyone who runs his PostgreSQL server behind a reverse proxy and wants to use SSL.

Does it support TLS SNI?

johto:
Like most Go programs, it uses [crypto/tls](https://golang.org/pkg/crypto/tls/) underneath, which supports SNI.

I don't think this needs any documentation.  Anyone who cares about this can just take a look at the source code or test it with a 10-line program.

johto:
Well, just as I hit send there I just realized that it's not actually completely true.  The postgres protocol has its own plaintext handshake before the TLS connection is established, which renders SNI useless unless the proxy/whatever you're interested in knows how to read that handshake.

cbandy:
We only set `tls.Config.ServerName` when `sslmode=verify-full`. I'll check what `libpq` does later today.

cbandy:
`libpq` never calls `SSL_set_tlsext_host_name` that I can see. A suggestion to add it was dismissed some time ago: https://www.postgresql.org/message-id/20131129201943.GA28300@gmail.com

If you have a compelling case for hosting multiple instances on a single port, we can reopen the discussion.

sudeeptoroy:
If the server is behind a reverse proxy which can proxy/route on SNI, in such scenario we  will require pq to support SNI

petr4:
Do you have any progress on it? We use many postgres instances on K8s and want to route traffic over ingress.

cbandy:
> We use many postgres instances on K8s and want to route traffic over ingress.

Kubernetes Ingress does not natively support TCP applications or services. What controller are you using that can?

sudeeptoroy:
Contour ingress support TCP application.
https://projectcontour.io/guides/tls/

cbandy:
> Contour supports HTTPS (TLS/SSL) ingress…

I see a lot about HTTP and HTTPS at https://projectcontour.io/ and nothing about TCP routing or ingress.

sudeeptoroy:
> > Contour supports HTTPS (TLS/SSL) ingress…
> 
> I see a lot about HTTP and HTTPS at https://projectcontour.io/ and nothing about TCP routing or ingress.

Coutour does support TCP proxying.  I have used it extensively in my projects. More information here.   
 
[...]

PR #1088, opened 2022-08-20
Set SNI for TSL connections

kelvich:
This allows an SNI-aware proxy to route connections. Patch adds a new
connection option (`sslsni`) for opting out of the SNI, to have the same
behavior as `libpq` does. See more in `sslsni` sections at
<https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-PARAMKEYWORDS>.

### DIFF
--- a/ssl.go
+++ b/ssl.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"os/user"
 	"path/filepath"
+	"strings"
 )
 
 // ssl generates a function to upgrade a net.Conn based on the "sslmode" and
@@ -48,6 +49,16 @@ func ssl(o values) (func(net.Conn) (net.Conn, error), error) {
 		return nil, nil
 	default:
 		return nil, fmterrorf(`unsupported sslmode %q; only "require" (default), "verify-full", "verify-ca", and "disable" supported`, mode)
+	}
+
+	// Set Server Name Indication (SNI), if enabled by connection parameters.
+	// By default SNI is on, any value which is not starting with "1" disables
+	// SNI -- that is the same check vanilla libpq uses.
+	if sslsni := o["sslsni"]; sslsni == "" || strings.HasPrefix(sslsni, "1") {
+		// RFC 6066 asks to not set SNI if the host is a literal IP address (IPv4
+		// or IPv6). This check is coded already crypto.tls.hostnameInSNI, so
+		// just always set ServerName here and let crypto/tls do the filtering.
+		tlsConf.ServerName = o["host"]
 	}
 
 	err := sslClientCertificates(&tlsConf, o)
--- a/ssl_test.go
+++ b/ssl_test.go
@@ -3,12 +3,19 @@ package pq
 // This file contains SSL tests
 
 import (
+	"bytes"
 	_ "crypto/sha256"
+	"crypto/tls"
 	"crypto/x509"
 	"database/sql"
+	"fmt"
+	"io"
+	"net"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
+	"time"
 )
 
 func maybeSkipSSLTests(t *testing.T) {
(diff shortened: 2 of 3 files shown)


kelvich:
Relevant discussion: https://github.com/lib/pq/issues/488
`libpq` sets SNI since v14

ssl_test.go
```diff
@@ -280,3 +287,132 @@ func TestSSLClientCertificates(t *testing.T) {
...
+			serverErrChan := make(chan error, 1)
+			serverSNINameChan := make(chan string, 1)
+			go mockPostgresSSL(listener, serverErrChan, serverSNINameChan)
+
+			defer listener.Close()
+			defer close(serverErrChan)
+			defer close(serverSNINameChan)
+
+			// Try to establish a connection with the mock server. Connection will error out after TLS
+			// clientHello, but we it is enough to catch SNI data on the server side
```
cbandy, line 344:
```suggestion
			// clientHello, but it is enough to catch SNI data on the server side
```

ssl_test.go
```diff
@@ -280,3 +287,132 @@ func TestSSLClientCertificates(t *testing.T) {
...
+
+			// Check SNI data
+			select {
+			case sniHost := <-serverSNINameChan:
+				if sniHost != tt.expected_sni {
+					t.Fatalf("Expected SNI to be 'localhost', got '%+v' instead", sniHost)
+				}
+			case err = <-serverErrChan:
+				t.Fatalf("mock server failed with error: %+v", err)
+			case <-time.After(time.Millisecond * 100):
```
cbandy, line 358: Should this timeout be larger? The other I see is 5 seconds.

cbandy (approved):
lgtm

ssl_test.go
kelvich: fixed, thanks

ssl_test.go
kelvich: Set both to be 1 second. On my laptop, `TestSNISupport` takes 14-15ms

kelvich:
@rafiss, any chance you can take a look? This PR is hanging approved for some amount of time.

ssl.go
```diff
@@ -50,6 +50,16 @@ func ssl(o values) (func(net.Conn) (net.Conn, error), error) {
 		return nil, fmterrorf(`unsupported sslmode %q; only "require" (default), "verify-full", "verify-ca", and "disable" supported`, mode)
 	}
 
+	// Set Server Name Indication (SNI), if enabled by connection parameters.
+	// By default SNI is on, any value != "1" disables SNI.
+	if sslsni := o["sslsni"]; sslsni == "" || sslsni == "1" {
```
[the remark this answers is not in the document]
sean-: It only sets SNI if the first character in the variable is `1`:

https://github.com/postgres/postgres/blob/master/src/interfaces/libpq/fe-secure-openssl.c#L1216

`libpq` also limits the input, too, so you can't use `true` as a value: https://github.com/postgres/postgres/blob/master/src/interfaces/libpq/fe-connect.c#L302-L304

rafiss (changes requested):
errors from linting:
```
Error: ssl_test.go:347:4: this value of err is never used (SA4006)
Error: ssl_test.go:348:4: this value of err is never used (SA4006)
Error: ssl_test.go:392:5: should use !bytes.Equal(startupMessage, []byte{0, 0, 0, 0x8, 0x4, 0xd2, 0x16, 0x2f}) instead (S1004)
```

ssl.go
```diff
@@ -50,6 +50,16 @@ func ssl(o values) (func(net.Conn) (net.Conn, error), error) {
...
+		// Per RFC 6066, do not set SNI if the host is a literal IP address (IPv4
+		// or IPv6).
+		if net.ParseIP(o["host"]) == nil {
```
darinpp, line 58: There is no point of doing that check. It won't work for square bracket notation and TLS will do the determination anyway by calling `crypto.tls.hostnameInSNI`

kelvich: Good catch, thank you

kelvich:
Thanks for the review, I've fixed mentioned issues

ssl.go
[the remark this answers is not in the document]
cbandy: > It only sets SNI if the first character in the variable is `1`:
> 
> https://github.com/postgres/postgres/blob/master/src/interfaces/libpq/fe-secure-openssl.c#L1216

Hmm. I wonder if we should have a similarly silly check. `strings.HasPrefix`?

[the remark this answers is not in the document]
kelvich: > Hmm. I wonder if we should have a similarly silly check. strings.HasPrefix?

I'm not excited about `libpq` behavior around `sslsni` values but makes sense to just replicate `libpq` approach for compatibility reasons. So changed PR to use `strings.HasPrefix`. In the end, I don't think it matters a lot because nobody cares to disable SNI, except for testing purposes.

rafiss (approved):
thanks for your contribution!

pschultz:
This seems to have broken connections with sslmode=verify-ca. This setting is supposed to ignore the DNS names in the server certificate, but now connections fail with "x509: certificate is valid for x, y, z, not a".

From what I can tell, to preserve backward compatibility `tlsConf.ServerName` should not be assigned if sslsni is absent and sslmode is anything other than verify-full.

cbandy:
@pschultz I that may be #1106.

PR #1107, opened 2023-02-15
Use bytes.Buffer to read in recvMessage

Preetam:
In our testing we've noticed that attempts to connect to hosts that are not actually PostgreSQL can lead to excessive memory usage and DoS. Instead of allocating a potentially large slice up front, this PR uses a `*bytes.Buffer` to dynamically resize the slice as needed. If EOF is encountered early (like in the example below) then the program avoids allocating extremely large slices.

Here is a small program to reproduce the issue. It easily uses over 9 GB RAM locally and it takes a while for the GC to reclaim the memory.

https://go.dev/play/p/9t07FdtaW9q

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1016,17 +1016,12 @@ func (cn *conn) recvMessage(r *readBuf) (byte, error) {
 	// read the type and length of the message that follows
 	t := x[0]
 	n := int(binary.BigEndian.Uint32(x[1:])) - 4
-	var y []byte
-	if n <= len(cn.scratch) {
-		y = cn.scratch[:n]
-	} else {
-		y = make([]byte, n)
-	}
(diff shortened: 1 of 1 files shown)


johto:
Wouldn't this make everyone pay a penalty when the connection actually succeeded?

I think a more promising approach would be to do [what libpq does](https://github.com/postgres/postgres/blob/393430f57544dfd550135e0191cc91139926b682/src/interfaces/libpq/fe-connect.c#L3226) and validate the first response a bit more carefully.

Preetam:
Yes, agreed. Unfortunately I don't have time to work on this so I will be closing this PR. We have fixed this issue by requiring SSL, which does not have this problem.

PR #1112, opened 2023-03-21
allow ssl over UNIX domain sockets

drakkan:
It is not supported in Postgres 9.6

https://www.postgresql.org/docs/9.6/runtime-config-connection.html

>  SSL communication is only possible with TCP/IP connections.

but it is supported in Postgres 10 and above

https://www.postgresql.org/docs/10/runtime-config-connection.html

Here are the supported Postgres versions:

https://www.postgresql.org/support/versioning/

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,16 +9,16 @@ jobs:
       fail-fast: false
       matrix:
         postgres:
+          - '15'
+          - '14'
           - '13'
           - '12'
           - '11'
-          - '10'
-          - '9.6'
         go:
+          - '1.20'
+          - '1.19'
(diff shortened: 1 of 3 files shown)


cbandy:
I haven't looked at postgres code for this recently, but the [docs for libpq](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-PARAMKEYWORDS) say:

> `sslmode` is ignored for Unix domain socket communication.

drakkan:
you're right, this is an oversight. I was working on multi host connections and that code creates problems if one of the hosts is a unix domain socket. We have to keep that check. Thank you

PR #1113, opened 2023-03-22
CI: update supported Postgres and Go versions

drakkan:


ssl_test.go
```diff
@@ -86,6 +87,10 @@ func TestSSLVerifyFull(t *testing.T) {
 	if err == nil {
 		t.Fatal("expected error")
 	}
+	// in Go 1.20 the error is wrapped in tls.CertificateVerificationError
+	if errWrapped := errors.Unwrap(err); errWrapped != nil {
```
rafiss, line 91: could this use `errors.As` so that the test also checks that the outer error is `CertificateVerificationError`?

ssl_test.go
```diff
@@ -101,6 +106,9 @@ func TestSSLVerifyFull(t *testing.T) {
 	if err == nil {
 		t.Fatal("expected error")
 	}
+	if errWrapped := errors.Unwrap(err); errWrapped != nil {
```
rafiss, line 109: and similar for here

rafiss (changes requested):
thank you! i just had minor comments

otan:
using https://github.com/lib/pq/pull/1124

Issue #363, opened 2015-04-22
Exposing the Field and table object ID's


callesg:
Looking at:
http://www.postgresql.org/docs/current/static/protocol-message-formats.html
Scrolling down to:
RowDescription (B)

One can see that postgres does submit the id of the column & table of each result field.

looking at the function parseMeta:
https://github.com/lib/pq/blob/master/conn.go#L1401
We ignore the 6 bytes that is the column & table id could we expose these Id's in the api somehow? Or would that be hinderd by go's sql package?

uhoh-itsmaciek:
As far as I can tell, yes, that is among the many things that `database/sql` makes impossible (at least with any sensible API). Do you have any ideas on how we might expose this given the constraints of the interface?

callesg:
### It is't a pretty solution but..

Something like adding a **field_oid_hack=true** parameter to the **sql.Open** string argument.

```go
db, err := sql.Open("postgres", "user=pqgotest dbname=pqgotest field_oid_hack=true")
```

The parameter would cause the **Rows.Columns()** function to instead of just returning a column name return a comma separated string with all the interesting values.

uhoh-itsmaciek:
I think that could technically work but that's not what I would consider "a sensible API". Maybe a non-`database/sql` driver like pgx is a better fit for your use case? It [seems to support this](https://godoc.org/github.com/jackc/pgx#FieldDescription).

callesg:
No, i agree.
Had not seen that one before, definitely.

johto:
I wonder if we could change database/sql to add a method to Rows() which would allows drivers to export an `interface{}` with additional information of their choosing.  The patch should be relatively trivial, too.

johto:
I commented on https://github.com/golang/go/issues/5606 suggesting roughly the above.  The issue has been open for quite some time, though, so I'm not sure if anything will come of it.

arp242:
Closing per my comment here: https://github.com/lib/pq/pull/1117#issuecomment-3734418161 – as far as I can tell there just isn't a reasonable way to expose this through database/sql. And arguably it's not really desirable either because everything will become very database/driver specific – the entire point of database/sql is to abstract that so you can use multiple databases.

PR #1117, opened 2023-04-09
Patch for saving Table OID and Attribute Number

calledit:
When columns are renamed in sql queries, tracking the original table and column is not possible in the current implementation.

Example:
```
SELECT users.user_names AS username, posts.text AS post_text from users
LEFT INNER JOIN posts ON posts.user_id = users.id
```
In this query the original table will not be accessible, despite postgres acctually returning that information.

This patch adds two propertys to the fieldDesc struct and code to parse those fields from the postgres response.
Originally mentioned in https://github.com/lib/pq/issues/363

### DIFF
--- a/rows.go
+++ b/rows.go
@@ -11,8 +11,16 @@ import (
 const headerSize = 4
 
 type fieldDesc struct {
+	
+	//If the field can be identified as a column of a specific table, the object ID of the table; otherwise zero.
+	TableOID oid.Oid
+	
+	//If the field can be identified as a column of a specific table, the attribute number of the column; otherwise zero.
+	AtribNr int
+	
 	// The object ID of the data type.
 	OID oid.Oid
+	
 	// The data type size (see pg_type.typlen).
 	// Note that negative values denote variable-width types.
 	Len int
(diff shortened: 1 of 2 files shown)


rows.go
```diff
@@ -11,8 +11,16 @@ import (
 const headerSize = 4
 
 type fieldDesc struct {
+	
+	//If the field can be identified as a column of a specific table, the object ID of the table; otherwise zero.
```
rafiss, line 15: nit: add a space after the `//`

rows.go
```diff
@@ -11,8 +11,16 @@ import (
...
+	TableOID oid.Oid
+	
+	//If the field can be identified as a column of a specific table, the attribute number of the column; otherwise zero.
```
rafiss, line 18: nit: add a space

rafiss (changes requested):
thanks for the contribution. could you please add a test that confirms it works as expected?

arp242:
As near as I can tell these fields are never actually exposed? So I'm not sure how you would go about actually using them? I guess that's also why a test was never added?

It's not really possible to expose this through database/sql. For better or worse, it doesn't really have any good facilities for exposing very database-specific things such as this.

Let me know if I'm wrong, but as far as I can tell, this is just not possible with a database/sql compatible driver.

PR #1122, opened 2023-04-21
Add recent versions to test matrix

ptman:


otan:
1.14 failure is known, but fails 1.20 unfortunately

otan:
using https://github.com/lib/pq/pull/1124

PR #1129, opened 2023-06-09
Use and respect the passfile connection parameter

keymon:
The postgres documentation[1] regarding the password file, states
that:

password file to use can be specified using the connection parameter
passfile or the environment variable PGPASSFILE.

The current implementation of lib/pq only respects the environment
variable PGPASSFILE. This is not correct, but also limiting, as
the PGPASSFILE is global and we might want to use different files
for different clients in the same program.

Fixing that is easy, by just checking the parameter passfile first,
and if not, pull the value from PGPASSFILE.

[1] https://www.postgresql.org/docs/current/libpq-pgpass.html

How to test
-----------

Start a postgres
```
docker run --name libpq-postgres \
    --net host \
    -e POSTGRES_DB=pqgotest \
    -e POSTGRES_USER=$(whoami) \
    -e POSTGRES_HOST_AUTH_METHOD="trust" \
    -d postgres
```

Run the test `go test -run TestPgpass -v .`

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -233,11 +233,8 @@ func (cn *conn) handlePgpass(o values) {
 	if _, ok := o["password"]; ok {
 		return
 	}
-	// Get passfile from the options, if empty, get it from envvar
+	// Get passfile from the options
 	filename := o["passfile"]
-	if filename == "" {
-		filename = os.Getenv("PGPASSFILE")
-	}
 	if filename == "" {
 		// XXX this code doesn't work on Windows where the default filename is
 		// XXX %APPDATA%\postgresql\pgpass.conf
@@ -2042,6 +2039,8 @@ func parseEnviron(env []string) (out map[string]string) {
 			accrue("user")
 		case "PGPASSWORD":
 			accrue("password")
+		case "PGPASSFILE":
+			accrue("passfile")
 		case "PGSERVICE", "PGSERVICEFILE", "PGREALM":
 			unsupported()
 		case "PGOPTIONS":
--- a/conn.go
+++ b/conn.go
@@ -233,7 +233,11 @@ func (cn *conn) handlePgpass(o values) {
 	if _, ok := o["password"]; ok {
 		return
 	}
-	filename := os.Getenv("PGPASSFILE")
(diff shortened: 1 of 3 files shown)


conn.go
```diff
@@ -233,7 +233,11 @@ func (cn *conn) handlePgpass(o values) {
 	if _, ok := o["password"]; ok {
 		return
 	}
-	filename := os.Getenv("PGPASSFILE")
+	// Get passfile from the options, if empty, get it from envvar
```
rafiss, line 236: i think the correct place to handle the env var is here: https://github.com/lib/pq/blob/381d253611d666974d43dfa634d29fe16ea9e293/conn.go#L2009

then the logic after `parseEnviron` is called will automatically take care of making the connection string parameter override the env var: https://github.com/lib/pq/blob/381d253611d666974d43dfa634d29fe16ea9e293/connector.go#L61

rafiss (changes requested):
thanks for your contribution!

conn_test.go
```diff
@@ -221,8 +222,11 @@ localhost:*:*:*:pass_C
 	// localhost also matches the default "" and UNIX sockets
 	assertPassword(values{"host": "", "user": "some_user"}, "pass_C")
 	assertPassword(values{"host": "/tmp", "user": "some_user"}, "pass_C")
+	// passfile connection parameter takes precedence
+	os.Setenv("PGPASSFILE", "/tmp")
+	assertPassword(values{"host": "server", "dbname": "some_db", "user": "some_user", "passfile": pgpassFile}, "pass_A")
+
 	// cleanup
-	os.Remove(pgpassFile)
```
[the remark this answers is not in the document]
keymon: it is in a defer next to where the file is created

conn.go
keymon: I see you are right, it makes sense. I didn't see that I focused in that function. Pushed a new commit.

keymon:
> thanks for your contribution!
Thank you for looking at it and the review!

rafiss (approved):
thank you!

PR #1133, opened 2023-07-10
use build tag to not build ssl_permissions.go on Plan9

matloob:
That file uses syscall.Stat_t, which is not defined on Plan 9.


Arusekk:
Hi, is there anything missing in this PR? If not, can it get merged?

PR #1140, opened 2023-10-16
ssl: allow sslinline without client certs

bmon:
Currently, when using sslinline=true, client certificates are always attempted to be parsed
even when not provided. This differs from the behaviour when sslinline is not used, where
client certificates are optional and will not be set if not specified.

This change allows connections to be established when sslinline is used but client certificates
are not provided.

A use-case for this would be verifying the database SSL via `verify-ca` and inlining sslrootcert,
while still leaving client certificates unspecified.


evilensky:
@bmon Just ran into this very issue and came to make the same fix :) Should a unit test be created for this option? Is there anything as an outsider I could do to help get this merged?

PR #1150, opened 2024-01-11
Remove arbitrarly chosen extra_float_digits connection option.

orian:


johto:
1. Why do we want to remove this?
2. Not arbitrarily chosen.

orian:
It caused me some problems and I would prefer the driver to stick to the database default (or at least document it).

After some thought, I think documenting it is ok.

I'm gonna close the PR, I've forked the lib and using my fork.

arp242:
Fixed via #1212

Issue #914, opened 2019-10-29
Use a safe implementation of SCRAM.


mberhault:
The [scram](https://github.com/lib/pq/tree/master/scram) implementation has no unittests and ignores parts of the RFC (eg: the `m` field is supposed to trigger an authentication failure). This should be replaced with a fuller (and better tested) implementation. https://github.com/xdg-go/scram might be a candidate.

Neustradamus:
Note that SCRAM-SHA-256 has been added in:
- https://github.com/lib/pq/pull/833

Linked to:
- https://github.com/lib/pq/issues/817
- https://github.com/lib/pq/issues/914
- https://github.com/lib/pq/issues/941
- https://github.com/lib/pq/pull/608
- https://github.com/lib/pq/pull/621
- https://github.com/lib/pq/pull/788
- https://github.com/scram-sasl/info/issues/1

gmbuell:
I audited current master `10237a6` against PostgreSQL's SCRAM behavior and found three concrete cases that fit this umbrella issue:

- The client responds with SCRAM-SHA-256 even when the mechanism list is unterminated, empty, or advertises only SCRAM-SHA-256-PLUS: [`TestProtocolRegressionSASLMechanismNegotiation`](https://github.com/gmbuell/pq/blob/1c9dd44e4cd92b7b48838b20618e55617b40078f/protocol_auth_state_regression_test.go#L48).
- Iteration parsing accepts nonpositive and absurdly large counts while rejecting some short positive counts: [SCRAM iteration regressions](https://github.com/gmbuell/pq/blob/1c9dd44e4cd92b7b48838b20618e55617b40078f/scram/regression_test.go#L23).
- Password preparation differs from PostgreSQL's Unicode 3.2 SASLprep mapping, normalization, rejection ordering, and raw fallback: [`TestRegressionPostgreSQLSASLprep`](https://github.com/gmbuell/pq/blob/1c9dd44e4cd92b7b48838b20618e55617b40078f/scram/regression_test.go#L102).

These focused tests were written against current upstream and fail before the candidate changes in [the audit fix](https://github.com/gmbuell/pq/commit/96963c8acb8ca25807fb238b542bab4105ddce6e).

Neustradamus:
@gmbuell: Thanks for your message!

Can you talk about it here: https://www.postgresql.org/list/
- pgsql-hackers@postgresql.org
- pgsql-bugs@lists.postgresql.org

PR #1152, opened 2024-02-08
Fix scram.*Client.Step so it returns an "ok" flag consistently

momeni:
The package documentation asks to call Step method as while as it returns true, so it has to return false only on the first error or if all expected steps were taken previously.
This condition was implemented correctly in the first if-condition where it returned false when there were no more steps or an error was detected.
However, the ultimate return statement and the method godoc used to return false as while as the Step should be called again. This commit resolves this inconsistency and updates the relevant godocs.

```go
-       return c.step > 2 || c.err != nil
+       return c.step <= 2 && c.err == nil
```

The RFC 5802 example test case is added too.

### DIFF
--- a/scram/rfc5802_consts_test.go
+++ b/scram/rfc5802_consts_test.go
@@ -1,0 +1,50 @@
+// Copyright (c) 2010 IETF Trust and the persons identified as the
+// document authors. All rights reserved. Redistribution and use in source
+// and binary forms, with or without modification, are permitted provided
+// that the following conditions are met:
+//
+//   - Redistributions of source code must retain the above copyright
+//     notice, this list of conditions and the following disclaimer.
+//   - Redistributions in binary form must reproduce the above copyright
+//     notice, this list of conditions and the following disclaimer in
+//     the documentation and/or other materials provided with the
+//     distribution.
+//   - Neither the name of Internet Society, IETF or IETF Trust, nor the
+//     names of specific contributors, may be used to endorse or promote
+//     products derived from this software without specific prior
+//     written permission.
+//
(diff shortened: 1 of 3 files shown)


arp242:
Does this fix a bug or change the function? This is an exported function, so changing it is a non-starter as it's not a compatible change.

momeni:
It is a bugfix. Function has two return statements which were coded inconsistently.

arp242:
Okay. What does it fix? Because the test you added also passes without this change.

scram.go was added in #833, and was taken from https://github.com/go-mgo/mgo/blob/v2/internal/scram/scram.go (also used in the fork https://github.com/globalsign/mgo/blob/master/internal/scram/scram.go). It seems unlikely that it can be used in three different projects for over 10 years and be completely wrong?

arp242:
Closing this as I'm not really confident this is correct; the scram package is an ad-hoc untested implementation, but it does work, has worked for over 10 years, and I'm hesitant to change it.

I'll also add that using the scram package outside of pq is completely unsupported. This really should have been in internal/scram. When I replace it at some point (#914) I'll mark the entire package as deprecated and will remove it in v2.

momeni:
I added a test case which fails with old code in 7751823 commit.
At first step, `c.step` is zero and becomes one by `c.step++` statement. So final `return c.step > 2 || c.err != nil` statement will return false. Therefore, `for client.Step(in) {` will break early and second step cannot be executed.

Current usage of `client.Step(in)` does not trigger that bug because it is not called in a `for` loop. Instead, file `conn.go` lines 1305, 1327, and 1347 call the `sc.Step(nextStep)` three times and ignore its boolean return value.

PR #1160, opened 2024-05-23
Allow to change (or disable) the default driver name for registration

demoManito:
## Description
A link variable now allows to change or disable the name of the driver that is automatically registered with `database/sql`.
This allows users to give the `postgres` name to another `database/sql` driver.

The implementation is the same as in: https://github.com/go-sql-driver/mysql driver. See https://github.com/go-sql-driver/mysql/blob/af8d7931954ec21a96df9610a99c09c2887f2ee7/driver.go#L92

### Usage
Change the driver name to `custom`:
```
go build "-ldflags=-X github.com/lib/pq.driverName=custom"
```

Disable the automatic driver registration (set `driverName` to an empty string):
```
go build "-ldflags=-X github.com/lib/pq.driverName="
```

In the same way, a variable overridable at link time is also provided to override the driver name used in the test suite. This allows to run our test suite on another driver.
```
go build "-ldflags=-X github.com/lib/pq.driverNameTest=custom"
```

`driverName` is propagated to `driverNameTest` unless `driverNameTest` is explicitely defined.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -57,8 +57,19 @@ func (d Driver) Open(name string) (driver.Conn, error) {
 	return Open(name)
 }
 
+// This variable can be replaced with -ldflags like below:
+// go build "-ldflags=-X github.com/lib/pq.driverName=custom"
+var driverName = "postgres"
+
 func init() {
-	sql.Register("postgres", &Driver{})
+	if driverName != "" {
+		sql.Register(driverName, &Driver{})
+	}
+}
+
+// DriverName return default driver name
+func DriverName() string {
+	return driverName
 }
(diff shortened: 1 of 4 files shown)


arp242:
Seems okay at a glance, but what is the use case for this? It's not mentioned in the PR, or in the PR for the MySQL driver (https://github.com/go-sql-driver/mysql/pull/1499).

arp242:
I'm closing this for now as it's unclear what problem this addresses or why we want to do this, so it's also not clear if this is the best way to address it.

Can reopen if more details are added.

PR #1161, opened 2024-06-03
Enforce integer limits of the Postgres wire protocol

paul-gerste-sonarsource:
## Problem
The Postgres wire protocol has certain fields that contain fixed-size integers, such as the 32-bit message length ([see here](https://www.postgresql.org/docs/current/protocol-message-formats.html)). This library does not check if values fit into these fixed-size fields before writing them to the buffer. This can lead to the truncation of the values being written, causing corrupted messages.

Example: If the message to be sent is `4**32 + 4` bytes long, the length cannot fit into the 4-byte length field. The value written to the buffer will be `\x00\x00\x00\x04`, which gets decoded to `4` by the Postgres database when it receives such a message. Therefore, the Postgres database will think the message has a length of 4 and will try to read the next message after those 4 bytes.

After such a malformed message, the client and the database have different understandings of where messages start and end. In the best case, this causes a connection abort due to a parsing error. In the worst case, this leads to the execution of malicious SQL statements that an attacker has injected into a large payload that ends up in an SQL query.

## Solution
I added size checks that error if the value being written is too large for its field. Since the functions that I changed don't return errors, I had to `panic()` in the error case. I opted for this route because I noticed that other functions do the same ([example 1](https://github.com/lib/pq/blob/3d613208bca2e74f2a20e04126ed30bcb5c4cc27/encode.go#L85), [example 2](https://github.com/lib/pq/blob/3d613208bca2e74f2a20e04126ed30bcb5c4cc27/encode.go#L373)). Please let me know if there's a different error-handling mechanism I should use; then I'll amend the PR.

### DIFF
--- a/buf.go
+++ b/buf.go
@@ -3,6 +3,7 @@ package pq
 import (
 	"bytes"
 	"encoding/binary"
+	"math"
 
 	"github.com/lib/pq/oid"
 )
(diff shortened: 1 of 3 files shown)


arp242:
Thanks; merged via #1209; I wanted to update the commit message and tweak the errors, but I can't update this PR since it's created from the fork's master branch (you're still recorded as the author in the git commit).

PR #1162, opened 2024-06-27
avoid excessive array wrapping

toqueteos:
`pq.Array` shouldn't wrap values that are already an optimal pq Array type.

```go
x := []string{"a", "b", "c"}
pq.Array(x) // is (*pq.StringArray)
pq.Array(pq.Array(x)) // is (pq.GenericArray)
```

Repro: https://go.dev/play/p/piFLACJYhC2

### DIFF
--- a/array.go
+++ b/array.go
@@ -60,9 +60,26 @@ func Array(a interface{}) interface {
 		return (*StringArray)(a)
 	case *[][]byte:
 		return (*ByteaArray)(a)
+
+	case *BoolArray:
+		return a
+	case *Float64Array:
+		return a
+	case *Float32Array:
+		return a
+	case *Int64Array:
+		return a
+	case *Int32Array:
+		return a
+	case *StringArray:
+		return a
+	case *ByteaArray:
+		return a
+	case *GenericArray:
+		return a
 	}
 
-	return GenericArray{a}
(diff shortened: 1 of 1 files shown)


dpordomingo (approved):

arp242:
I would expect `pq.Array(pq.Array(pq.Int32Array{1,2,3}))` to be a multi-dimensional array like `{{1,2,3}}`. That doesn't really work (yet), but I wouldn't really expect it to be "silently" flattened to `{1,2,3}`.

In general it's unclear to me what problem or use case this solves?

toqueteos:
The reality is that multi-dimensional arrays are very underutilized in the _real world_ and while correct that approach leads to serialization issues for the most common scenario.

Which is how I found our DIY _orm-like_ layer was sometimes wrapping already-wrapped arrays.

No matter if this change is accepted or not, the docs should be updated to reflect the decision to support multi-dimensional arrays or to flatten.

arp242:
> The reality is that multi-dimensional arrays are very underutilized in the real world and while correct that approach leads to serialization issues for the most common scenario.

Well maybe, but it seems rather inappropriate to start making opinionated decisions about that in a PostgreSQL driver – that should support whatever PostgreSQL supports, no?

PR #1166, opened 2024-08-04
Support float64/32 type

nhdms:


rows.go
```diff
@@ -29,6 +29,10 @@ func (fd fieldDesc) Type() reflect.Type {
 		return reflect.TypeOf(int32(0))
 	case oid.T_int2:
 		return reflect.TypeOf(int16(0))
+	case oid.T_float8:
+		return reflect.TypeOf(float64(0))
```
Grax01, line 33: what is the use of adding this can you please explain, does it help when creating a query in any way?

PR #1179, opened 2024-11-16
Bugfix: use filepath package when connecting to UNIX sockets

stapelberg:
The code before this commit is non-portable: it only works on UNIX systems where path.Join() is close enough to filepath.Join(). However, path.Join() is the wrong choice: UNIX domain sockets are identified via their file system name, so — conceptually and in practice — we need to use the filepath package instead.

This commit fixes connecting to a PostgreSQL database that is listening on a UNIX domain socket on Windows. If you are surprised to see UNIX domain sockets on Windows, they were introduced with Windows 10, i.e. are widely available now: https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/

filepath.IsAbs() is equivalent to strings.HasPrefix(s, "/") on UNIX, but also works on Windows, where absolute paths start with e.g. C:\ Similarly, filepath.Join() uses the correct path separator.

related to https://github.com/zombiezen/postgrestest/issues/3

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -15,7 +15,6 @@ import (
 	"net"
 	"os"
 	"os/user"
-	"path"
 	"path/filepath"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


stapelberg:
BTW, I verified that the `github.com/jackc/pgx` package behaves correctly on Windows when connecting to UNIX domain sockets. I think it would be good to fix this issue in lib/pq as well :)

johto (approved):
LGTM.  I can push this next week unless someone beats me to it.

cbandy:
👍🏻 filepath

🌱 Should this code consider abstract sockets, too?

stapelberg:
Thanks for the quick review!

> 🌱 Should this code consider abstract sockets, too?

That’s a good point. I wasn’t sure if PostgreSQL actually supported abstract sockets, but yes, indeed it does: https://www.postgresql.org/docs/current/runtime-config-connection.html#GUC-UNIX-SOCKET-DIRECTORIES

I updated the Pull Request accordingly and verified that both listening and connecting to abstract UNIX domain sockets works on Linux.

Issue #1145, opened 2023-11-17
RFC 9266: Channel Bindings for TLS 1.3 support


Neustradamus:
Can you add the support of RFC 9266: Channel Bindings for TLS 1.3?
- https://datatracker.ietf.org/doc/html/rfc9266 

Little details, to know easily:
- tls-unique for TLS =< 1.2
- tls-server-end-point
- tls-exporter for TLS = 1.3 

I think that you have seen the jabber.ru MITM and Channel Binding is the solution:
- https://notes.valdikss.org.ru/jabber.ru-mitm/
- https://snikket.org/blog/on-the-jabber-ru-mitm/
- https://www.devever.net/~hl/xmpp-incident
- https://blog.jmp.chat/b/certwatch

Thanks in advance.

Linked to:
- https://github.com/lib/pq/issues/817
- https://github.com/lib/pq/issues/914
- https://github.com/lib/pq/issues/941
- https://github.com/lib/pq/pull/608
- https://github.com/lib/pq/pull/621
- https://github.com/lib/pq/pull/788
- https://github.com/lib/pq/pull/833
- https://github.com/scram-sasl/info/issues/1

Neustradamus:
@conradludgate has done a PR here:
- https://github.com/lib/pq/pull/1181

We can thank @conradludgate for this good job about tls-server-end-point!
It does not add tls-exporter yet but a big improvement!

cc: @lib team, @fdr, @otan, @johnto, @rafiss.

PR #1181, opened 2024-11-24
feat: support tls-server-end-point channel binding for SCRAM-SHA-256

conradludgate:
Postgres supports "channel bindings" for SCRAM-SHA-256 authentication. This prevents certain MITM attacks. This authentication is known as "SCRAM-SHA-256-PLUS". 
* https://datatracker.ietf.org/doc/html/rfc5802#section-6
* https://www.postgresql.org/docs/17/sasl-authentication.html#SASL-SCRAM-SHA-256

Postgres supports the "tls-server-end-point" channel binding form which is derived by hashing the raw bytes of the first certificate sent by the server in the TLS handshake. 
* https://datatracker.ietf.org/doc/html/rfc5929#section-4

Postgres also supports "requiring" channel binding. This additionally prevents a MITM sending cleartext the chosen authentication mechanism. 
* https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-CHANNEL-BINDING

Related to #1145

### DIFF
--- a/buf.go
+++ b/buf.go
@@ -38,6 +38,20 @@ func (b *readBuf) string() string {
 	return string(s)
 }
 
+func (b *readBuf) strings() []string {
+	ss := []string{}
+	for (*b)[0] != 0 {
+		i := bytes.IndexByte(*b, 0)
+		if i < 0 {
+			errorf("invalid message format; expected string terminator")
+		}
+		s := (*b)[:i]
+		*b = (*b)[i+1:]
+		ss = append(ss, string(s))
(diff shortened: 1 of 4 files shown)


Neustradamus:
Dear @conradludgate,

I have discovered today by chance your PR, one year after, nice, good job! :)

Dear @lib team, @fdr, @otan, @johnto, @rafiss: Have you seen this PR?

It solves but not fully:
- https://github.com/lib/pq/issues/1145

PR #1188, opened 2025-04-28
Remove assumption that the auth response is AuthenticateOk

yoshihikoueno:
This PR aims to solve an issue where lib/pq fails to connect to a PostgreSQL cluster via PgPool-II.

# Issue description
lib/pq currently assumes the message sent from the server after sending PasswordMessage to be AuthenticationOk.
However, this assumption doesn't always stand as the server may send some other messages other than AuthenticationOk such as another authentication request.

That occurs when lib/pq is trying to connect to a PgPool-II server, where the following procedure is required as PgPool-II wants to have the password in both Cleartext and MD5 formats:

[Client -> Server] StartupMessage
[Server -> Client] AuthenticationCleartextPassword
[Client -> Server] PasswordMessage
[Server -> Client] AuthenticationMD5Password
[Client -> Server] PasswordMessage
[Server -> Client] AuthenticationOk
[Server -> Client] ReadyForQuery

# Solution description
Current implementation of auth function of lib/pq will fail if the message after sending credentials is not AuthenticationOk, but we can just return before checking it and let another loop of auth check it instead. 
The official client libpq (and of course softwares that use it such as psql and pgAdmin) does support handling multiple authentication requests, thus it can connect to PgPool-II just fine with the procedure above. So making lib/pq check the message sent from the server after PasswordMessage will solve the described issue and align its behavior with the official client.

Thank you for checking out this PR. Please let me know if there's anything that is missing or needs improvements with this PR.


arp242:
It would be great to have some tests for this. I added pgpool to the CI with #1218, but unfortunately this entire auth() function is never tested at the moment (code is always 0).

arp242:
Added tests in #1222.

yoshihikoueno:
@arp242 
Thank you for taking the time to review and merge this. Much appreciated!

PR #1191, opened 2025-06-05
fix TLS tests on newer go versions and MacOS

zregvart:
The TLS tests in `ssl_test.go` assert that specific errors are reported for specific situations. This fails on newer (1.20+) versions of go since `tls.CertificateVerificationError` is returned instead of the expected, contrary to older versions where `x509.UnknownAuthorityError` or `x509.HostnameError` would be returned.

This maintains the same assertions on go < 1.20, and allows for the new error on go >= 1.20.

In addition to that, on MacOS the error string `x509: “postgres” certificate is not standards compliant` is returned instead of any of these, this seems to be due to the version of OpenSSL provided by MacOS, see https://github.com/golang/go/issues/51991 for details.


arp242:
I can't test myself as I don't have a macOS machine, but seems good.

Should really test macOS in the CI – right now only Linux is tested.

PR #1195, opened 2025-06-13
fix: `pq:unnamed prepared statement does not exist` with pgbouncer

### LINKED ISSUE #889: unnamed prepared statement does not exist ERROR on production/staging servers
We started implementing pgbouncer on our stage and prod nodes and we started seeing `pq:unnamed prepared statement does not exist` error, it would be great if i can get some help on the same to understand where to fix this.

exfly:
Fixes #889
Thanks to: #759


arp242:
What's the difference between this PR and  #759?

arp242:
Closing because this seems a duplicate of #759.

PR #1196, opened 2025-06-21
feat: manage range and multirange

thecampagnards:
Hello,
This is a pr to manage range and multi range using generics (compare to https://github.com/lib/pq/pull/612).
ie postgres https://www.postgresql.org/docs/current/rangetypes.html doc.
Thx !


arp242:
I don't think this should do `//go:linkname convertAssign database/sql.convertAssign`.

Related: https://github.com/golang/go/issues/62146

Issue #1156, opened 2024-04-04
pq: invalid array literal.


wnesbv:
in the postgres database
`{"2024-04-05 00:00:00","2024-04-10 00:00:00"}
`
`

    func qSchBkg(w http.ResponseWriter, conn *sql.DB,start time.Time,end time.Time) (rows *sql.Rows, err error) {

        fmt.Println("qSchBkg start.. ", start)

        // start,end print 2024-04-05 00:00:00 +0000 UTC 2024-04-10 00:00:00 +0000 UTC

        rows,err = conn.Query("SELECT * FROM booking WHERE days_booking BETWEEN $1 AND $2", start,end)

        if err != nil {
            switch {
                case true:
                fmt.Fprintf(w, "Error: Query()..! : %+v\n", err)
                break
            }
            return
        }
        return rows,err
    }
`
Error: Query()..! : pq: invalid array literal: "2024-04-05 00:00:00Z"

arp242:
The "invalid array literal" error comes from PostgreSQL, not pq. I can't really reproduce this in a test. Unfortunately there's not enough information in here to reproduce it, but I suspect they did something wrong somewhere with some query.

Since this is a few years old and the creator has not been active on GitHub for several years, I'll just go ahead and close it, but feel free to come back with more details if you're still having problems and we can re-open.

Issue #530, opened 2016-10-31
lib/pg assumes that all queries that start with "copy" (case-insensitive) are special bulk import queries


flgr:
lib/pg special cases queries starting with "copy" (case-insensitive) to a code path for bulk imports from passed in data. That's nice and all, but unfortunately it has a false positive.

RedShift (pg-compatible, mostly) has a special statement "copy" that is used to bulk import data from S3 and stuff (http://docs.aws.amazon.com/redshift/latest/dg/r_COPY.html)

Unfortunately such copy statements will end up in the wrong code path in lib/pg:
https://github.com/lib/pq/blob/ae8357db35d721c58dcdc911318b55bef6b1b001/conn.go#L768

I think you could either use invalid syntax such as "*copy" to do this, or use a magic inline comment in the query, e.g. "/* bulk */copy".

My work around until this is fixed upstream is prefixing RedShift copy queries with a space...

freeformz:
All of the AWS COPY statements require a `credentials '....'` clause. We could look for that as well and if present handle it like a normal query. I'm not overly fond of that, but it works (at least until Postgres' COPY also takes a `credentials` clause.

freeformz:
There is also the sentinel that each of the from locations are AWS specific:

`from 's3://....'`
`from 'emr://...'`
`from 'dynamodb://...'`

freeformz:
postgres would interpret those from locations as file paths and error AFAICT.

freeformz:
I guess the real question is, how much other redshift stuff is non standard and would be required for full support and should any of this be included in lib/pq ?

flgr:
From what I can tell other than that issue things seem to be working perfectly fine!

Von meinem iPhone gesendet

cbandy:
> I guess the real question is, how much other redshift stuff is non standard and would be required for full support and should any of this be included in lib/pq ?

@freeformz do you mean

1. specifically related to `COPY`? or
2. more generally that lib/pq take some stance on Redshift compatibility?

@flgr #508 is about how Redshift requires TLS negotiation after some amount of traffic.

uhoh-itsmaciek:
I don't suppose there's a dev Redshift available for testing? It's unfortunate that we can't actually add Travis tests to ensure we maintain compatibility (if we do decide to support it).

arp242:
According to https://docs.aws.amazon.com/redshift/latest/dg/r_COPY.html

[...]

Issue #213, opened 2013-12-29
Better support for COPY operations


drjokepu:
The approach discussed and implemented in #181 works well and it is very useful, however, it has a number of problems:
1. It doesn’t (and maybe can’t) support COPY FROM operations.
2. It doesn’t support the binary format.
3. It doesn't support piping in data directly from an external source, data has to be provided row by row, cell by cell.
4. It reuses the existing Go SQL interface in non-trivial, confusing ways. In particular, it makes Prepare() and Exec() do things they are not meant (and documented) to do.
5. It decides whether a statement is a COPY statement or not by looking at the first four characters of the statement text and comparing it to the "COPY" word in a non case-sensitive manner. This will work most of the time but it’s not too difficult to imagine scenarios where this will fail to identify a copy operation (or maybe even identify a non-copy operation as one). The right way to do this is by parsing the query text into an AST and the right place to do that is in the backend. The front-end is simply not the right place to make decisions about the nature of queries.

In my opinion, the first three are very real shortcomings that need to be looked at, number 4 is an inconvenience the very least and number 5 is the least important as it’s only an edge case.

To address these shortcomings, I propose the following approach (which is basically an adaptation of [Npgsql’s approach](http://npgsql.projects.pgfoundry.org/docs/api/index.html), something I have been using for years with much success):
- Two structs: CopyIn, CopyOut (or CopyInWriter and CopyInReader), each conforming to io.Writer and io.Reader respectively.
- Both structs maintaining a reference to the Stmt of the copy operation.
- The operation is started with copyIn.Start() or copyOut.Start()
- Then data is written to / read from these structs via the io.Writer and io.Reader interfaces or via the aid of a number of higher-level helper functions
- copyIn.End() / copyOut.End() will complete the operation.

This is a somewhat different approach than what is currently implemented but I’m convinced that it will make the support for copy operations more complete and the application code more maintainable. The downside is that it cannot be made compatible with the current approach and that it would require the application code to work with the driver directly (although that is something that’s already being done when using the existing CopyIn() and CopyInSchema() functions). A possible way to mitigate the downsides is to keep around the old approach, temporarily or permanently.

I’m keen to get this thing working (as I need a solution to problems 1 and 2) so unless there is serious opposition to this, I will start working on this immediately.

PR #1198, opened 2025-08-12
Detect COPY even if the query starts with whitespace or comments

### LINKED ISSUE #1154: Exectuing pq.CopyIn fails with one of two errors depending on usage
Query is
* defined as `query := pq.CopyIn("public.foo", "col_foo", "col_bar")`
* prints as `COPY "public.foo" ("col_foo", "col_bar") FROM STDIN`.

Executing the prepared statement 
* prepared via `tx.Prepare(query)`
* `tx` being defined as `tx, err := db.Begin()`
* `db` is of type `sqlx.DB`, a connection established via `sqlx.Connect` if that matters 

via `_, err := statement.Exec("foo", "bar")` returns either

* `sql: expected 0 arguments, got 2`
    * This occurs when I use "pgx" as the driver when establishing the connection.
* `pq: relation "public.foo" does not exist` when preparing
    * This occurs when I use "postgres" (pq) as the driver when establishing the connection. and is infinitely more puzzling, because not only does the relation exist when checked manually (`SELECT * FROM public.foo;` works fine in Postgres shell), but the application also contains a `SELECT` statement that leftjoins on `public.foo`, which works fine and returns no errors.

I don't really see a reason for why this would happen, no relevant issues for either of the two points. I'd prefer to use `pq`, because `pgx`'s `CopyFrom` straight up does not support usage from transactions for whatever reason and requires a direct access to connection.

Let me know if I can provide anything else to elaborate.

darccio:
# What this PR does

Updated the handling of the `COPY` statement in the Prepare function to utilize the `internal.StartsWithCOPY` utility to detect if a query is a `COPY` statement even if it has comments or whitespaces.

# Motivation

Our `Datadog/dd-trace-go` library supports instrumenting `database/sql`, including the DBM-APM link based on `SQLcommenter` spec, with the difference that we are prepending the comment instead of appending it. This breaks how `lib/pq` detects a `COPY` statement.

We are aware that there were previous opened issues around `COPY` statements, but we feel this one is an improvement, as SQL comments are not uncommon in the wild. Also, it's the only place where this kind of check happen in `lib/pq`.

This would fix #1154, although it might break the workaround described in #530. We didn't attempt to address the concerns explained in #213.

# Benchmarks

```
goos: darwin
goarch: arm64
pkg: github.com/lib/pq/internal
cpu: Apple M1 Max
BenchmarkStartsWithCOPY-10    	39674830	        28.50 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/lib/pq/internal	1.363s

goos: darwin
goarch: arm64
pkg: github.com/lib/pq/internal
cpu: Apple M1 Max
BenchmarkEqualFold-10    	292420224	         3.771 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/lib/pq/internal	1.721s
```

Although it's almost eight times slower at the nanoseconds scale, it should negligible in the context of preparing a query.


arp242:
I put my version here: https://github.com/lib/pq/commit/2e1e922c6f9a011be758018521ef686b6600dcd1 – personally I think it's rather easier to follow/understand.

Performance is a few ns slower:

	goos: linux
	goarch: amd64
	pkg: github.com/lib/pq/internal
	cpu: AMD Ryzen 7 PRO 7840U w/ Radeon 780M Graphics
	BenchmarkStartsWithCOPY-16      61637460        16.39 ns/op    0 B/op  0 allocs/op
	BenchmarkStartsWithCOPY2-16     52043742        19.92 ns/op    0 B/op  0 allocs/op
	BenchmarkEqualFold-16           459326388        2.569 ns/op   0 B/op  0 allocs/op

But that's okay.

What do you think?

arp242:
> This would fix #1156 

How would this fix that issue by the way?

darccio:
Thanks for taking care, @arp242. I'll check this tomorrow morning EU time.

darccio:
@arp242 Re: #1156, I think you meant #1154. In any case, your commit 2e1e922 behaves in the same way than my implementation. If you feel yours is easier to maintain, let's go forward with yours.

arp242:
> Re: #1156, I think you meant #1154

Eh, yeah; that stupid obnoxious dropdown probably mangled my comment. But I don't see how this PR fixes that issue?

darccio:
@arp242 I vaguely recall that I found the same messages in #1154, and I was convinced that it worked at the moment. Let's keep it open just in case.

What are our next steps with this PR? As we are pushing forward your solution, I guess it can be closed.

arp242:
> What are our next steps with this PR? As we are pushing forward your solution, I guess it can be closed.

I updated the PR and will merge when ready. I want to look a bit at adding a higher-level test and some documentation on what this actually does. One reason for that is that your PR didn't actually do anything as it still retained the string indexing:

	if len(q) >= 4 && internal.StartsWithCOPY(q[:4]) {

So on `"/* cmt */ copy"` it would just send `/* cm`.

darccio:
You are right, good catch @arp242! Thank you for your time.

darccio:
@arp242 How often `lib/pq` is released? I'd like to revert [a PR in `dd-trace-go`](https://github.com/DataDog/dd-trace-go/pull/3871) to support `COPY` using `lib/pq` in our DBM product.

arp242:
I'm hoping to do a release next week or thereabouts.

PR #1205, opened 2025-12-31
Test newer PostgreSQL and Go versions in CI

arp242:
- Run some newer PostgreSQL versions and remove old EOL PostgreSQL versions. PostgreSQL 18 fails with:

      /init/hba.sh: line 1: /var/lib/postgresql/data/pg_hba.conf: No such file or directory

  Fixing that is for a future PR

- Update the Go versions to test, and set minimum version to 1.18, which seems like a reasonable version to use in 2025 (almost 2026). Only test the first and latest versions, to avoid spawning a million jobs.

- Update staticcheck integration.

- Had to fix the timezone name to make the tests work; US/Eastern is a (deprecated) alias for America/New_York.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,27 +3,21 @@ name: Test
 on: [push, pull_request]
 
 jobs:
+  staticcheck:
+    name:    'staticcheck'
+    runs-on: 'ubuntu-latest'
+    steps:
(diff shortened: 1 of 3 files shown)


otan:
I removed blocking reviews - lmk if it doesn't work

arp242:
Cheers, I thought it might be a bit presumptuous to ask 😅

PR #1235, opened 2026-01-19
fix(kerberos): Force TGS Exchange

disq:
Not a krb expert and this is yet to be tested, but I think this is missing.


arp242:
> Not a krb expert and this is yet to be tested, but I think this is missing.

This does not exactly inspire great confidence in its correctness, especially not without any description of what this does or why this should be there.

disq:
Looks like `AcquireCred` just calls `AffirmLogin` which in turn can call `Login` if needed, but `init()` already does that. My bad.

PR #1240, opened 2026-01-20
Add Config, NewConfig,  NewConnectorConfig

### LINKED ISSUE #856: Open() DSN has race condition when using "database" rather than "dbname"
The following works reliably:

```go
db, err := sql.Open("postgres", "dbname=my_db")
if err == nil {
  err = db.Ping()
}
// err should be nil if my_db exists
```

However the following will randomly use the value of the `PGDATABASE` environment variable instead of the given DSN.

```go
db, err := sql.Open("postgres", "database=my_db")
if err == nil {
  err = db.Ping()
}
// err should be nil if my_db exists
```

This seems to be due to `lib/pq` implicitly relying on predictable map iteration internally, which Go does not provide.

### LINKED ISSUE #879: If password is "" in dsn, I get a error options after parseOpts
test code like this:
```
        o := make(map[string]string)
	dsn := fmt.Sprintf("host=%s port=%d user=%s password=%s dbname=%s sslmode=disable",
		"localhost",
		5432,
		"test",
		"",
		"testDB",
	)
	parseOpts(dsn, o)
	fmt.Println(o)
```

output:
``` 
map[host:localhost password:dbname=testDB port:5432 sslmode:disable user:test]
```

in this map, we can see the passwd value is that "dbname=testDB" rather than "" 

In this case，it is a wrong dsn parser？

Or when my passwd is null, I can only put passwd end of dsn，like this：
```
dsn := fmt.Sprintf("host=%s port=%d user=%s dbname=%s sslmode=disable password=%s",
		"localhost",
		5432,
		"test",
		"testDB",
		"",
	) 
```

### LINKED ISSUE #909: MySQL Config analog
What about MySQL Config analog?
https://godoc.org/github.com/go-sql-driver/mysql#Config

It could be helpful in case you need to build DSN on the fly?

### LINKED ISSUE #1202: make parseOpts public
It is not uncommon to see code that need to parse the DSN to extract the DB name to `CREATE DATABASE` when it does not exist. ParseOpts can be used to retrieve such parameter. Any design concern about exporting `parseOpts` ?

arp242:
This adds a Config struct and NewConfig() to create it from a DSN and environment. With this, you can do something like:

	cfg, err := pq.NewConfig(os.Args[1])
	if err != nil {
		return err
	}
	if cfg.Host == "localhost" {
		cfg.Host = "127.0.0.1"
	}

	c, err := pq.NewConnectorConfig(cfg)
	if err != nil {
		t.Fatal(err)
	}

	db := sql.OpenDB(c)
	defer db.Close()

It also gives a nice place to document the options we support.

The actual code still uses a map, because rewriting all of that is a lot of work. "TODO".

Fixes #856
Fixes #879
Fixes #909
Fixes #1202


qustavo:
this is great, thanks for doing it. Do you think you can cut a new release for this features?

Issue #638, opened 2017-07-20
High usage of memory space in the DialOpen() method


felipeparpinelli:
Hi, 
Our application receives a high load of requests per minute, about 100k/min. We noticed that our memory steadily grown. Doing the memory profiling, we notice a high in-use memory size in the following piece of code inside pq/conn.go in method DialOpen():
![screen shot 2017-07-20 at 11 56 37](https://user-images.githubusercontent.com/1221706/28426083-7f230550-6d48-11e7-89e4-88d5b10a4b75.png)

So, the doubt is: there is a limit to this grown (is it normal?) or we can consider a memory leak? 
We didn't observe a drop in the size of this buffer.

- go version go1.7.6 linux/amd64
- lib/pq version - commit: dd1fe2071026ce53f36a39112e645b4d4f5793a4 (master)

Thank you for your attention.

felipeparpinelli:
Hi, just to give an answer on this problem. The memory leak was actually in this piece of code above.
We changed to the [pgx](https://github.com/jackc/pgx) library and the leak disappeared.

thank you.

arp242:
I wasn't really able to reproduce this in my testing. Closing this on the assumption that it's fixed by #1249.

PR #1248, opened 2026-01-28
Handle pre-protocol errors to prevent memory exhaustion

lieut-data:
When PostgreSQL cannot fork a backend process (e.g., an external process limit), it sends a plain text error like `Ecould not fork new process for connection: Resource temporarily unavailable` before any protocol handshake. The 'E' byte gets parsed as an ErrorResponse message type, but the following bytes are plain text, not a binary length field.

This causes lib/pq to interpret ASCII text as a message length (e.g., "coul" = 0x636f756c = 1.6GB), leading to massive memory allocations and a likely OOM under connection pressure.

This fix matches [libpq's behavior in fe-connect.c](https://github.com/postgres/postgres/blob/c6a10a89feab903b03e2eb664bd6e931b28416a2/src/interfaces/libpq/fe-connect.c#L4067-L4084): if an ErrorResponse has msgLength < 8 or > 30000, treat it as a pre-protocol plain text error and read it as a null-terminated string instead.

There's a decent chance this is related to https://github.com/lib/pq/issues/638.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1029,6 +1029,24 @@ func (cn *conn) saveMessage(typ proto.ResponseCode, buf *readBuf) error {
 	return nil
 }
 
+// maxErrorLength matches libpq's MAX_ERRLEN for detecting pre-protocol errors.
+const maxErrorLength = 30000
+
+// readPreProtocolError reads a pre-protocol plain text error message.
+// When PostgreSQL cannot start a backend (e.g., an external process limit),
+// it sends plain text like "Ecould not fork new process for connection: ..."
+// The 'E' looks like an ErrorResponse type, but the rest is plain text.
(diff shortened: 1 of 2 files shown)


arp242:
Was just about to do a release, so fairly good timing on this.

Do these tweaks seem alright to you? https://github.com/lib/pq/commit/c3c7698c9b472f6708e8130d1a4de5941ac065cd (I can't update this PR because it's created from `master` in your fork instead of a branch; pushing from the git CLI doesn't let me update the `master` branch and I've never been able to figure out how to get around that other than using the GitHub web UI).

arp242:
Merged via #1249

lieut-data:
Nice, thanks @arp242! I'll re-test with my project to verify, but it looks good!

PR #1252, opened 2026-01-29
Fix []byte types incorrectly converted to PostgreSQL array

lieut-data:
In running Mattermost's sqlstore test suite with the new release -- and also to validate https://github.com/lib/pq/pull/1249 -- I ran into a series of errors handling byte slice types. Here's a proposed fix that now passes our test suite.

CheckNamedValue was incorrectly converting byte slice types to PostgreSQL arrays. This affected:
- `*[]byte` pointers (dereference would result in `[]byte` hitting Slice case)
- Named `byte` slice types like `json.RawMessage`

The fix checks if the value's element type is `uint8` after pointer dereferencing, which correctly handles both `[]byte` and named types whose underlying type is `[]byte`.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -829,18 +829,14 @@ func (cn *conn) CheckNamedValue(nv *driver.NamedValue) error {
 	if !v.IsValid() {
 		return driver.ErrSkip
 	}
+	t := v.Type()
+	for t.Kind() == reflect.Ptr {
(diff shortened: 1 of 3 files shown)


arp242:
Cheers, thanks; kind of obvious that should have used reflect instead of a type assertion there 😅

PR #1260, opened 2026-02-09
fix: empty options not accepted by Supavisor

### LINKED ISSUE #1259: `lib/pq` v1.11.1 cannot connect to Supavisor, but v1.10.9 can
I'm not sure what exact change between [v1.10.9](https://github.com/lib/pq/releases/tag/v1.10.9) and [v1.11.1](https://github.com/lib/pq/releases/tag/v1.11.1) is causing the issue, but connecting to [Supavisor](https://github.com/supabase/supavisor) using `lib/pq` v1.11.1 does not work, where v1.10.9 does.

Supabase logs the following when trying to connect with `lib/pq` v1.11.1:

```
supavisor-1  | 02:53:06.913 region=local peer_ip=172.18.0.5 [error] ClientHandler: Client startup message error: :bad_startup_payload
supavisor-1  | 02:53:06.913 region=local peer_ip=172.18.0.5 [warning] client_join is called with a mismatched id: nil
```

Related:
* https://github.com/amacneil/dbmate/issues/746
* https://github.com/amacneil/dbmate/pull/750

dossy:
Supavisor will close the connection and log a :bad_startup_payload
if an empty `options` connection parameter is sent.

Fixes #1259

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,17 @@
+v1.11.2 (2025-02-10)
+--------------------
+This fixes two regressions:
+
+- Don't send startup parameters if there is no value, improving compatibility
+  with Supavisor ([#1260]).
+
+- Don't send `dbname` as a startup parameter if `database=[..]` is used in the
+  connection string. It's recommended to use dbname=, as database= is not a
+  libpq option, and only worked by accident previously. ([#1261])
+
+[#1260]: https://github.com/lib/pq/pull/1260
+[#1261]: https://github.com/lib/pq/pull/1261
+
 v1.11.1 (2025-01-29)
 --------------------
 This fixes two regressions present in the v1.11.0 release:
--- a/conn.go
+++ b/conn.go
@@ -1188,20 +1188,27 @@ func (cn *conn) startup(cfg Config) error {
 	w := cn.writeBuf(0)
 	w.int32(proto.ProtocolVersion30)
 
-	w.string("user")
-	w.string(cfg.User)
-	w.string("database")
-	w.string(cfg.Database)
+	if cfg.User != "" {
+		w.string("user")
+		w.string(cfg.User)
+	}
+	if cfg.Database != "" {
+		w.string("database")
+		w.string(cfg.Database)
+	}
 	// w.string("replication") // Sent by libpq, but we don't support that.
-	w.string("options")
-	w.string(cfg.Options)
+	if cfg.Options != "" {
+		w.string("options")
+		w.string(cfg.Options)
+	}
 	if cfg.ApplicationName != "" {
 		w.string("application_name")
 		w.string(cfg.ApplicationName)
 	}
-	w.string("client_encoding")
-	w.string(cfg.ClientEncoding)
-
+	if cfg.ClientEncoding != "" {
+		w.string("client_encoding")
+		w.string(cfg.ClientEncoding)
+	}
 	for k, v := range cfg.Runtime {
 		w.string(k)
 		w.string(v)


dossy:
@arp242 FYI - I've almost got a clean run of the tests against Supavisor - found a number of places where connections weren't being released, etc., which would cause the session mode connection pool to run out of connections.

arp242:
I can't get it to work on my machine:

	% docker compose up supavisor-create-tenant
	Attaching to supavisor-create-tenant-1
	Container pqgo-supavisor-db-1 Waiting
	Container pqgo-supavisor-db-1 Healthy
	Container pqgo-supavisor-1 Waiting
	Container pqgo-supavisor-1 Error dependency supavisor failed to start
	dependency failed to start: container pqgo-supavisor-1 exited (1)

	% docker compose up supavisor
	Attaching to supavisor-1
	Container pqgo-supavisor-db-1 Waiting
	Container pqgo-supavisor-db-1 Healthy
	supavisor-1  | Setting RLIMIT_NOFILE to 100000
	supavisor-1  | /app/limits.sh: line 6: ulimit: open files: cannot modify limit: Operation not permitted
	supavisor-1 exited with code 1

So, hm, not sure why that fails? 🤔

---

Does it need to have `supavisor-db` PostgreSQL instance? Can't it use the `pg18`?



Ideally it should use the same connection parameters as everything else, except the port being different. This is also how it can "detect" if it's running pgbouncer or supaviser or anything else, which is a bit hacky but straight-forward and works.

It uses the following users for tests:

	pqgo             password=pqgo
	pqgomd5          password=wordpass  (md5 auth only)
	pqgopassword     password=wordpass  (password auth only)
	pqgoscram        password=wordpass  (scram-sha-256 auth only)

	pqgossl          password=pqgo  (require SSL auth)
	pqgosslcert      password=pqgo  (require SSL client certificate auth)

Only the first one is strictly needed, for the rest the tests can be skipped if it's not supported by supaviser or too hard to set up. e.g. for pgbouncer it skips those SSL tests as I couldn't get it to work. Some tests is better than no tests.

It also needs to be added to GitHub actions as a final step.

arp242:
Don't mind doing all of the testing/CI stuff in a separate PR by the way, and just merging the fix now.

dossy:
> I can't get it to work on my machine:
> ```
> % docker compose up supavisor-create-tenant
> ...
> ```

You need to `docker compose up -d supavisor-create-tenant` so that the `supavisor-db` and `supavisor` services continue to run in the background afterwards.  However, this is puzzling:

```
supavisor-1  | /app/limits.sh: line 6: ulimit: open files: cannot modify limit: Operation not permitted
```

The `supavisor` container must run as `root`.  Are you running Docker in rootless mode, or have an override file that's running the container as a non-root user?

> Does it need to have `supavisor-db` PostgreSQL instance? Can't it use the `pg18`?

It may be able to use `pg18` but I'm just following the self-hosted Supavisor guidance in their [`docker-compose.yml`](https://github.com/supabase/supavisor/blob/b360d3fa619aba83d068a20b33310a09f789c4f1/docker-compose.yml) which uses their own `supabase/postgres:15.1.0.148` image.  I don't think they have modified Postgres itself in any way, but presumably the image is configured to work with Supavisor, as opposed to having to adapt the official Postgres Docker image with any configuration or modifications necessary.  Seemed like the path of least resistance, so I took it.

Since Supavisor stores its own data for its own operation in whatever Postgres database it sits in front of, reusing that database for other tests could unintentionally impact Supavisor's operation, so it's best to keep it isolated for the purpose of testing.

Officially, I believe Supavisor supports up to Postgres 17 at the moment, but I suspect it'll work with Postgres 18 just fine, but I don't want to find out that it doesn't and chase ghosts trying to get tests passing only to find out the issue is caused by some change in v18 that requires a change to Supavisor ...

> Don't mind doing all of the testing/CI stuff in a separate PR by the way, and just merging the fix now.

That works for me.  I'll submit the testing changes in a separate PR.

arp242:
Okay, I backed out the compose.yaml changes and merged just the fix.

---

> Are you running Docker in rootless mode, or have an override file that's running the container as a non-root user?

No, just the standard Void Linux setup. I spent 0 time looking in to it so may be something simple.

None of the tests ever write to the database (all uses `create temp table`, which is how they can run in parallel also), so that shouldn't be an issue. It can use pg17 instead of pg18 (or even pg15).

dossy:
> None of the tests ever write to the database (all uses `create temp table`, which is how they can run in parallel also), so that shouldn't be an issue. It can use pg17 instead of pg18 (or even pg15).

Right, but Supavisor does write to the database to store its metadata (tenant configuration, etc.) which, however unlikely, could interfere with tests as the tests are implemented assuming an empty database, but it won't be if the database is shared/being used by Supavisor.

PR #1265, opened 2026-03-03
Fix SSL key permission check to allow modes stricter than 0600/0640

### LINKED ISSUE #1257: ssl key mode 400 can not used
https://github.com/lib/pq/blob/master/internal/pqutil/perm.go

func checkPermissions(fi os.FileInfo) error {
	// The maximum permissions that a private key file owned by a regular user
	// is allowed to have. This translates to u=rw. Regardless of if we're
	// running as root or not, 0600 is acceptable, so we return if we match the
	// regular user permission mask.
	if fi.Mode().Perm()&os.FileMode(0777)^0600 == 0 {
		return nil
	}

	// We need to pull the Unix file information to get the file's owner.
	// If we can't access it, there's some sort of operating system level error
	// and we should fail rather than attempting to use faulty information.
	sys, ok := fi.Sys().(*syscall.Stat_t)
	if !ok {
		return ErrSSLKeyUnknownOwnership
	}

	// if the file is owned by root, we allow 0640 (u=rw,g=r) to match what
	// Postgres does.
	if sys.Uid == 0 {
		// The maximum permissions that a private key file owned by root is
		// allowed to have. This translates to u=rw,g=r.
		if fi.Mode().Perm()&os.FileMode(0777)^0640 != 0 {
			return ErrSSLKeyHasWorldPermissions
		}
		return nil
	}

	return ErrSSLKeyHasWorldPermissions
}

mdz:
The permission check in `checkPermissions` used XOR (`^`) which only accepted
an exact mode match (0600 for user-owned, 0640 for root-owned). Modes more
restrictive than the maximum (e.g. 0400, read-only) were incorrectly rejected.

Use AND NOT (`&^`) to check that no permission bits exceed the allowed maximum
instead.

Fixes #1257

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,17 @@
 unreleased
 ----------
+
+### Features
+
 - Support protocol 3.2, and the `min_protocol_version` and
   `max_protocol_version` DSN parameters ([#1258]).
 
+### Fixes
+
+- Fix SSL key permission check to allow modes stricter than 0600/0640#1265 ([#1265]).
+
 [#1258]: https://github.com/lib/pq/pull/1258
+[#1265]: https://github.com/lib/pq/pull/1265
 
 v1.11.2 (2026-02-10)
 --------------------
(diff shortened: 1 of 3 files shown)


arp242:
Thanks!

PR #1267, opened 2026-03-05
Allow for client certs signed by an intermediate CA.

### LINKED ISSUE #1266: Client certs signed by intermediate CAs are not handled correctly.
### Summary
When a client certificate is signed by an intermediate CA (rather than directly by a root CA), lib/pq fails to send the intermediate certificate as part of the TLS handshake. The server receives only the leaf client certificate, cannot build a chain back to its trusted root, and drops the connection.

### Root Cause
In ssl.go, sslClientCertificates loads the client certificate via tls.LoadX509KeyPair(sslcert, sslkey) (line 216), which produces a tls.Certificate containing only the leaf cert. It never appends intermediate CA certificates to the chain.

Even when sslrootcert contains the intermediate CA cert (which is common practice — bundling root + intermediate in one file), those certs are only loaded into tlsConf.RootCAs by sslCertificateAuthority for server cert verification. They are never added to tlsConf.Certificates[0].Certificate, which is what  Go's TLS client sends during the handshake.

### Symptoms
  - Server logs indicate no client certificate was provided, or that the client certificate could not be verified
  - Client sees a generic connection error (EOF, driver: bad connection, or similar disconnect)
  - Works fine when the client cert is signed directly by the root CA
  - Works fine when the server's ssl_ca_file contains both root and intermediate (since the server can build the chain itself), but fails when the server has only the root CA

benchub:
Before this change, client certs signed by an intermediate CA, when the server only trusts the root CA, were not handled correctly.

Add a unit test for this situation (which now passes) as well as tests for other intermediate CA situations (which passed before this change and continue to pass afterwards.)

Fixes #1266.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,8 +15,12 @@ unreleased
 - Clearer error when starting a new query while pq is still processing another
   query ([#1272]).
 
+- Send intermediate CAs with client certificates, so they can be signed by an
+  intermediate CA ([#1267]).
+
 [#1258]: https://github.com/lib/pq/pull/1258
 [#1265]: https://github.com/lib/pq/pull/1265
+[#1267]: https://github.com/lib/pq/pull/1267
 [#1270]: https://github.com/lib/pq/pull/1270
 [#1272]: https://github.com/lib/pq/pull/1272
 


arp242:
Does it need to use the mockPostgresSSLServer()? All the other SSL tests just run against PostgreSQL. Similarly, does it need to generate the certificates? For the others it's just stored in ./testdata/init (and can be re-generated with the Makefile).

Both are a lot of code to maintain.

benchub:
Ah, I was trying to keep things as self-contained as possible. I'll make a new commit to just keep certs in ./testdata/init/ instead and modify the Makefile to remake them as needed. That's a better idea. Thanks.

I wasn't able to show a failing unit test until I used mockPostgresSSLServer(), because the problem being fixed is how this package does not send up the full chain and that was making Postgres sad.

arp242:
Do the tests not fail if you run that new TestSSLIntermediateCA() against PostgreSQL?

Aside from the extra code to maintain, right now it also doesn't test whether this actually works with PostgreSQL: it only tests that it works with your mock, which may not necessarily be the same as what PostgreSQL does.

For simpler protocol tests that's okay (we already have pqtest.Fake), but SSL verification is probably complex adn subtle enough that it pays to test against PostgreSQL, unless this is unfeasible for some reason (although I don't see why that would be the case off-hand?)

benchub:
Oh! I see your point now. Again, I was trying to make things self-contained, so that we don't have to configure an external postgres install. But of course other tests assume there is an external postgres. Let me lean into that pattern and see where it takes me.

arp242:
You already need PostgreSQL to run many tests. `docker compose up -d` should start a PostgreSQL that should work with `go test`. Also see the README. Let me know if you have trouble with this.

Or you can rely on the GitHub CI.

benchub:
Thanks for the push @arp242; you were absolutely correct that the test was only testing the mock server, which was insufficient.

While fixing this I got annoyed about the consistent two tests test that were failing before I started any of this, and just made the change to let them run cleanly.

arp242:
Looks like there are some test failures?

Note I pushed some small fixes: notably, capturing the `tt` loop variable in the tests. Because go.mod specifies 1.21, it didn't use the new (well, "new") for loops and it appeared to pass as it always used the last test, which is the only one that works.

After fixing that, it consistently fails on my machine and in CI:

```
% go test -run TestSSLClientCertificateIntermediate

--- FAIL: TestSSLClientCertificateIntermediate (0.00s)
    --- FAIL: TestSSLClientCertificateIntermediate/file_certs (0.00s)
        ssl_test.go:200: wrong error
            want:
            have: remote error: tls: unknown certificate authority
    --- FAIL: TestSSLClientCertificateIntermediate/inline_certs (0.00s)
        ssl_test.go:200: wrong error
            want:
            have: pq: connection requires a valid client certificate (28000)
FAIL
exit status 1
FAIL    github.com/lib/pq       0.011s
```

benchub:
Well that exposed some things. New fix incoming.

arp242:
Cheers, thanks

benchub:
\o/

PR #1292, opened 2026-03-18
Update oid package for PostgreSQL 18.3

arp242:
This was last updated for PostgreSQL 9.6 in 2017.

Some types were removed, but keep them around for compatibility as some code references it.


arp242:
Actually not sure this is the correct approach; see my comment: https://github.com/lib/pq/pull/1291#issuecomment-4086349967

I think what we want is to rewrite all of this to only define common types (`text`, `int4`, `bytea`, etc.) and then just query `pg_type` for anything else to get the type info. That will work with custom types and $anything_else.

Issue #1294, opened 2026-03-26
Broken .pgpass path in v1.12.0


mxmCherry:
Extra `.postgresql` dir (suffix) is included in `internal/pqutil.Home()`, looks to be added for certificates, but kills `$HOME/.pgpass` auth.

`.pgpass` is still supposed to be directly in home dir:  https://www.postgresql.org/docs/current/libpq-pgpass.html

Introduced in https://github.com/lib/pq/commit/5e55bb7c2a77b91cf8298cc56a63e92318c692ac#diff-5c9340a8db6deccad27db2ac49dd65f19ff7de98482d73781e7639e0cb9fdbfdR43

Can work around it locally by setting `PGPASSFILE=~/.pgpass` env var for tests/app-runs.

PR #1295, opened 2026-03-26
fix: pqutil: separate Home() and PostgresqlHome()

mxmCherry:
Address https://github.com/lib/pq/issues/1294

Decided that the `.../.postgresql` dir is quite special (and rarely used, only for certs in this project) so decided it deserves a separate helper.

~Sorry, did not bother figuring out how to run tests locally, relying on CI.~ Tests are passing locally. Looks like Home() helper itself is not tested at all, so did not add PostgresqlHome() tests as well.

Pls review carefully.

### DIFF
--- a/connector_test.go
+++ b/connector_test.go
@@ -10,7 +10,6 @@ import (
 	"os"
 	"path/filepath"
 	"reflect"
-	"runtime"
 	"slices"
 	"strings"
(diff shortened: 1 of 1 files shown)


arp242:
Thanks; fixed via #1300

PR #1296, opened 2026-03-26
fix scram count

tl-alex-nicot:
### Problem
The SCRAM-SHA-256 authentication fails with the error:
```
pq: SCRAM-SHA-256 error: server sent an invalid SCRAM-SHA-256 iteration count: "i=256"
```

This occurs when connecting to servers (e.g., AWS RDS) that use iteration counts below 1000.

### Root Cause
The length validation check `len(fields[2]) < 6` requires the iteration field to have at least 6 characters (e.g., `i=1000`). This incorrectly rejects valid iteration counts with fewer than 4 digits, such as `i=256` (5 characters).

### Fix
Changed the minimum length check from `< 6` to `< 3`, which only requires `i=` plus at least one digit. The actual integer validation is still performed by `strconv.Atoi` on the subsequent line.

### Changed Files
- `scram/scram.go`: Line 164

### DIFF
--- a/scram/scram.go
+++ b/scram/scram.go
@@ -161,7 +161,7 @@ func (c *Client) step2(in []byte) error {
 	if !bytes.HasPrefix(fields[1], []byte("s=")) || len(fields[1]) < 6 {
 		return fmt.Errorf("server sent an invalid SCRAM-SHA-256 salt: %q", fields[1])
 	}
-	if !bytes.HasPrefix(fields[2], []byte("i=")) || len(fields[2]) < 6 {
+	if !bytes.HasPrefix(fields[2], []byte("i=")) || len(fields[2]) < 3 {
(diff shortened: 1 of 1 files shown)


arp242:
Could you post the output of connecting to your database `PQGO_DEBUG=1` set? Note this will leak your password, so use a temporary user or password.

I have not looked at this scram code at all, but it's over ten years old and haven't seen many complaints (and I've also seen people using it to connect to RDS). I'm hesitant to "just change it", especially because it's in the authentication workflow.

tl-alex-nicot:
@arp242  I can try but the error surfaces because this package is used in https://github.com/cyrilgdn/terraform-provider-postgresql 

essentially for pg16+ if someone sets scram_iterations <1000 and uses a set of credentials with that strength you get the error

PR #1299, opened 2026-03-28
fix: treat io.ErrUnexpectedEOF as driver.ErrBadConn to prevent connection pool poisoning

### LINKED ISSUE #1298: io.ErrUnexpectedEOF causes permanent connection pool poisoning via stuck inProgress flag
## Summary
A single `io.ErrUnexpectedEOF` during row reading can permanently poison a pooled connection. After the error, every subsequent query on that connection will fail with:
`pq: there is already a query being processed on this connection`

The connection is never evicted from the pool because `database/sql` doesn't see `driver.ErrBadConn`.

## Root cause

`handleError` classifies `io.EOF` as `driver.ErrBadConn`, but not `io.ErrUnexpectedEOF`. When `recvMessage` does an `io.ReadFull` on a partially-received message body and the connection drops mid-read, the result is `io.ErrUnexpectedEOF`, not `io.EOF`. Since `handleError` doesn't recognize this, `cn.err` is never set, `IsValid()` returns `true`, and `database/sql` keeps handing out the broken connection. The CompareAndSwap guard rejects every query, and the non-`ErrBadConn` error means `database/sql` won't retry on a fresh connection.

## How to reproduce
I've hosted a reproduceable demo of this bug at: https://github.com/henhouse/pq-bug-demo.

It uses a TCP fault-injection proxy between a Go client and a real PostgreSQL 16 instance (via Docker). The proxy forwards all traffic transparently, and then on a specific query it will truncate a DataRow response mid-body -- sending the full 5-byte header but only a fraction of the body bytes before closing the connection. This forces `io.ReadFull` to return `io.ErrUnexpectedEOF`.
```bash
git clone https://github.com/henhouse/pq-bug-demo
cd pq-bug-demo
make test-buggy  # demonstrates the poisoning on v1.12.0
make down
```

The Makefile also includes make test-downgrade (v1.10.9, no bug) and make test-fix (v1.12.0 + proposed patch) for comparison.

## Real-world impact
We hit this in production with CockroachDB, where node drains and rebalances routinely produce partial TCP reads. A single truncated response poisoned the connection pool.

## PR Fix
https://github.com/lib/pq/pull/1299

henhouse:
Fixes #1298

When a TCP read is interrupted mid-message (partial DataRow), pq's `recvMessage` returns `io.ErrUnexpectedEOF`. Prior to this fix, `handleError` did not classify this as `driver.ErrBadConn`, so:

1. `cn.err` was never set, and `IsValid()` returned true
2. The `inProgress` atomic flag remained stuck at true (since ReadyForQuery
   never arrived)
3. `database/sql` kept handing out the broken connection
4. The CompareAndSwap guard rejected every subsequent query with
   "there is already a query being processed on this connection"

This is especially impactful with CockroachDB, where node drains and rebalances routinely produce partial reads.

Two changes:
 1. `error.go`: Treat `io.ErrUnexpectedEOF` the same as `io.EOF` in `handleError` - both indicate a dead connection.
 2. `conn.go`: Wrap `errQueryInProgress` with `driver.ErrBadConn` so `database/sql` evicts the connection instead of recycling it. This acts as defense for any future code path that leaves `inProgress` stuck.

Includes unit test for `handleError` and integration test using a TCP fault-injection proxy that truncates a DataRow mid-body to reproduce the exact failure mode.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+unreleased
+----------
+
+- Treat io.ErrUnexpectedEOF as driver.ErrBadConn so database/sql discards the
+  connection. Since v1.12.0 this could result in permanently broken connections,
+  especially with CockroachDB which frequently sends partial messages ([#1299]).
+
+[#1299]: https://github.com/lib/pq/pull/1299
+
 v1.12.1 (2026-03-30)
 --------------------
 
--- a/conn_test.go
+++ b/conn_test.go
@@ -16,6 +16,7 @@ import (
 	"runtime"
 	"strconv"
 	"strings"
+	"sync/atomic"
 	"testing"
 	"time"
 
@@ -507,7 +508,7 @@ func TestErrorDuringStartupClosesConn(t *testing.T) {
 
 func TestBadConn(t *testing.T) {
 	t.Parallel()
-	for _, tt := range []error{io.EOF, &Error{Severity: pqerror.SeverityFatal}} {
+	for _, tt := range []error{io.EOF, &Error{Severity: pqerror.SeverityFatal}, io.ErrUnexpectedEOF} {
 		t.Run(fmt.Sprintf("%s", tt), func(t *testing.T) {
 			var cn conn
 			err := cn.handleError(tt)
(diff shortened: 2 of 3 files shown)


arp242:
Can't this use the existing pqtest.Fake to test this error? Over 250 lines of code for a single test is quite a lot to maintain going forward, and it's not immediately obvious what it does exactly either.

I'm also okay with just running this against a real database. I made #1304 to run it in the CI, although I need to finish up some bits.

Is changing errQueryInProgress intentional or a left-over from some previous debugging?

henhouse:
@arp242 Oops yeah that was left behind, thanks.

Pushed an update that uses pqtest instead which reduces it a lot. If you'd prefer not having a test for this case entirely, I can just remove it. It was mostly to provide a in-repo example of the demo I linked to that reproduces the case in full. The check in `error_test.go` is minimally sufficient to check for a regression in this case.

arp242:
Thanks!

Issue #1103, opened 2022-12-29
Rewrite the array.go to support generic array type?


liur180855:
Hello,

AFAIK this is current use case of Generic Array
```go
//  var x []sql.NullInt64
//  db.QueryRow(`SELECT ARRAY[235, 401]`).Scan(pq.Array(&x))
```

if we can rewrite or add another version of generic array to support the following use case in sqlx

```go
type Foobar struct {
}

type Foo struct {
	Bar pq.GenericArrayV2[bool]
	Baz pq.GenericArrayV2[Foobar]
}
```

PR #1309, opened 2026-04-02
implement fmt.Stringer for specialised arrays

Ferada:
This adds a bunch of methods to the specialised arrays that come out of `Array` to support debugging.

At the moment printing them with `%v" just results in a pointer (except for the `GenericArray` actually), which is obviously not super useful when you just want to see the arguments that get passed to a query. There's probably any number of ways to do this, love to hear your feedback if there's a better way.

### DIFF
--- a/array.go
+++ b/array.go
@@ -76,6 +76,14 @@ type ArrayDelimiter interface {
 // BoolArray represents a one-dimensional array of the PostgreSQL boolean type.
 type BoolArray []bool
 
+// String implements the fmt.Stringer interface.
+func (a *BoolArray) String() string {
+	if a == nil {
+		return "<nil>"
+	}
+	return fmt.Sprint(*a)
+}
+
 // Scan implements the sql.Scanner interface.
 func (a *BoolArray) Scan(src any) error {
 	switch src := src.(type) {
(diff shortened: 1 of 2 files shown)


arp242:
Not sure about this to be honest. It prints things identical to what fmt.Println() would print without a String() method, but hides some bits. This seems very confusing: it's not at all obvious there's a fmt.Stringer there and that it's printing something other from what you have. This seems the worst of everything: it neither prints things accurately, but it's also not significantly more helpful than "just dump the data".

Printing the Value() would probably be better, if we must change something here. Maybe with some extra spacing for readability.

I'd rather pick this up as part of #1103 rather than change soon-to-be-deprecated methods. I have a working branch for that already.

Ferada:
> Not sure about this to be honest. It prints things identical to what fmt.Println() would print without a String() method, but hides some bits.

Well right now it's printing pointers where for debugging purposes I'd rather see the data in the slice itself, which isn't printed.

> Printing the Value() would probably be better, if we must change something here. Maybe with some extra spacing for readability.

Maybe? My main concern was the wrapper structs not being transparent when printing values. If the Value() return value was printed at least for some values where otherwise you wouldn't see anything (but pointers) that'd make sense too.

> I'd rather pick this up as part of #1103 rather than change soon-to-be-deprecated methods. I have a working branch for that already.

Oh sure, if it's in flux anyway don't bother with this.

arp242:
> for debugging purposes I'd rather see the data in the slice itself, which isn't printed.

Except that it's obscuring that it is a pointer, which is exactly the sort of thing that can be very confusing when debugging. I don't love how fmt.Print prints pointers, but it is what it is.

Also, if I run all your test cases on master (without these changes) then it all fails with:

    --- FAIL: TestArrayString/[]string#02 (0.00s)
        array_test.go:650:
            have: "&[a b c]"
            want: "[a b c]"

So not really clear what it improves either.

PR #1314, opened 2026-04-07
Add generic ArrayOf[T any] type

### LINKED ISSUE #1103: Rewrite the array.go to support generic array type?

arp242:
Add generic ArrayOf[T any] type

This adds a new generic ArrayOf[T any] type:

		var a pq.ArrayOf[int]
		db.QueryRow("...").Scan(&a)

This replaces all the existing Array* types, which are now implemented
with ArrayOf type aliases and marked as deprecated. The existing tests
are unchanged except for some error texts, and I believe ths should be
fully compatible.

Only GenericArray is left as-is, as re-implementing that with ArrayOf in
a way that's fully compatible is kind of tricky. I'd rather just leave
it as-is instead of spending the effort and risk breakage.

The main issue is that we rely on:

		//go:linkname convertAssign database/sql.convertAssign
		func convertAssign(dest, src any) error

Which does not exactly spark joy, however there is already a comment on
convertAssign():

		// convertAssign should be an internal detail,
		// but widely used packages access it using linkname.
		// Notable members of the hall of shame include:
		//   - ariga.io/entcache
		//
		// Do not remove or change the type signature.
		// See go.dev/issue/67401.

So if compatibility is already guaranteed because of other packages,
then I guess it's okay to join this "hall of shame". Also, it looks like
it'll probably just get exported in the future:
https://github.com/golang/go/issues/62146#issuecomment-3921700836

Fixes #1103

### DIFF
--- a/array.go
+++ b/array.go
@@ -4,758 +4,270 @@ import (
 	"bytes"
 	"database/sql"
 	"database/sql/driver"
-	"encoding/hex"
 	"fmt"
 	"reflect"
 	"strconv"
 	"strings"
+	"time"
+	_ "unsafe"
 )
 
-var typeByteSlice = reflect.TypeFor[[]byte]()
-var typeDriverValuer = reflect.TypeFor[driver.Valuer]()
-var typeSQLScanner = reflect.TypeFor[sql.Scanner]()
+var (
+	_ sql.Scanner   = (*ArrayOf[any])(nil)
+	_ driver.Valuer = (*ArrayOf[any])(nil)
+)
 
-// Array returns the optimal driver.Valuer and sql.Scanner for an array or
-// slice of any dimension.
+// TODO: hopefully will be exported in the future?
+// https://github.com/golang/go/issues/62146#issuecomment-3921700836
 //
-// For example:
+//go:linkname convertAssign database/sql.convertAssign
+func convertAssign(dest, src any) error
+
+// ArrayOf wraps a slice with Scan() and Value() methods that use PostgreSQL's
+// array syntax.
 //
-//	db.Query(`SELECT * FROM t WHERE id = ANY($1)`, pq.Array([]int{235, 401}))
+// T may optionally implement the "ArrayDelimiter() string" method to use a
+// different array delimiter.
 //
-//	var x []sql.NullInt64
-//	db.QueryRow(`SELECT ARRAY[235, 401]`).Scan(pq.Array(&x))
+// XXX: document array element conversion:
 //
-// Scanning multi-dimensional arrays is not supported.  Arrays where the lower
-// bound is not one (such as `[0:0]={1}') are not supported.
-func Array(a any) interface {
-	driver.Valuer
-	sql.Scanner
-} {
-	switch a := a.(type) {
-	case []bool:
-		return (*BoolArray)(&a)
-	case []float64:
-		return (*Float64Array)(&a)
-	case []float32:
-		return (*Float32Array)(&a)
-	case []int64:
-		return (*Int64Array)(&a)
-	case []int32:
-		return (*Int32Array)(&a)
-	case []string:
-		return (*StringArray)(&a)
-	case [][]byte:
-		return (*ByteaArray)(&a)
+// [sql.ConvertAssign]
+//
+// T can be any primitive type (string, int, etc.). The Scan() method is called
+// for every array element if T implements [sql.Scanner].
+type ArrayOf[T any] []T
 
-	case *[]bool:
-		return (*BoolArray)(a)
-	case *[]float64:
-		return (*Float64Array)(a)
-	case *[]float32:
-		return (*Float32Array)(a)
-	case *[]int64:
-		return (*Int64Array)(a)
-	case *[]int32:
-		return (*Int32Array)(a)
-	case *[]string:
-		return (*StringArray)(a)
-	case *[][]byte:
-		return (*ByteaArray)(a)
-	}
-
-	return GenericArray{a}
-}
-
-// ArrayDelimiter may be optionally implemented by driver.Valuer or sql.Scanner
-// to override the array delimiter used by GenericArray.
-type ArrayDelimiter interface {
-	// ArrayDelimiter returns the delimiter character(s) for this element's type.
-	ArrayDelimiter() string
-}
-
-// BoolArray represents a one-dimensional array of the PostgreSQL boolean type.
-type BoolArray []bool
-
-// Scan implements the sql.Scanner interface.
-func (a *BoolArray) Scan(src any) error {
+func (a *ArrayOf[T]) Scan(src any) error {
 	switch src := src.(type) {
 	case []byte:
-		return a.scanBytes(src)
+		return a.scan(src)
 	case string:
-		return a.scanBytes([]byte(src))
+		return a.scan([]byte(src))
 	case nil:
 		*a = nil
 		return nil
 	}
-
-	return fmt.Errorf("pq: cannot convert %T to BoolArray", src)
+	return fmt.Errorf("pq: cannot convert %T to %T", src, a)
 }
 
-func (a *BoolArray) scanBytes(src []byte) error {
-	elems, err := scanLinearArray(src, []byte{','}, "BoolArray")
+func (a *ArrayOf[T]) scan(src []byte) error {
+	var zero T
+	dims, elems, err := parseArray(src, arrayDelimiter(any(zero)))
 	if err != nil {
 		return err
 	}
-	if *a != nil && len(elems) == 0 {
-		*a = (*a)[:0]
-	} else {
-		b := make(BoolArray, len(elems))
-		for i, v := range elems {
-			if len(v) != 1 {
-				return fmt.Errorf("pq: could not parse boolean array index %d: invalid boolean %q", i, v)
-			}
-			switch v[0] {
-			case 't':
-				b[i] = true
-			case 'f':
-				b[i] = false
-			default:
-				return fmt.Errorf("pq: could not parse boolean array index %d: invalid boolean %q", i, v)
-			}
-		}
-		*a = b
-	}
-	return nil
-}
-
-// Value implements the driver.Valuer interface.
-func (a BoolArray) Value() (driver.Value, error) {
-	if a == nil {
-		return nil, nil
-	}
-
-	if n := len(a); n > 0 {
-		// There will be exactly two curly brackets, N bytes of values,
-		// and N-1 bytes of delimiters.
-		b := make([]byte, 1+2*n)
-
-		for i := range n {
-			b[2*i] = ','
-			if a[i] {
-				b[1+2*i] = 't'
-			} else {
-				b[1+2*i] = 'f'
-			}
-		}
-
-		b[0] = '{'
-		b[2*n] = '}'
-
-		return string(b), nil
-	}
-
-	return "{}", nil
-}
-
-// ByteaArray represents a one-dimensional array of the PostgreSQL bytea type.
-type ByteaArray [][]byte
-
-// Scan implements the sql.Scanner interface.
-func (a *ByteaArray) Scan(src any) error {
-	switch src := src.(type) {
-	case []byte:
-		return a.scanBytes(src)
-	case string:
-		return a.scanBytes([]byte(src))
-	case nil:
-		*a = nil
-		return nil
-	}
-
-	return fmt.Errorf("pq: cannot convert %T to ByteaArray", src)
-}
-
-func (a *ByteaArray) scanBytes(src []byte) error {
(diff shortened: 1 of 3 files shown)


deprecated.go
```diff
@@ -138,3 +141,301 @@ func makeStmt(b *bytes.Buffer, columns ...string) {
...
+	case *[]float32:
+		return (*Float32Array)(a)
+	case *[]int64:
+		return (*Int64Array)(a)
+	case *[]int32:
+		return (*Int32Array)(a)
+	case *[]string:
+		return (*StringArray)(a)
+	case *[][]byte:
+		return (*ByteaArray)(a)
```
Ferada:
```suggestion
	case *[]bool:
		return (*ArrayOf[bool])(a)
	case *[]float64:
		return (*ArrayOf[float64])(a)
	case *[]float32:
		return (*ArrayOf[float32])(a)
	case *[]int64:
		return (*ArrayOf[int64])(a)
	case *[]int32:
		return (*ArrayOf[int32])(a)
	case *[]string:
		return (*ArrayOf[string])(a)
	case *[][]byte:
		return (*ArrayOf[[]byte])(a)
```
Unless there was a reason to use the deprecated names here?

array.go
```diff
@@ -4,758 +4,257 @@ import (
...
+		case uint32:
+			b = strconv.AppendUint(b, uint64(v), 10)
+		case uint64:
+			b = strconv.AppendUint(b, v, 10)
+		case float32:
+			b = strconv.AppendFloat(b, float64(v), 'f', -1, 32)
+		case float64:
+			b = strconv.AppendFloat(b, v, 'f', -1, 64)
+		case bool:
+			if any(aa).(bool) {
```
Ferada:
```suggestion
			if v {
```

array.go
```diff
@@ -4,758 +4,257 @@ import (
...
+			switch rv.Kind() {
+			case reflect.String:
+				b = appendArrayQuotedText(b, []byte(rv.String()))
+			case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+				b = strconv.AppendInt(b, rv.Int(), 10)
+			case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64:
+				b = strconv.AppendUint(b, rv.Uint(), 10)
+			case reflect.Float32, reflect.Float64:
+				b = strconv.AppendFloat(b, rv.Float(), 'f', -1, 32)
+			case reflect.Ptr:
```
Ferada:
```suggestion
			case reflect.Bool:
				if rv.Bool() {
					b = append(b, 't')
				} else {
					b = append(b, 'f')
				}
			case reflect.Ptr:
```
Probably just missing the boolean case?

array.go
```diff
@@ -4,758 +4,257 @@ import (
...
-
-	if b, del, err = appendArrayElement(b, rv.Index(0)); err != nil {
-		return b, del, err
-	}
-
-	for i := 1; i < n; i++ {
-		b = append(b, del...)
-		if b, del, err = appendArrayElement(b, rv.Index(i)); err != nil {
-			return b, del, err
+	for i, aa := range a {
```
Ferada, line 178: Given that the element type is already known, wouldn't it be faster to move the type assertion out of the loop for the common cases?

That is, specialise this block for the common slice types via something like:

```go
switch xx := any([]T(a)).(type) {
case []bool:
	...
case []uint8:
	b = appendAllUint(xx, b, del)
case []uint16:
	b = appendAllUint(xx, b, del)
...
default:
	// only use reflection on individual elements here
	for i, aa := range a {
		...
	}
}
```

The duplication for the signed and unsigned types could also be minimised with something like:

```go
func appendAllUint[T uint8 | uint16 | uint32 | uint64](xx []T, b []byte, del []byte) []byte {
	for i, aa := range xx {
		if i > 0 {
			b = append(b, del...)
		}
		b = strconv.AppendUint(b, uint64(aa), 10)
	}
	return b
}

// similar for the other types
```

Ferada (commented):
Tried this on a project and everything still works, replacing the deprecated types with `ArrayOf` is also easy :+1:

Edit: Pushed suggestions below to [this commit](https://github.com/lib/pq/commit/d2e62ed7b2e61249fe5929ee61caec1c38a98e71).

AGWA:
I tried to use `ArrayOf` with the following type, which expects to be scanned from a bytea value:

```go
type SHA256Digest [32]byte 
        
func (digest *SHA256Digest) Scan(src any) error {
        switch src := src.(type) {
        case []byte:
                if len(src) != len(digest) {
                        return fmt.Errorf("cannot scan SHA256Digest from %d bytes (%d bytes expected)", len(src), len(digest))
                }
                copy(digest[:], src)
                return nil
        case nil:
                return errors.New("cannot scan SHA256Digest from NULL column")
        default:
                return fmt.Errorf("cannot convert column of type %T to SHA256Digest", src)
        }       
}
```

Unfortunately, it doesn't work because the value which `ArrayOf[T].scan` passes to `convertAssign` is formatted as text (e.g. `\x001122`) instead of being the raw bytes that you normally get when scanning a bytea value, so my Scan method fails.

This could be fixed if `ArrayOf[T].scan` passed a string to `convertAssign` instead of a `[]byte`:

```go
err := convertAssign(&b[i], string(v))
```

Then, my Scan method could add a case for string which parses the textual format.

I believe passing a string is more correct than `[]byte` because PostgreSQL array elements are always text, not raw bytes. It also seems more consistent with how [`ArrayOf[any]` is treated as string](https://github.com/lib/pq/blob/53e356fe35b76b93102d10d8c69d3e36143dbc72/array.go#L96).

My type doesn't work with `GenericArray` either, but fixing it there would break compatibility. Since you're leaving `GenericArray` as-is, this seems like a good opportunity to make this change.

P.S.  Many thanks for taking over maintainership of this package!

arp242:
@AGWA The Scan() method gets called for every array element, so using that SHA256Digest.Scan() with ArrayOf (or GenericArray) is not really correct: that method should process a single PostgreSQL array element, not the entire array.

This is how GenericArray works, so I copied that to ArrayOf. It kind of makes sense, but I also have to admit that I was confused for a bit about this myself after coming back to this after several months. At the very least it should be documented properly (I added a TODO comment in my rebase). There are also some changes in the upcoming Go 1.27 surrounding this that may help, but I haven't really looked at them yet. I've been holding off merging this until Go 1.27 for that reason.

Using an array (rather than slice) should probably work, and check the length. Then your Scan() method becomes superfluous and you can use ArrayOf[[32]byte] or ArrayOf[SHA256Digest].

Passing []byte or string to convertAssign() seems unrelated? I think you're just "using it wrong" (which is not your fault, as it is more confusing than it should be).

AGWA:
@arp242 Thanks for looking at this.

I think I'm using things correctly. Let me try to explain my use case better.

Currently, my `SHA256Digest` type works great for scanning a single (non-array) 32-byte `bytea` value.
The `Scan` method gets called with a 32-byte `[]byte` containing the raw bytes.  This code works as expected:

```go
var digest SHA256Digest
if err := db.QueryRow(`SELECT '\x01ba4719c80b6fe911b091a7c05124b64eeece964e09c058ef8f9805daca546b'::bytea`).Scan(&digest); err != nil {
	log.Fatal(err)
}
fmt.Printf("%x\n", digest) // prints 01ba4719c80b6fe911b091a7c05124b64eeece964e09c058ef8f9805daca546b
```

But when I try to use it with `ArrayOf` and scan a `bytea[]`, it fails instead of printing two lines as expected:

```go
var digests pq.ArrayOf[SHA256Digest]
if err := db.QueryRow(`SELECT ARRAY['\x01ba4719c80b6fe911b091a7c05124b64eeece964e09c058ef8f9805daca546b'::bytea, '\xb5bb9d8014a0f9b1d61e21e796d78dccdf1352f23cd32812f4850b878ae4944c'::bytea]`).Scan(&digests); err != nil {
	log.Fatal(err)
}
for _, digest := range digests {
	fmt.Printf("%x\n", digest)
}
```

It fails because `Scan` is called with a 66 byte `[]byte` containing `\x` followed by 64 hex characters, rather than 32 raw bytes.

If I change `Scan` like this, then it works with `ArrayOf`:

```diff
 func (digest *SHA256Digest) Scan(src any) error {
 	switch src := src.(type) {
 	case []byte:
-		if len(src) != len(digest) {
-			return fmt.Errorf("cannot scan SHA256Digest from %d bytes (%d bytes expected)", len(src), len(digest))
-		}
-		copy(digest[:], src)
-		return nil
+		_, err := hex.Decode(digest[:], src[2:])
+		return err
 	case nil:
 		return errors.New("cannot scan SHA256Digest from NULL column")
 	default:
 		return fmt.Errorf("cannot convert column of type %T to SHA256Digest", src)
 	}
 }
```

But now it can't scan a non-array `bytea` anymore.

I could define two different types, `SHA256Digest` and `SHA256DigestArrayItem`, each with a different `Scan` implementation, but ideally I could use a single type and `Scan` could distinguish whether it's being called with raw bytes or a hex string. For example:

```go
 func (digest *SHA256Digest) Scan(src any) error {
 	switch src := src.(type) {
 	case []byte:
		if len(src) != len(digest) {
			return fmt.Errorf("cannot scan SHA256Digest from %d bytes (%d bytes expected)", len(src), len(digest))
		}
		copy(digest[:], src)
		return nil
	case string:
		hexString, ok := strings.CutPrefix(src, `\x`)
		if !ok {
			return fmt.Errorf("cannot scan SHA256 from string without \\x prefix")
		}
		rawBytes, err := hex.DecodeString(hexString)
		if err != nil {
			return fmt.Errorf("cannot scan SHA256 from this string because it contains invalid hex: %s", err)
		}
		if len(rawBytes) != len(digest) {
			return fmt.Errorf("cannot scan SHA256 from %d bytes (%d bytes expected)", len(rawBytes), len(digest))
		}
		copy(digest[:], rawBytes)
		return nil
	case nil:
		return errors.New("cannot scan SHA256Digest from NULL column")
	default:
		return fmt.Errorf("cannot convert column of type %T to SHA256Digest", src)
	}
 }
```

My suggestion to pass a `string` to convertAssign instead of a `[]byte` would enable this, but maybe there's a better way?

arp242:
Ehm, yes, you are right – I had my mental model set to another project I worked on and I got confused. Thanks for explaining.

When I tried to just change it to a string this afternoon some other tests broke, I didn't look in to why though – may be a simple fix (or not).

Essentially the issue can be summarised as "make sure a Scan() method behaves identical for both regular (non-array) and array cases" – it doesn't matter if it's a string or []byte *as such*, it matters that they're different.

AGWA:
That is a correct summary.

When I tried changing this a few weeks ago, only one test broke and it was trivial (it was looking for `[]uint8` in an error message instead of `string`) but I see you've made some changes since then so I will take another look.

AGWA:
It's still just the one test. Here's the complete diff:

```diff
diff --git a/array.go b/array.go
index db10c49..478628b 100644
--- a/array.go
+++ b/array.go
@@ -113,7 +113,7 @@ func (a *ArrayOf[T]) scan(src []byte) error {
                                continue // Just use zero value
                        }
 
-                       err := convertAssign(&b[i], v)
+                       err := convertAssign(&b[i], string(v))
                        if err != nil {
                                return fmt.Errorf("pq: array index %d: %s", i, strings.TrimPrefix(err.Error(), "sql/driver: "))
                        }
diff --git a/array_test.go b/array_test.go
index 0c32087..4fc7053 100644
--- a/array_test.go
+++ b/array_test.go
@@ -590,7 +590,7 @@ func TestArrayOf(t *testing.T) {
 
                {&ArrayOf[int]{}, `{1,NULL,2}`, `array index 1: cannot convert NULL to int`},
                {&ArrayOf[string]{}, `{"a",NULL,"b"}`, `array index 1: cannot convert NULL to string`},
-               {&ArrayOf[int]{}, `{"asd"}`, `array index 0: converting driver.Value type []uint8 ("asd") to a int: invalid syntax`},
+               {&ArrayOf[int]{}, `{"asd"}`, `array index 0: converting driver.Value type string ("asd") to a int: invalid syntax`},
                {&ArrayOf[time.Time]{}, `{"2020-02-03 19:20:21Z",NULL,"2021-02-03 19:20:21Z"}`, `array index 1: cannot convert NULL to time.Time`},
                {&ArrayOf[time.Time]{}, `{"asd"}`, `array index 0: invalid timestamp`},
        }
```

PR #1319, opened 2026-04-27
Simplify tests by using t.Setenv

alexandear:
This PR refactors tests with `t.Setenv` instead of `os.Setenv`. This allows us to remove `t.Cleanup` sections.


alexandear:
@arp242 Thanks for reviewing. Could you share why this was closed and whether a revised version would be welcome? I would appreciate guidance on the preferred direction.

PR #1321, opened 2026-05-09
Route trailing readReadyForQuery error through handleError on extended-protocol cleanup

### LINKED ISSUE #1320: Discarded readReadyForQuery() EOF causes connection pool poisoning via stuck inProgress flag
## Summary

A pooler-side `ErrorResponse` followed by an immediate connection close (no trailing `ReadyForQuery`) can permanently poison a pooled connection. After the trigger, every subsequent query on that connection fails with: `pq: there is already a query being processed on this connection`.

The connection is never evicted from the pool because `database/sql` doesn't see `driver.ErrBadConn`. This is the same end-state as #1298, but reached through a different bug-path that the merge of #1299 (commit `6d77ced`) does not close.

## Root cause

Five sites in `conn.go` (`readParseResponse`, `readStatementDescribeResponse`, `readPortalDescribeResponse`, `readBindResponse`, `postExecuteWorkaround`) handle a mid-extended-protocol `ErrorResponse` by draining the trailing `ReadyForQuery` and discarding whatever it returns:

```go
case proto.ErrorResponse:
    err := parseError(r, "")
    _ = cn.readReadyForQuery()
    return err
```

When the peer closes mid-stream, `readReadyForQuery()` returns `io.EOF`. The `_ =` drops it before `handleError` can classify it, so `cn.err` is never set, `IsValid()` returns `true`, and `database/sql` keeps handing out the broken connection. The CompareAndSwap guard rejects every subsequent query with `errQueryInProgress` — which is not `driver.ErrBadConn`, so `(*DB).retry` won't retry on a fresh connection either. The change merged via #1299 only addresses `io.ErrUnexpectedEOF` in `handleError`; on this path the EOF is dropped *before* `handleError` is reached.

## How to reproduce

A reproducer is at: https://github.com/m1ralx/pq-bug-demo

It uses a TCP fault-injection proxy between a Go client and a real PostgreSQL 16 instance (via Docker). On a specific Parse, the proxy writes a hand-crafted `ErrorResponse` (severity ERROR, SQLSTATE 08P01) directly to the client and closes the connection before any `ReadyForQuery` reaches the client. This mirrors pgbouncer 1.15's `disconnect_server(false, ...) -> send_pooler_error(client, false, ...)` byte sequence.

```bash
git clone https://github.com/m1ralx/pq-bug-demo
cd pq-bug-demo
make test-buggy  # demonstrates the poisoning on v1.12.3
make test-fix    # passes with the proposed patch
make down
```

## Real-world impact

We hit this in stage environment with services routed through pgbouncer ≥1.15. 

## PR Fix

https://github.com/lib/pq/pull/1321

m1ralx:
Fixes #1320 

When a transparent pooler (e.g. pgbouncer's `disconnect_server(false, ...) -> send_pooler_error`) emits a synthetic `ErrorResponse` mid-extended-protocol and closes the socket before sending a trailing `ReadyForQuery`, the trailing `readReadyForQuery()` inside `conn.go`'s five extended-protocol cleanup sites returns `io.EOF` — and that error is silently dropped by `_ =`. Prior to this fix, this means:

1. `cn.err` was never set (the EOF never reached `handleError`), and `IsValid()` returned true
2. The `inProgress` atomic flag remained stuck at true (since `ReadyForQuery` never arrived to clear it in `recvMessage`)
3. `database/sql` kept handing out the broken connection from the pool
4. The CompareAndSwap guard rejected every subsequent query with `pq: there is already a query being processed on this connection`

This is the same end-state as #1298, but reached through a different bug-path that the merge of #1299 ([commit 6d77ced41719616090c9e7eec2c313a18640bc3f](https://github.com/lib/pq/commit/6d77ced41719616090c9e7eec2c313a18640bc3f)) does not close. The merged change classifies `io.ErrUnexpectedEOF` in `handleError`, but for this path the EOF is dropped *before* `handleError` ever sees it. The defensive `errQueryInProgress -> driver.ErrBadConn` wrap from that PR was rejected in review and is not in upstream, so `(*DB).retry` does not silently retry either. The path is fully open in current `master`.

This is especially impactful in production behind pgbouncer ≥1.15: the byte sequence is severity ERROR (not FATAL) followed by a TCP close with no `ReadyForQuery`. The non-FATAL severity also means the parsed `*pq.Error` is not classified as `driver.ErrBadConn` by `handleError`, so the only remaining signal — the EOF on the trailing `ReadyForQuery` read — was the one that needed to propagate.

One change:

1. `conn.go`: route the result of `cn.readReadyForQuery()` through `cn.handleError` at the five extended-protocol cleanup sites (`readParseResponse`, `readStatementDescribeResponse`, `readPortalDescribeResponse`, `readBindResponse`, `postExecuteWorkaround`). `handleError(nil)` is a no-op so the happy path is unaffected; for `io.EOF`, `io.ErrUnexpectedEOF`, and `*net.OpError` the existing `cn.err.set(driver.ErrBadConn)` side effect runs even though the return value is still discarded by `_ =`. `*sql.DB.putConn -> dc.IsValid()` then returns false and the conn is closed instead of being returned to `freeConn`.

```diff
 case proto.ErrorResponse:
     err := parseError(r, "")
-    _ = cn.readReadyForQuery()
+    _ = cn.handleError(cn.readReadyForQuery())
     return err
```

Includes integration test using `pqtest.Fake` that emits a non-fatal `ErrorResponse` (SQLSTATE `08P01`) and closes the connection without `ReadyForQuery`, then asserts that a subsequent `db.Exec` does **not** short-circuit to `errQueryInProgress` — i.e. the poisoned conn was actually evicted from the pool rather than merely flagged. The test reproduces the exact production failure mode reported in #1320.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1760,7 +1760,7 @@ func (cn *conn) readParseResponse() error {
 		return nil
 	case proto.ErrorResponse:
 		err := parseError(r, "")
-		_ = cn.readReadyForQuery()
+		_ = cn.handleError(cn.readReadyForQuery())
 		return err
 	default:
 		cn.err.set(driver.ErrBadConn)
@@ -1788,7 +1788,7 @@ func (cn *conn) readStatementDescribeResponse() (paramTyps []oid.Oid, colNames [
 			return paramTyps, colNames, colTyps, nil
 		case proto.ErrorResponse:
 			err := parseError(r, "")
-			_ = cn.readReadyForQuery()
+			_ = cn.handleError(cn.readReadyForQuery())
 			return nil, nil, nil, err
 		default:
 			cn.err.set(driver.ErrBadConn)
@@ -1809,7 +1809,7 @@ func (cn *conn) readPortalDescribeResponse() (rowsHeader, error) {
 		return rowsHeader{}, nil
 	case proto.ErrorResponse:
 		err := parseError(r, "")
-		_ = cn.readReadyForQuery()
+		_ = cn.handleError(cn.readReadyForQuery())
 		return rowsHeader{}, err
 	default:
 		cn.err.set(driver.ErrBadConn)
@@ -1827,7 +1827,7 @@ func (cn *conn) readBindResponse() error {
 		return nil
 	case proto.ErrorResponse:
 		err := parseError(r, "")
-		_ = cn.readReadyForQuery()
+		_ = cn.handleError(cn.readReadyForQuery())
 		return err
 	default:
 		cn.err.set(driver.ErrBadConn)
@@ -1853,7 +1853,7 @@ func (cn *conn) postExecuteWorkaround() error {
 		switch t {
 		case proto.ErrorResponse:
 			err := parseError(r, "")
-			_ = cn.readReadyForQuery()
+			_ = cn.handleError(cn.readReadyForQuery())
 			return err
 		case proto.CommandComplete, proto.DataRow, proto.EmptyQueryResponse:
 			// the query didn't fail, but we can't process this message
--- a/conn_test.go
+++ b/conn_test.go
@@ -7,6 +7,7 @@ import (
 	"database/sql"
 	"database/sql/driver"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"maps"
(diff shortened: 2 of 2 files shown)


arp242:
I'm looking at this, and I wonder if it wouldn't be a better fix to reset inProgress in a more reliable way? Because this is already the second such bug to occur with it (previous: #1299). When I added this (#1272) I thought this should be fairly safe to do because it adapted a long-existing mechanism, but it seems I was wrong about that – I guess that was always a bit buggy but no one noticed because it's used fairly infrequently. So I wonder how many more subtle bugs exist here?

arp242:
Or maybe just backing out #1272 altogether, or re-think how to do that from scratch. It didn't really fix a bug per-se, but rather just clarifies the error when using pq in the wrong way. So in that sense it's not super-critical.

I had to read your description carefully to follow what's going on here and it was not obvious from just the diff. That's a good indication that this entire logic is not quite right, and even if we do get it 100% right now I can see missing something here in a future change.

m1ralx:
Yes, I think the error-handling mechanism here is fairly fragile in its current form. 

I considered wrapping `cn.err.set(driver.ErrBadConn)` in some helper that would also do `inProgress.Store(false)`, and calling it inside `readReadyForQuery` when handling the `cn.recv1` error – but I held off, since that starts to look more like a refactor that requires more context than I have.

arp242:
This seems to work well:

```
diff --git i/conn.go w/conn.go
index 3ed7ecf..7e28fb9 100644
--- i/conn.go
+++ w/conn.go
@@ -289,6 +289,7 @@ restartAll:
                if app(err, cfg) {
                        continue
                }
+               cn.c = connWrap{Conn: cn.c, cn: cn}

                err = cn.ssl(cn.cfg, mode)
                if err != nil && mode == SSLModePrefer {
@@ -442,6 +443,20 @@ func (cn *conn) checkTSA(tsa TargetSessionAttrs) error {
        }
 }

+type connWrap struct {
+       net.Conn
+       cn *conn
+}
+
+func (c connWrap) Read(b []byte) (int, error) {
+       n, err := c.Conn.Read(b)
+       return n, c.cn.handleError(err)
+}
+func (c connWrap) Write(b []byte) (int, error) {
+       n, err := c.Conn.Write(b)
+       return n, c.cn.handleError(err)
+}
+
 func dial(ctx context.Context, d Dialer, cfg Config) (net.Conn, error) {
        network, address := cfg.network()
```

Seems to fix both this issue and #1299 if I back out the fix for that.

Does cause handleError() to be run twice for various cases, but I think that should be fine.

---

For context: previously all of the error handling was done via panics; that handleError() used to be the panic recover(). I changed that back in January and seems to work well (#1232), but unfortunately does mean this code is structured in a way that's less than ideal.

m1ralx:
```
+       n, err := c.Conn.Read(b)
+       return n, c.cn.handleError(err)
```

Won't this break anything? 
It changes the result returned by `io.ReadFull`, since `handleError` rewrites `io.EOF` into `driver.ErrBadConn`:

https://github.com/lib/pq/blob/6d77ced41719616090c9e7eec2c313a18640bc3f/error.go#L310-L311

 and that's the error `io.ReadFull` relies on to convert a partial read into
  `io.ErrUnexpectedEOF`:

```
	} else if n > 0 && err == EOF {
		err = ErrUnexpectedEOF
	}
``` 
https://github.com/golang/go/blob/release-branch.go1.23/src/io/io.go#L340-L342

RaduBerinde:
I believe the issue described in https://github.com/lib/pq/issues/1324#issuecomment-4628088584 is related to this.

arp242:
I decided to just revert the new behaviour for now, as I don't really have the time to properly fix it now, and correct behaviour beats a clearer error message. So that makes this PR superfluous.

PR #1323, opened 2026-05-18
conn: mark connection bad if readReadyForQuery fails after ErrorResponse

### LINKED ISSUE #1320: Discarded readReadyForQuery() EOF causes connection pool poisoning via stuck inProgress flag

c-tonneslan:
Closes #1320.

The five mid-extended-protocol error handlers (`readParseResponse`, `readStatementDescribeResponse`, `readPortalDescribeResponse`, `readBindResponse`, `postExecuteWorkaround`) drain the trailing `ReadyForQuery` and silently discard its result. When the peer closes the connection at that moment, `readReadyForQuery` returns `io.EOF`, the error is dropped, `cn.err` stays clean, and `database/sql` keeps treating the connection as healthy. The `CompareAndSwap` that guards `inProgress` then rejects every subsequent query with *there is already a query being processed*.

Set `driver.ErrBadConn` when the drain fails so `database/sql` evicts the poisoned connection from the pool.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1760,7 +1760,9 @@ func (cn *conn) readParseResponse() error {
 		return nil
 	case proto.ErrorResponse:
 		err := parseError(r, "")
-		_ = cn.readReadyForQuery()
+		if rfqErr := cn.readReadyForQuery(); rfqErr != nil {
+			cn.err.set(driver.ErrBadConn)
+		}
 		return err
 	default:
 		cn.err.set(driver.ErrBadConn)
@@ -1788,7 +1790,9 @@ func (cn *conn) readStatementDescribeResponse() (paramTyps []oid.Oid, colNames [
 			return paramTyps, colNames, colTyps, nil
 		case proto.ErrorResponse:
 			err := parseError(r, "")
-			_ = cn.readReadyForQuery()
+			if rfqErr := cn.readReadyForQuery(); rfqErr != nil {
+				cn.err.set(driver.ErrBadConn)
+			}
 			return nil, nil, nil, err
 		default:
 			cn.err.set(driver.ErrBadConn)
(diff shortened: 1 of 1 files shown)


arp242:
Stop creating these duplicate PRs.

PR #1326, opened 2026-07-05
Only treat oversized ErrorResponse as plain text during startup

### LINKED ISSUE #1324: ErrorResponse > 30000 bytes is misparsed as a pre-protocol plain-text error
### Summary

Since **v1.11.0** (PR #1249, commit `0ad30496`), `recvMessage` treats *any* `ErrorResponse` whose payload length exceeds `MaxErrlen` (30000) as a pre-protocol plain-text error. A legitimate v3 `ErrorResponse` larger than 30 KB is therefore never parsed into its fields — the caller gets a garbled string instead of the real error.

### Affected code

`conn.go` (v1.12.3, ~line 1088):

```go
// libpq checks "if ErrorResponse && (msgLength < 8 || msgLength > MAX_ERRLEN)" ...
if t == proto.ErrorResponse && (n < 4 || n > proto.MaxErrlen) {
    msg, _ := cn.buf.ReadString('\x00')
    return 0, fmt.Errorf("pq: server error: %s%s", string(x[1:]), strings.TrimSuffix(msg, "\x00"))
}
```

`MaxErrlen = 30_000` (`internal/proto/proto.go`).

### Why it's wrong

The guard from PR #1249 is correct for its intended case (a backend that can't fork sends `Ecould not fork…` as plain text *before* the protocol handshake).

But libpq only treats `msgLength > 30000` as a pre-protocol error during connection establishment (`fe-connect.c`, `PQconnectPoll` → `CONNECTION_AWAITING_RESPONSE`), where the comment notes it's coping with a pre-3.0-protocol server. In steady-state message parsing (`fe-protocol3.c`, `pqParseInput3`) there *is* a 30000 sanity check, but `ErrorResponse` is **explicitly exempt** via `VALID_LONG_MESSAGE_TYPE` (alongside `DataRow`, `CopyData`, `NoticeResponse`, etc.):

```c
#define VALID_LONG_MESSAGE_TYPE(id) \
    ((id) == PqMsg_CopyData || \
     (id) == PqMsg_DataRow || \
     (id) == PqMsg_ErrorResponse || \
     (id) == PqMsg_FunctionCallResponse || \
     (id) == PqMsg_NoticeResponse || \
     (id) == PqMsg_NotificationResponse || \
     (id) == PqMsg_RowDescription)

...

if (msgLength > 30000 && \!VALID_LONG_MESSAGE_TYPE(id))
{
    handleSyncLoss(conn, id, msgLength);
    return;
}
```

So libpq deliberately allows large error messages once the protocol is established; lib/pq instead applies the startup-only cap to every `ErrorResponse`, mangling valid errors > 30 KB.

The returned string is also nonsense: `string(x[1:])` is the 4 raw big-endian length bytes, so the error reads like `pq: server error: <binary length><start of body>` (e.g. `pq: server error: �NSERROR`, where `�N` are length bytes and `SERROR` is the start of the `S`everity field).

### Reproduction

Against any PostgreSQL server (no special setup):

```go
db, _ := sql.Open("postgres", connStr)
// Raise an error whose message field alone is ~40 KB.
_, err := db.Exec(`DO $$ BEGIN RAISE EXCEPTION '%', repeat('x', 40000); END $$;`)
fmt.Println(err)
// v1.10.9:  pq: xxxxxxxx... (the real message)
// v1.11.0+: pq: server error: <garbage>   (fields never parsed)
```

We hit this in CockroachDB, whose schema-changer errors embed graphviz URLs and routinely exceed ~50 KB.

### Suggested fix

Scope the `> MaxErrlen` heuristic to the connection-startup path only (mirroring libpq), or drop the upper bound in steady-state `recvMessage` and rely on the length prefix. The `< 8`/`< 4` lower-bound and OOM protection for the pre-handshake case can be preserved.

Affected versions: **v1.11.0 – v1.12.3** (latest).

momomuchu:
Fixes #1324

Since #1249, recvMessage treats any ErrorResponse over MaxErrlen (30000) as a pre-protocol plain-text error. libpq only applies that heuristic while awaiting the startup response (fe-connect.c, PQconnectPoll); after the handshake an ErrorResponse is explicitly allowed to exceed 30000 bytes (fe-protocol3.c, VALID_LONG_MESSAGE_TYPE).

Applying it in steady state does two things to a legitimate large error, e.g. `RAISE EXCEPTION '%', repeat('x', 40000)`: the error comes back garbled (the raw length bytes plus the first field instead of the parsed message), and the connection is left desynced, since the early return never drains the message body or the trailing ReadyForQuery. inProgress stays stuck, the conn is not marked bad, and the pool hands it back out, so the next query fails with "there is already a query being processed on this connection" (see the second comment on #1324 for the full trace).

This gates the heuristic on a new startup parameter to recvMessage, true only from conn.recv, which is only used during the startup/auth sequence. recv1Buf and the direct callers in copy.go and notify.go pass false and parse large ErrorResponses normally, matching libpq. The pre-handshake "could not fork" case from #1249 keeps working as before.

This is unrelated to #1320/#1321, which is an EOF-handling bug in the extended-protocol cleanup paths; same end symptom, different cause and code path.

Test uses pqtest.Fake to send a MaxErrlen+10000 ErrorResponse and asserts the message and SQLSTATE parse intact and that a second query on the same pooled connection still works.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1101,7 +1101,14 @@ func (cn *conn) saveMessage(typ proto.ResponseCode, buf *readBuf) error {
 
 // recvMessage receives any message from the backend, or returns an error if
 // a problem occurred while reading the message.
-func (cn *conn) recvMessage(r *readBuf) (proto.ResponseCode, error) {
+//
+// startup must only be true for callers reading messages during the
+// connection-startup sequence (see [conn.recv]). It is never appropriate for
+// steady-state message parsing: a v3 ErrorResponse is a normal, properly
+// length-prefixed message post-handshake, and treating a large one as plain
+// text (see the comment below) desynchronizes the wire and leaves inProgress
+// stuck, poisoning the connection.
+func (cn *conn) recvMessage(r *readBuf, startup bool) (proto.ResponseCode, error) {
 	// workaround for a QueryRow bug, see exec
 	if cn.saveMessageType != 0 {
 		t := cn.saveMessageType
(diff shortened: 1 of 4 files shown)


arp242:
I fixed it via #1327, which more accurately copies what libpq does wrt. message sizes.

PR #1332, opened 2026-07-15
Set ALPN for direct SSL connect

### LINKED ISSUE #1331: Server rejects connection when sslnegotiation=direct is set
## Actual Behavior

Connecting with `sslnegotiation=direct` fails against a PostgreSQL 17 server with SSL enabled, even though `sslnegotiation=direct` is meant to be supported (`conn.go` explicitly skips the legacy SSLRequest handshake when it's set). The client gets `EOF`, and in the server logs we see:

```
LOG:  received direct SSL connection request without ALPN protocol negotiation extension
```

## Expected behavior

A connection with `sslmode=require&sslnegotiation=direct` (or `verify-ca`/`verify-full`, the other two modes `NewConfig` accepts alongside `sslnegotiation=direct`) should succeed against a PostgreSQL 17+ server with SSL enabled.

## Minimal reproduction

Requires Docker and Go.

1. Generate a self-signed cert and start plain Postgres 17 with SSL on:

```sh
openssl req -new -x509 -days 3650 -nodes -text \
  -out server.crt -keyout server.key -subj "/CN=localhost"
chmod 600 server.key

docker run -d --name pq-repro \
  -e POSTGRES_PASSWORD=postgres \
  -p 55432:5432 \
  -v "$PWD/server.crt:/var/lib/postgresql/server.crt:ro" \
  -v "$PWD/server.key:/var/lib/postgresql/server.key:ro" \
  postgres:17 \
  -c ssl=on -c ssl_cert_file=/var/lib/postgresql/server.crt -c ssl_key_file=/var/lib/postgresql/server.key
```

2. Connect with lib/pq:

```go
package main

import (
        "database/sql"
        "fmt"

        _ "github.com/lib/pq"
)

func main() {
        db, err := sql.Open("postgres",
                "postgres://postgres:postgres@127.0.0.1:55432/postgres?sslmode=require&sslnegotiation=direct")
        if err != nil {
                panic(err)
        }
        defer db.Close()

        if err := db.Ping(); err != nil {
                fmt.Println("ping error:", err) // ping error: EOF
                return
        }
        fmt.Println("connected OK")
}
```

`go.mod`:

```
module pq-repro

go 1.23

require github.com/lib/pq v1.12.3
```

Output: `ping error: EOF`. The server log shows `received direct SSL connection request without ALPN protocol negotiation extension`.

When we drop `sslnegotiation=direct` from the DSN (plain `sslmode=require`), the client connects fine against the same server. So the TLS setup per se works fine.

## Environment

- lib/pq v1.12.3
- Go 1.26.5, darwin/arm64 (also reproduces on linux/amd64)
- PostgreSQL 17.10 (official `postgres:17` Docker image)

danielmitterdorfer:
When `sslnegotiation=direct` is set, clients receive `EOF`. This happens because Postgres expects the ALPN `postgresql` to be set but the client's TLS configuration doesn't specify it. With this commit we explicitly set the expected ALPN, allowing connections to succeed.

Closes #1331

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ unreleased
 
 ### Fixes
 
+- `sslnegotiation=direct` didn't work due to missing ALPN protocol [[#1332]).
+
 - Revert "clearer error when starting a new query while pq is still processing
   another query ([#1272])" as the mechanism proved to be fragile ([#1327]).
 
@@ -31,6 +33,7 @@ unreleased
 [#1317]: https://github.com/lib/pq/pull/1317
 [#1326]: https://github.com/lib/pq/pull/1326
 [#1327]: https://github.com/lib/pq/pull/1327
+[#1332]: https://github.com/lib/pq/pull/1327
 
 
 v1.12.3 (2026-04-03)
--- a/internal/pqtest/pqtest.go
+++ b/internal/pqtest/pqtest.go
@@ -35,6 +35,16 @@ func SkipCockroach(t testing.TB) {
 	}
 }
 
+// SkipBeforeVersion skips the test unless the connected server reports at least
+// the given major version (e.g. 16, 17, etc.)
+func SkipBeforeVersion(t testing.TB, want int) {
+	t.Helper()
+	have := QueryRow[int](t, MustDB(t), `show server_version_num`)["server_version_num"] / 10000
+	if want > have {
+		t.Skipf("skipped for PostgreSQL %d (want %d)", have, want)
+	}
+}
+
 func ForceBinaryParameters() bool {
 	v, ok := os.LookupEnv("PQTEST_BINARY_PARAMETERS")
 	if !ok {
--- a/internal/proto/proto.go
+++ b/internal/proto/proto.go
@@ -14,6 +14,7 @@ const (
 	CancelRequestCode = (1234 << 16) | 5678
 	NegotiateSSLCode  = (1234 << 16) | 5679
 	NegotiateGSSCode  = (1234 << 16) | 5680
+	ALPNProtocol      = "postgresql"
 )
 
 // Constants from fe-protocol3.c
--- a/ssl.go
+++ b/ssl.go
@@ -15,6 +15,7 @@ import (
 	"sync"
 
 	"github.com/lib/pq/internal/pqutil"
+	"github.com/lib/pq/internal/proto"
 )
 
 // Registry for custom tls.Configs
@@ -120,6 +121,12 @@ func ssl(cfg Config, mode SSLMode) (func(net.Conn) (net.Conn, error), error) {
 
 	tlsConf.MinVersion = cfg.SSLMinProtocolVersion.tlsconf()
 	tlsConf.MaxVersion = cfg.SSLMaxProtocolVersion.tlsconf()
+
+	// ALPN is mandatory with direct SSL connections; PostgreSQL only supports
+	// one protocol.
+	if cfg.SSLNegotiation == SSLNegotiationDirect {
+		tlsConf.NextProtos = []string{proto.ALPNProtocol}
+	}
 
 	// RFC 6066 asks to not set SNI if the host is a literal IP address (IPv4 or
 	// IPv6). This check is coded already crypto.tls.hostnameInSNI, so just
--- a/ssl_test.go
+++ b/ssl_test.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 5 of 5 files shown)


danielmitterdorfer:
In https://github.com/lib/pq/actions/runs/29415703853/job/87371180560, all tests for Postgres before 17.0 were failing. This happened because we are testing a feature that has been introduced with 17.0. I searched the code base for established practices and it seems there are skip helpers but none of them based on versions. I have pushed https://github.com/lib/pq/pull/1332/commits/8b7f21de91e610539465d343d2be45f51f0d92d5 to introduce a smaller helper to ensure that these tests only run on Postgres 17+. Please let me know if 

* the chosen approach is fine for you
* whether it's ok to keep it in this PR (on the other hand it probably doesn't make a lot of sense to introduce these helpers without using them)

arp242:
Thanks; seems good.

The tests can probably be a bit simpler? I think just establishing an actual connection with `sslnegotiation=direct` should be enough? I'm not sure if any of the other tests you added are really needed? That is:

```
diff --git i/internal/pqtest/pqtest.go w/internal/pqtest/pqtest.go
index 8a174b9..7d853a8 100644
--- i/internal/pqtest/pqtest.go
+++ w/internal/pqtest/pqtest.go
@@ -35,6 +35,16 @@ func SkipCockroach(t testing.TB) {
 	}
 }
 
+// SkipBeforeVersion skips the test unless the connected server reports at least
+// the given major version (e.g. 16, 17, etc.)
+func SkipBeforeVersion(t testing.TB, want int) {
+	t.Helper()
+	have := QueryRow[int](t, MustDB(t), `show server_version_num`)["server_version_num"] / 10000
+	if want > have {
+		t.Skipf("skipped for PostgreSQL %d (want %d)", have, want)
+	}
+}
+
 func ForceBinaryParameters() bool {
 	v, ok := os.LookupEnv("PQTEST_BINARY_PARAMETERS")
 	if !ok {
diff --git i/internal/proto/proto.go w/internal/proto/proto.go
index 5aaa516..32c1ec9 100644
--- i/internal/proto/proto.go
+++ w/internal/proto/proto.go
@@ -14,6 +14,7 @@ const (
 	CancelRequestCode = (1234 << 16) | 5678
 	NegotiateSSLCode  = (1234 << 16) | 5679
 	NegotiateGSSCode  = (1234 << 16) | 5680
+	ALPNProtocol      = "postgresql"
 )
 
 // Constants from fe-protocol3.c
diff --git i/ssl.go w/ssl.go
index 081aa9d..0b7603a 100644
--- i/ssl.go
+++ w/ssl.go
@@ -15,6 +15,7 @@ import (
 	"sync"
 
 	"github.com/lib/pq/internal/pqutil"
+	"github.com/lib/pq/internal/proto"
 )
 
 // Registry for custom tls.Configs
@@ -121,6 +122,12 @@ func ssl(cfg Config, mode SSLMode) (func(net.Conn) (net.Conn, error), error) {
 	tlsConf.MinVersion = cfg.SSLMinProtocolVersion.tlsconf()
 	tlsConf.MaxVersion = cfg.SSLMaxProtocolVersion.tlsconf()
 
+	// ALPN is mandatory with direct SSL connections; PostgreSQL only supports
+	// one protocol.
+	if cfg.SSLNegotiation == SSLNegotiationDirect {
+		tlsConf.NextProtos = []string{proto.ALPNProtocol}
+	}
+
 	// RFC 6066 asks to not set SNI if the host is a literal IP address (IPv4 or
 	// IPv6). This check is coded already crypto.tls.hostnameInSNI, so just
 	// always set ServerName here and let crypto/tls do the filtering.
diff --git i/ssl_test.go w/ssl_test.go
index 8b8641b..c8ff68c 100644
--- i/ssl_test.go
+++ w/ssl_test.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"strings"
 	"testing"
 	"time"
 
@@ -92,7 +93,9 @@ func TestSSLMode(t *testing.T) {
 		// sslmode=disable
 		{"sslmode=disable user=pqgossl", "or:no encryption|login rejected (08P01)|authentication rejected by configuration (28000)"},
 
-		// sslnegotiation=direct should fail if ssl isn't required, like libpq:
+		// sslnegotiation=direct
+		{"sslmode=require sslnegotiation=direct", ""},
+		// should fail if ssl isn't required, like libpq:
 		// psql: error: weak sslmode "allow" may not be used with sslnegotiation=direct (use "require", "verify-ca", or "verify-full")
 		{"sslmode=disable sslnegotiation=direct", "weak sslmode"},
 		{"sslmode=allow sslnegotiation=direct", "weak sslmode"},
@@ -103,6 +106,10 @@ func TestSSLMode(t *testing.T) {
 		t.Run("", func(t *testing.T) {
 			t.Parallel()
 
+			if strings.Contains(tt.connect, "sslnegotiation=direct") {
+				pqtest.SkipBeforeVersion(t, 17) // Only supported on PostgreSQL≥17.
+			}
+
 			_, err := pqtest.DB(t, tt.connect)
 			switch {
 			case tt.wantErr == "" && err != nil:

```

danielmitterdorfer:
Thanks for your feedback. I have applied all your suggestions now in https://github.com/lib/pq/pull/1332/commits/d763c6959da8bc9f9b62ced00c85f98d630454a3.

arp242:
Cheers, thanks!

